PR #6, opened 2018-03-19
Remove RPM code and move Documentation folder

jsws:
This pull removes the unused RPM things and moves the documentation folder form /WebApp/Documentation to /Documentation.
Although at some point most of this documentation can be put into the Wiki.


Anders-Markvardsen:
As discussed

Issue #15, opened 2018-03-21
Manual Submission script submits jobs with the wrong location


jsws:
The manual submission script gets the datafile location by querying ICAT for the cycle using the dates of the investigation that the run is in.

This is unreliable for various reasons and we can simply get the file location from the 'location' field of a datafile in ICAT, this will always be correct.

PR #19, opened 2018-03-23
Improvements to manual submission script

jsws:
- Fixes issue #15 by getting the data file location from the Datafile entity in ICAT rather than using unreliable cycles.
- Simplifies the script making it more readable.
- Adds a README with how to use the script
- Uses argparse so it's clearer how to call the script with command line arguments.
- Removes the "walk through" way of getting the input from the user, using command line arguments are the only way to call the script now.


Scripts/ManualSubmissionScript/manual-submission.py
```diff
@@ -0,0 +1,77 @@
+import icat
+from settings import ACTIVE_MQ
+from settings import ICAT
```
ElliotAOram: @jsws The settings.py.template should be updated to use `ICAT = {...}` instead of the current `ICAT_SETTINGS = {...}`

ElliotAOram:
Also server time seems to be an hour ahead. Which is odd as if anything I'd expect it an hour behind after our change to BST?

jsws:
@ElliotAOram 
The incorrect time being displayed is an unrelated issue and is to do with the web app, I've created a new issue #22. 

The reason the example fails is because the calibration file isn't on the dev machine but the job itself has been submitted correctly. It doesn't really matter which instrument/run I put as the example because there's no guarantee that the calibration files for that run will exist in the future. It's more just to give users the idea of the syntax of the commands.

Scripts/ManualSubmissionScript/manual-submission.py
jsws: Fixed in c4c1561

Issue #17, opened 2018-03-22
Add Continuous integration with Travis


ElliotAOram:
I've already done some testing and this should be pretty easy to do. The `.travis.yml` script written (atleast a basic one for pylint) and I have got this running on my own fork of `autoreduce`. I need organisation rights on the repo to be able to add travis so this issue will just cover adding the basic script. In addition, this will also resolve the Pylint errorr current known in the QueueProcessor

ElliotAOram:
@jsws This issue is for just adding travis to the repo (can be combined with #21)

Issue #11, opened 2018-03-19
Fallback mechanism for end-of-run-monitor


Anders-Markvardsen:
For end of run monitor script. Should we have a fallback to monitor ICAT also for new jobs. Create an issue for this

ElliotAOram:
We discussed that we should have a fallback mechanism (possibly 2). The current mechanism queries if the `endofrunmonitor.txt` file in each instrument directory on the data archive has changed. 
Fallback possibilities discussed include:
* Querying ICAT to check the most recent run and comparing this to the autoreduction database (MySQL db) - discussed this should be roughly once every minute or so. 
  * *If the current checking of the `.txt` files fails, ensure that when we use ICAT data, we say this in the logs.*
* Checking for actually file creation in the data archive itself and comparing this to the reduction table.

ElliotAOram:
Remaining work:
* [ ] merge db + travis ticket
* [ ] populate local host db
* [ ] access localhost db from test 
* [ ] run as a service from daemon
* [ ] work out how best to kick off a rebuild if db is out of sync with archive
* [ ] ensure that this does not throw exceptions (should instead log output to file on error / debug)

ElliotAOram:
restart EndOfRunMonitor when archive monitor is un-synced

ElliotAOram:
being replaced by #203 #204 #205

Issue #7, opened 2018-03-19
Add tests for queue processors


Anders-Markvardsen:
Issue should cover adding new unit tests the the `AutoReductionProcessor` and the `QueueProcessor` with a view to improving robustness. Later these should be migrated to automated tests with a CI tool.
* [x] Assess what is currently in place to test the queue processor**s**.
  * Nothing currently
* [x] Investigate what tests we require - match the test criteria to the common faults (investigate these if unknown)
  * Currently unknown so will start with base cases for testing / setting up a structure to enable testing
* [ ] Write unit tests to cover test cases.

ElliotAOram:
Closing this as it is now duplicate of #230

PR #20, opened 2018-03-26
Database build scripts and travis file

ElliotAOram:
Files added to generate an identical schema to that which is currently used in autoreduction. Additionally, a .travis.yml file has been added. The travis config currently performs a simple test to ensure all expected TABLE names are present in the database generated at local host. travis has not been added to the main repo yet but only has to be enabled and then added as the CI solution to Pull requests. 

The set up scripts are in two flavours (these should not both be in the repository permanently!). One `travis-db-full.sql` has the full construction of all the schema in one file. The other `SQL` files have singular database constructions. My reason for adding these is to highlight that we have 3 separate databases:
* `reduction`
* `django`
* `webapp`
yet they are all stored in one database. Looking at an ER diagram for these, there is no link between these 3 sets of data. I would suggest that storing them in a single database is confusing. Unless I am missing something, would there be a reason not to separate these into 3 separate db's on a single server?

Ideally, these files should be tested using Travis, but it should be easy enough to perform the same tests locally. The `.travis.yaml` file defines the steps required to do this:

1. Ensure you have MySQL and a way to start it as a service on localhost  (On my machine  (windows) I use xxamp) 
1. With xxamp, start the service (open client and Start the `MySQL` process`)
1. run `mysql -uroot < Scripts\Build\travis-db-full.sql` to create the database structure on local host
1. pip install `nose` if you don't already have this
1. run `nosetests Scripts\Build\test\test_db_generation.py` and see that the tests passes

This issue is associated with #7, #11, #17 (it does not close any)


ElliotAOram:
Put on hold until we decide about project structure re. git repo/sub repos

ElliotAOram:
I believe this is now required. 

We discussed pulling the data out from Django using the `manage.py showmigrations` functionality. However, as far as I understand this would require login details for the database and would do not want to expose this on travis as they would be openly accessible.

ElliotAOram:
Pylint errors were introduced due to moving the `test_settings.py` to `settings.py`

jsws:
Looks good to me :+1:

PR #23, opened 2018-03-28
Add activemq to travis setup

ElliotAOram:
This allow travis to set up an ActiveMQ queuing system on local host. There is also an accompanying test.
When we come to use this, we might need to add the expected queues in the setup, but I believe this will be done automatically.

To test this we will need to integrate travis into the repo. Providing the travis output passes, we are good to go


ElliotAOram:
This should now me targeted correctly at the `develop` branch. Although I want to investigate how much effort it would be to get the existing tests running on local host (those in `Scripts/ActiveMQTests/`)

jsws:
Looks good to me :+1:

PR #25, opened 2018-04-05
Fix Pylint warnings and add Pylint to travis

ElliotAOram:
This branch covers the change of adding pylint static analysis to the code base. The .travis.yml file has been added but we need to add travis to the repo to ensure that it runs smoothly.

In addition we need to start the various services on the development nodes to ensure they run as expected before merging to the development branch


ElliotAOram:
This is really good to get in 👍 I haven't managed to break anything myself so I think we are safe to merge this to dev

ElliotAOram:
This is really good to get in 👍 I haven't managed to break anything myself so I think we are safe to merge this to dev

PR #26, opened 2018-04-09
Add testconfig.py.template

ElliotAOram:
This removes the testconfig.py file in favour of using a template document instead (as is done else where in the project)

Given the nature of this change, it should be merged with master and development


ElliotAOram:
Note: CI failed as I cancelled the build, this branch does not have the CI changes for pylint and is not affected by them

jsws:
@ElliotAOram Would it make more sense to make this `settings.py.template` instead of `testconfig.py.template` to follow the other convention in the project?

ElliotAOram:
Agreed

ElliotAOram:
Should be fine now

PR #40, opened 2018-04-27
Fix issues on the development nodes

ElliotAOram:
The changes below fix the issues on the development nodes, and improve our ability to catch similar problems in the future.
After this issue is reviewed and merged, we should try to keep the dev node up-to-date with changes. I'll add a git pull request template that covers the manual testing that we should be doing to validate our changes before merging


jsws:
This looks like it will solve the issue :+1:

PR #42, opened 2018-04-27
add setup py file

### LINKED ISSUE #34: Add setup.py to python path setup
We need to add a `setup.py` to the project to allow for `pip install -e autoreduce`. This will mean that the python project can be added to the PYTHONPATH and the relative imports will work correctly.

We should also update the documentation to reflect this change

ElliotAOram:
**Description of changes**
* Add a basic setup.py file that can be used with pip install -e to initialise the python path for the project and allow for absolute imports.
* small tidy up to ignore migration files in project (these are generated when creating the test database)

**How to test**
* Test on local machine
* Run `pip install -e autoreduce` from the parent directory of the autoreduce project.
* Ensure the command executes as expected
* Ensure that you can perform absolute imports in the project

Fixes #34

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ settings.py
 */target/**
 
 venv/
+migrations/
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,7 @@
+from distutils.core import setup
+
+setup(name='AutoReduction',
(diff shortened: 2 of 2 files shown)


jsws:
Works well :+1:

PR #43, opened 2018-04-27
Simplfy travis script

ElliotAOram:
**Description of changes**
Minimise the code we write by hand in travis in favour of using build scripts

**How to test**
Check the build passes on travis


ElliotAOram:
This can be further improved, but is a good start for now. When we look at the system test, we should look at moving all the separate build scripts into one

jsws:
I made the shell scripts executable in the repository which allowed the removal of `chmod` from the `.travis.yml `.

ElliotAOram:
We should further clean up the requirements.txt to only include project requirements and then let those requirement download any sub requirements they need

PR #52, opened 2018-05-03
Implement archive monitor to poll data archive directory

### LINKED ISSUE #11: Fallback mechanism for end-of-run-monitor

ElliotAOram:
**Description of changes**
* Implemented the archive monitor to poll the ISIS data archive for files that may have been missed by watchdog observing EndOfRunMonitor.  
* Added generated data to the mock database being used for testing
* Added a data archive creator to create a fake data archive that can be used for testing
* Refactored monitors to use generic windows service class instead of duplicating code

**How to test**

* Ensure the unit tests pass and travis is happy. 

**Dev if data is coming through**

* Deploy this on development and turn off the EndOfRunMonitor. 
  * We may need to wait some time before a run is found, but this monitor should pick up the difference between the reduction database and the archive and run resubmit the job. 

**Locally**

* It should be possible to test this by:
  * [creating a local archive](https://github.com/ISISScientificComputing/autoreduce/wiki/Testing-and-validating#creating-a-fake-data-archive)
  * Set the `utils/settings.py` `INST_PATH` to be the file directory you created the archive in e.g. if I made the archive in `C:\data-archive` it would be: `INST_PATH = os.path.join('C:\\', 'data-archive', 'NDX{}', 'Instrument')`
  * Install the archive monitor service with `python isis_archive_monitor_win_service.py install`
  * You may wish to lower the polling rate (by default it is 10mins) you can do this in the `isis_archive_monitor_helper.py`
  * Start the service running in the service manager
  * Add data files to the archive data directories and check the logs pick them up correctly. If you like you can hook these into the dev node and check they get pushed through to the database/webapp

Fixes #11


ElliotAOram:
This is still having some trouble on the dev node (mainly due to not being able to fins the network drive). I'll update this when I work out what is causing this.

ElliotAOram:
I have now fixed the problem we were seeing on the dev node (I needed to start the process as a user that had permissions to the data archive - I have started this with my own credentials for the time being)

ElliotAOram:
I think we should now monitor this to see what happens. Note the errors seen in the dev queues are due to insufficient space on the queue nodes. I have now cleaned them and will keep an eye on what happens today.

ElliotAOram:
#66 should be merged before this PR

DavidFair:
@ElliotAOram This PR has developed merge conflicts

monitors/archive_monitor/isis_archive_monitor_helper.py
```diff
@@ -0,0 +1,48 @@
+"""
+File to store messages and data relating to ISIS_archive_monitor
+"""
+
+# ================================= Data ======================================= #
+
+LOG_FORMAT = '%(asctime)s : %(message)s'
+STOMP_LOG_FORMAT = '%(asctime)s :      %(message)s'
+VALID_INST = ['GEM', 'POLARIS', 'WISH', 'MUSR', 'OSIRIS']
```
DavidFair: Do we not have a list of valid instruments elsewhere in the codebase. I would be tempted to place this in a central location for other validators (present or future)

monitors/archive_monitor/isis_archive_monitor_helper.py
```diff
@@ -0,0 +1,48 @@
...
+SLEEP_TIME = 600
+
+# =========================== Logging Messages ================================= #
+
+INVALID_INSTRUMENT_MSG = 'Archive monitor could not be ' \
+                         'started as %s was not recognised as a valid instrument. ' \
+                         'If this is new, have you added it to the VALID_INST list?'
+
+NO_INSTRUMENT_IN_DB_MSG = '     Returning None -- Unable to find instrument %s.'
```
DavidFair: `Unable to find instrument %s in DB`

monitors/archive_monitor/isis_archive_monitor_win_service.py
```diff
@@ -0,0 +1,61 @@
+"""
+Holds the class for overidding QueueService framework
+"""
+# Can't import on travis (linux) server
+# pylint:disable=import-error
```
DavidFair: Could we do something like:
```
import os
if os.system() == "nt":
    return
```

So that the imports below are only attempted on Windows?

monitors/archive_monitor/isis_archive_monitor_win_service.py
```diff
@@ -0,0 +1,61 @@
...
+        servicemanager.LogMsg(servicemanager.EVENTLOG_INFORMATION_TYPE,
+                              servicemanager.PYS_SERVICE_STARTED,
+                              (self._svc_name_, ''))
+        monitor = ArchiveMonitor("POLARIS")
+        while 1:
+            monitor.perform_check()
+            sleep_timer = 0
+            end_service_loop = False
+            while sleep_timer < SLEEP_TIME:
+                time_to_sleep = 30 if SLEEP_TIME > 30 else SLEEP_TIME
```
DavidFair: Why 30? Could we encapsulate this magic value?

monitors/end_of_run_monitor/isis_monitor_win_service.py
```diff
@@ -6,71 +6,50 @@
 As the ISIS archive would be mounted as a drive, this causes
 difficulties whne watching for file changes.
 """
-# ToDo: check if unused-import warning is legitimate for win32con, win32evtlogutil
-# pylint: disable=import-error, unused-import
+# Can't import on travis (linux) server
+# pylint:disable=import-error
+import sys
```
DavidFair: Ditto about import on Windows

utils/test_settings.py
```diff
@@ -28,3 +29,17 @@
     'DB': 'autoreduction',
     'PORT': '3306',
 }
+
+PATH_TO_FILE = os.path.dirname(os.path.realpath(__file__))
+PATH_TO_TEST_OUTPUT = os.path.join(PATH_TO_FILE, 'test-output')
+
+try:
+    os.makedirs(PATH_TO_TEST_OUTPUT)
```
DavidFair: Do we want to place logic like this in `test_settings`? Having side effects of importing data structures could cause problems down the line

monitors/archive_monitor/isis_archive_monitor_win_service.py
ElliotAOram: I believe this is the max length of time a windows will try and stop a service for before it times out but you're right we should make this clearer!

utils/test_settings.py
ElliotAOram: You are right. Have you got a better suggestion for dealing with this I'd prefer not to use hard coded file paths if we can avoid it

monitors/archive_monitor/isis_archive_monitor_helper.py
ElliotAOram: I agree but this may form part of the larger discussion regarding settings in autoreduction which we had on Wednesday. We should decide what our structure for settings should look like first then move these there (could well be part of a separate PR depending on when that decision is made)

ElliotAOram:
@DavidFair do you know why we are not running coveralls on this (or at least why it doesn't show up in the checks?)

DavidFair:
I don't think this has the updated .travis

If you do the following and commit the result it will correctly pick up coveralls:
```
git fetch origin
git checkout origin/develop .travis.yml
```

utils/test_settings.py
DavidFair: It could be the setup / tear down methods in unit tests which are responsible?
Then this just returns a path which is generated at runtime.

ElliotAOram:
This still need alot of corrections to make it robust and I'm not happy with the code quality either. As such it'll be going on hold for a while longer

ElliotAOram:
This is no longer required as it will be replace by #203 #204 #205

PR #54, opened 2018-05-08
Improve pylint script

ElliotAOram:
**Description of changes**
Improve the pylint script to fail the travis build if it fails
Also reduce code duplication in pylint script

---
Ensure that the package changes work when running the queues (on the dev nodes)


ElliotAOram:
Note: we need to ensure that the `sparse-checkout` on the nodes includes the `setup.py` file from the top level dir. This is as easy as add the line `setup.py` to the `.git/info/sparecheckout` file. I have done this on the linux dev node but need to make sure it's done elsewhere as well

PR #77, opened 2018-05-25
Improve QueueProcessor restart.sh

ElliotAOram:
**Description of changes**
* The restart.sh script should now be callable from any location
* If the queues are not running (can't find `.pid`) restart will start them

**How to test**
* Stop daemons and try to use restart to start them - this should work
* Try to restart the queues while it's running - this should work
* Try to use the script from any location (don't `cd` into `QueueProcessors` folder)

### DIFF
--- a/QueueProcessors/restart.sh
+++ b/QueueProcessors/restart.sh
@@ -9,7 +9,7 @@
 ROOT_DIR="$(dirname "$0")"
 
 ## Autoreduction Processor
-pkill -9 -f "python AutoreductionProcessor/autoreduction_processor_daemon.py start" &&
+pkill -9 -f "python .*AutoreductionProcessor/autoreduction_processor_daemon.py start" &&
 echo "Stopped autoreduction_processor_daemon.py";
 if [ -e /tmp/AutoreduceQueueProcessorDaemon.pid ]
 then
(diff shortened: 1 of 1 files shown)


QueueProcessors/restart.sh
```diff
@@ -6,18 +6,33 @@
 #### WARNING: If there's currently reduction runs "processing" this may 
 #### cause them to get stuck in the "processing" state.
 
+ROOT_DIR="$(dirname "$0")"
+
+## Autoreduction Processor
 pkill -9 -f "python AutoreductionProcessor/autoreduction_processor_daemon.py start" &&
```
jsws: This commit breaks this because this process is not found now. Need to add wild card `.*` infront of `autoreduction_processor_daemon.py` so it doesn't matter where the script is ran from.

QueueProcessors/restart.sh
```diff
@@ -6,18 +6,33 @@
...
+    echo ".pid file not found - starting process"
+fi
+python $ROOT_DIR/AutoreductionProcessor/autoreduction_processor_daemon.py start &&
 echo "Started autoreduction_processor_daemon";
 
 echo ""; # New line
 
+
+## QueueProcessor
 pkill -9 -f "python QueueProcessor/queue_processor_daemon.py start" &&
```
jsws: Same again here

QueueProcessors/restart.sh
ElliotAOram: Ah yes I missed these thanks.

PR #78, opened 2018-05-30
Improve logging in end of run monitor

### LINKED ISSUE #59: Improve logging in End of Run Monitor
Currently the logging in end of run monitor is sparse, and also contains lots of STOMP queue logging. 
Adding better logging will allow us to see how the End Of Run Monitor is performing.
For example:
- Log which instruments we're watching on startup
- Add some debugging to log

jsws:
**Description of changes**

- Improve logging in end of run monitor
- Puts the monitor_log.txt directly on the C drive.

**How to test**

Restart end of run monitor and check log for output.

Doesn't fix #59 as the STOMP will still be noisy logging, haven't got time to investigate and fix this currently.

### DIFF
--- a/EndOfRunMonitor/ISISendOfRunMonitor.py
+++ b/EndOfRunMonitor/ISISendOfRunMonitor.py
@@ -18,7 +18,7 @@ INST_FOLDER = r"\\isis\inst$\NDX%s\Instrument"
 DATA_LOC = r"\data\cycle_%s\\"
 SUMMARY_LOC = r"\logs\journal\SUMMARY.txt"
 LAST_RUN_LOC = r"\logs\lastrun.txt"
-LOG_FILE = r"monitor_log.txt"
+LOG_FILE = r"C:\monitor_log.txt"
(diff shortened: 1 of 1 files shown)


EndOfRunMonitor/ISISendOfRunMonitor.py
```diff
@@ -18,7 +18,7 @@
 DATA_LOC = r"\data\cycle_%s\\"
 SUMMARY_LOC = r"\logs\journal\SUMMARY.txt"
 LAST_RUN_LOC = r"\logs\lastrun.txt"
-LOG_FILE = r"monitor_log.txt"
+LOG_FILE = r"C:\monitor_log.txt"
```
ElliotAOram: Do we want logging to go to `C:\`? Would it be better to have it in the project somewhere e.g. `EndOfRunMonitor/monitor_log.txt`. This could be done with something like: `os.path.join(os.path.dirname(os.path.abspath(__file__)))`

Issue #70, opened 2018-05-22
Create master build script for autoreduction


ElliotAOram:
travis should be able to call a single file and this should build the full project.
This build script should cover:
* Checking for pre-installed dependencies
  *  Install dependencies if required (in the most appropriate way for the platform)
* Checking if this is a test installation or a run install (supply with cmd arg)
* If test then:
  * start any services if possible (e.g. activemq)
  * migrate settings files
  * setup and populate database
  * test client access
  * test data generation

ElliotAOram:
After a bit of thought we might want to change as many of sub scripts into python as possible.
Reason being is it's a platform independent way of running things mean we don't have to mess about with multiple copies of what is fundamentally the same code. 
There might be some cases were it is unavoidable to not use a sh/bat script but I think in the cases we can we should go for cross platform. 

Additionally perhaps a some thought int other possible build tools should be done (completely remove the need for ever having a bat/sh)

PR #80, opened 2018-05-31
Move project requirements to setup.py

ElliotAOram:
**Description of changes**
* Move project requirements to install requires

**How to test**
* Run `pip install -e autoreduction` (from parent directory of the project) and ensure that the  pip requirements are checked/ installed.
  * This should also work with `python setup.py install`
* Ensure that the build still passes on travis (with new build instructions)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,13 +1,2 @@
-Django==1.11.12
-django_extensions==2.0.7
-django-user-agents==0.3.2
-MySQL-python==1.2.5
 nose==1.3.7
 pylint==1.8.4
-python-daemon==2.1.2
-requests==2.18.4
-SQLAlchemy==1.2.7
-stomp.py==4.1.20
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,13 +1,2 @@
-Django==1.11.12
-django_extensions==2.0.7
-django-user-agents==0.3.2
-MySQL-python==1.2.5
 nose==1.3.7
 pylint==1.8.4
```
DavidFair: Is there any reason these requirements can't be satisfied by PIP?

ElliotAOram: The reason these are separate is because they are not required for project installation (they are only used in testing) as such I think it would be the wrong move to install them as default with the package. However, with the enhancements to setup.py coming soon in #70 it is possible we can move these to be installed with a developer installation.

PR #86, opened 2018-06-07
extend setup.py

### LINKED ISSUE #70: Create master build script for autoreduction

ElliotAOram:
**Description of changes**

* Remove multiple .sh/.bat scripts in favour of python scripts where possible
* Add several commands to the setup.py to allow for project build steps (e.g. build database, migrate test files etc.)
* Updated travis to run new script
* Pylint changes to multiple files
* update wiki pages

**How to test**
* Follow the new [installation guide](https://github.com/ISISScientificComputing/autoreduce/wiki/Installation-for-local-development)
* Try to use all the commands with the new `setup.py` 
  * `python setup.py help` will display all of these
* Try these out on both windows and linux - ideally in a clean environment to check the installation work correctly
* Check the logs `build.log` look reasonable
* Check the build runs successfully on travis

Fixes #70


ElliotAOram:
Known bug I'm currently resolving:
* [ ] Windows: "mysql -uroot" is nota command
* [ ] Windows: setup.py start - `error: [Error 193] %1 is not a valid Win32 application`

DavidFair:
[ ] Linux : The shebang is not POSIX compliant and fails to find a shell on my system. They should all use `#!/bin/sh` rather than `#!/usr/bin/env/sh` (such as in build/install/mantid.sh)

DavidFair:
Also could we move all the dependencies into `C:\autoreduce_deps\` or `/opt/autoreduce_deps` instead of using stand-alone folders. 
This way if we ever need to delete all the deps for testing, it's one top level folder.

jsws:
Documentation on [install externals](https://github.com/ISISScientificComputing/autoreduce/wiki/Installation-for-local-development#install-external-programs) isn't clear to me. `python setup.py help` doesn't give details on the `-s` option.

Also when I run `python setup.py externals` I'm getting `error: [Errno 2] No such file or directory. `

ElliotAOram:
@jsws I've just fixed the `error: [Errno 2] No such file or directory.` I'll improve the docs too

jsws:
`build.log` is overwritten rather than appended to. Is this what we want?

jsws:
Yeah that's a good point, leave it how it is. It was just my gut reaction that overwriting .log files is bad!

DavidFair:
I've attached a patch file with some changes to the installation scripts
[script_changes.txt](https://github.com/ISISScientificComputing/autoreduce/files/2106027/script_changes.txt)

- Adds a missing step, for adding upstream Mantid deps
- Sets the scripts to exit if any step fails (set -e)
- Prevents wget from spamming the current download progress using the -q flag
- Echo before we start downloading, so it doesn't appear the script has hung

DavidFair:
* [ ] `mkdir` does not consistently use the `-p` flag. If the directory already exists, this will fail

ElliotAOram:
I'm happy with this on windows now - Just going to check it on Ubuntu before giving it the go ahead

ElliotAOram:
Actually... batter idea let see if it works on a cloud instance

ElliotAOram:
All the options work on my cloud instance expect for the mantid install. I'll try take a better look at this today as it doesn't appear to even log anything useful (just seems to hang) - might be a download issue?

DavidFair:
Could we look at merging this and opening new issues?
The bulk of the issues addressed by this are resolved and we can pick up minor issues in future PRs

ElliotAOram:
Okay I'm happy to green flag this with David's changes. It's likely that we'll find other bits as we go, but the workflow still works with this issue and it is too large to keep adding to. Glad to be finally merging this one!

PR #98, opened 2018-06-12
Add validators to model and perform clean before saving

### LINKED ISSUE #91: Only try to delete resubmitted job, if it was successfully submitted
Currently the webapp tries to handle exception handling in `reduction_viewer\utils:140` by catching all exceptions, then deleting the new job handle.

If the new job fails to create (for example the model rejects a field) this will then try to delete a record that was never saved.

Some refactoring of the flow would resolve this. The whole try / catch block should be removed. Instead of calling `save` and hoping it will work we should instead use `clean` before attempting to `save`.

DavidFair:
**Description of changes**
- Adds validators to the model for better error reporting in the stack traces
- Reworks control flow so we don't attempt to save a record before a clean. 
- This could still fail a FK constraint on the DB side. So instead of attempting to delete the record (which will lose our trace) that wasn't saved, we let Django fail the entry.
- We let the database sort by version, instead of parsing each record and using max to achieve the same effect.
- We now catch specific exceptions, rather than the kitchen sink approach.

**How to test**
- Go to reduction_viewer\utils.py:112
- Change the line that reads: `run_version=last_version + 1` to `run_version='a'`.
- Check a sane stack trace and error message are printed
- Examine the stack trace, it should have failed on `full_clean()` instead of save

Fixes #91 

---

**To test**
* Log into the devolpement nodes
* `git checkout this-pull-request-branch`
* restart services on each node (information on how to do this can be found in the development documentation)
* Submit some runs with the manualsubmission.py script

Once you are happy, merge this pull request into develop, and update the development nodes to the devlop branch
```
> git checkout origin/devlop
> git fetch -p
> git pull origin/develop
```

### DIFF
--- a/WebApp/autoreduce_webapp/reduction_viewer/models.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/models.py
@@ -37,8 +37,8 @@ class ReductionRun(models.Model):
     graph = models.TextField(null=True, blank=True)
     message = models.TextField(blank=True)
     reduction_log = models.TextField(blank=True)
-    # Scripts should be 10,000 chars or less. The DB supports up to 4GB strings here
-    script = models.TextField(blank=False, validators=[MaxLengthValidator(10000)])
+    # Scripts should be 100,000 chars or less. The DB supports up to 4GB strings here
+    script = models.TextField(blank=False, validators=[MaxLengthValidator(100000)])
 
     # Date time fields
     created = models.DateTimeField(auto_now_add=True, blank=False)
--- a/WebApp/autoreduce_webapp/reduction_viewer/models.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/models.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.core.validators import MinValueValidator, MaxLengthValidator
 import autoreduce_webapp.icat_communication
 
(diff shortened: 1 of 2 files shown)


WebApp/autoreduce_webapp/reduction_viewer/models.py
```diff
@@ -20,34 +21,42 @@ class Status(models.Model):
...
+    admin_log = models.TextField(blank=True)
     graph = models.TextField(null=True, blank=True)
-    
     message = models.TextField(blank=True)
     reduction_log = models.TextField(blank=True)
-    admin_log = models.TextField(blank=True)
-    
-    retry_run = models.ForeignKey('self', on_delete=models.SET_NULL, null=True, blank=True)
+    # Scripts should be 10,000 chars or less. The DB supports up to 4GB strings here
+    script = models.TextField(blank=False, validators=[MaxLengthValidator(10000)])
```
jsws: Just a quick thing I noticed, currently the WISH script is 40,000 chars so maybe this should be more like 100,000?

DavidFair: Okay, I'm happy to raise the limit as required. I just don't want a 2b character entry to be accidentally written.

I'll put that change in now

PR #99, opened 2018-06-13
Remove saving to archive feature

### LINKED ISSUE #5: Remove ability to save reduced data to the ISIS archive
Saving reduced data to the ISIS archive is no longer a requirement since the implementation of CEPH.

The code that handles writing to the ISIS archive can be removed, making the application smaller and simpler. 

Origin: https://github.com/mantidproject/autoreduce/issues/365

jsws:
**Description of changes**

- Reduced data can no longer be saved on the ISIS archive, it all goes to CEPH
- This also means new instruments will not need adding to `ceph_instruments` in the queue processor settings files. The `excitation_instruments` setting still remains as their data is saved in a different format (No run number folders)

**How to test**

- Submit a non-excitation (defined in `QueueProcessors/AutoreductionProcessor/settings.py`)  instrument reduction job 
- Check it the reduced data gets saved with the correct directory strucure
    - Reduced data should be contained in a run number directory. e.g. `/instrument/GEM/RBNumber/RB1720007/autoreduced/84854/`
- Submit an excitation instrument reduction job
- Check the reduced data gets saved with the correct folder structure (No run number folders)
- Ensure `archive_directory` and `ceph_instruments` are removed from `QueueProcessors/AutoreductionProcessor/settings.py` and check the queue processor still works

Fixes #5

### DIFF
--- a/QueueProcessors/AutoreductionProcessor/post_process_admin.py
+++ b/QueueProcessors/AutoreductionProcessor/post_process_admin.py
@@ -218,11 +218,6 @@ class PostProcessAdmin(object):
 
             # Specify directories where autoreduction output will go
             reduce_result_dir = MISC["temp_root_directory"] + instrument_output_dir
-            if self.instrument not in MISC["excitation_instruments"]:
-                run_output_dir = os.path.join(MISC["temp_root_directory"],
-                                              instrument_output_dir[:instrument_output_dir.rfind('/') + 1])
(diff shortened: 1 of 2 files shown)


ElliotAOram:
Currently not possible to test this on dev. Blocked by #80

ElliotAOram:
This works and is good to go now

PR #113, opened 2018-07-16
Refactored QueueProcessor to use the utils library

gc-code:
QueueProcessor now makes use of the utils library.


QueueProcessors/QueueProcessor/queue_processor.py
```diff
@@ -424,96 +427,31 @@ def retry_run(reduction_run, retry_in):
...
+    activemq_client.connect()
+
+    # Register the event listener
+    conn = activemq_client.get_connection()    
+    listener = Listener(activemq_client)
+    conn.set_listener(consumer_name, listener)
+    time.sleep(0.5)
+
+    # Subscribe to the queues
+    destinations = ['/queue/DataReady',
```
ElliotAOram: We might want to move this to the `QueueClient`. I don't think it has much overhead and is probably required for 99% of cases in which we use `activemq`. Can you refactor the destination subscription to the client as well please and make sure it's called on initialisation.

gc-code: I'm not quite sure what you mean when you say "Can you refactor the destination subscription to the client as well please and make sure it's called on initialisation." Queue subscription is already directly invoked by the main method.

QueueProcessors/QueueProcessor/queue_processor.py
```diff
@@ -424,96 +427,31 @@ def retry_run(reduction_run, retry_in):
...
+    logger.info("Starting autoreduce queue connection")
+    # Connect to ActiveMQ
+    activemq_client = QueueClient()
+    activemq_client.connect()
+
+    # Register the event listener
+    conn = activemq_client.get_connection()    
+    listener = Listener(activemq_client)
+    conn.set_listener(consumer_name, listener)
+    time.sleep(0.5)
```
ElliotAOram: This sleep is not required here as it is now performed in the `QueueClient` [here](https://github.com/ISISScientificComputing/autoreduce/blob/master/utils/clients/queue_client.py#L57) as such it can be deleted.

QueueProcessors/QueueProcessor/queue_processor.py
ElliotAOram: So my thought behind this is that when we use `ActiveMQ`, (e.g. `QueueClient`) we will always want to subscribe to the various queues  (`'/queue/DataReady'`,`'/queue/ReductionStarted'`, etc). I can't think of a use case currently were we don't require queue subscription (expected some of the test cases.

As such we should move the functionality that is in this class to be in the QueueClient class instead. However, seeing as this change is probably going to affect all areas of the code base, on reflection this is probably out of scope for this issue. I'll create a new issue and assign you to it.

QueueProcessors/QueueProcessor/queue_processor.py
```diff
@@ -424,96 +425,31 @@ def retry_run(reduction_run, retry_in):
...
+    logger.info("Starting autoreduce queue connection")
+    # Connect to ActiveMQ
+    activemq_client = QueueClient()
+    activemq_client.connect()
+
+    # Register the event listener
+    conn = activemq_client.get_connection()
+    listener = Listener(activemq_client)
+    conn.set_listener(consumer_name, listener)
+    time.sleep(0.5)
```
ElliotAOram: This sleep is not required here as it is now performed in the QueueClient [here](https://github.com/ISISScientificComputing/autoreduce/blob/master/utils/clients/queue_client.py#L57) as such it can be deleted.

QueueProcessors/QueueProcessor/queue_processor.py
ElliotAOram: See created issue #114

PR #116, opened 2018-07-24
Check for sudo in externals install scripts

DavidFair:
*Cannot be merged before #86*

This now checks for sudo before trying to run the external installation scripts

**To Test**
- Attempt to run `python setup.py externals` without sudo
- Check the build.log gives sane output
- `sudo echo` to cache your credentials
- Attempt to run `python setup.py externals -s activemq` and ensure that installs
- Attempt to run `sudo python setup.py externals` and ensure all others install


ElliotAOram:
I'll test this on Friday as the SCD cloud is currently down for maintenance and I'd like to run this in a clean environment

PR #118, opened 2018-07-24
Apply pylint project wide

DavidFair:
**In Progress**

Switches on Pylint project wide. We currently print out the number of files with any warnings and the score from Pylint.

As it was intended to use a bitwise AND a lower returned score indicates a worse failure. Therefore 1 is fatal and 16+ is style. 

However to get us started I just keep a total score track in the file and we can look at a more long term solution in the future (or fix everything ;) )


build/tests/test_pylint.sh
```diff
@@ -1,40 +1,55 @@
...
-       Scripts/ActiveMQTests/test_stomp_activemq.py
-       Scripts/ManualSubmissionScript
-       Scripts/NagiosChecks/autoreduce_checklastrun.py
-       WebApp/autoreduce_webapp/autoreduce_webapp/
-       QueueProcessors/AutoreductionProcessor
-      "
-
-for file in $FILES
+
+errorThreshold=3413
```
ElliotAOram: Oh so we've only got a few to fix then!

ElliotAOram:
It looks as though travis is picking up some of our project dependencies and checking those too.  Specifically alot of the icat files.

Additionally we are flagging up a large amount of errors with the old webapp tests. We should just skip these files as they will proabbly be replaced at some stage (although that can be done in a separate PR)

PR #135, opened 2018-07-26
Ensure webapp is picked up in coverage

DavidFair:


ElliotAOram:
This will hopefully be the last time I am happy to make the test coverage lower! Good spot

PR #137, opened 2018-07-31
First system test

DavidFair:
Adds a new system test only for GEM
We will add new instruments in future PRs as part of #138 

To test:
Check the following works correctly locally
```
sudo python setup.py start
pytest -m systemtest
```
This should also skip the test on Windows


systemtests/test_end_to_end.py
```diff
@@ -0,0 +1,75 @@
...
+        send_data_to_queue("1", "GEM", data_loc, "123", self.queue_connection)
+        # check that the file has been successfully reduced in the db
+        time.sleep(5)
+
+
+def send_data_to_queue(rb_number, instrument, location, run_number, queue_processor_connection):
+    """
+    Packs the specified data into a dictionary ready to send to a processor queue
+    """
+    data_to_send = {'rb_number': rb_number,
```
ElliotAOram: General comment: If we always send data in the same format, can we handle this in the queue client maybe? This change would not belong in this PR, but just a thought as I remember writing almost identical code for this recently in the archive monitor

systemtests/test_end_to_end.py
```diff
@@ -0,0 +1,75 @@
...
+        self.data_archive_creator.make_data_archive(["GEM"], 17, 18, 2)
+        self.data_archive_creator.add_data_to_most_recent_cycle("GEM", ['GEM123.raw'])
+        self.data_archive_creator.add_reduce_file("GEM", TEST_SCRIPT)
+        self.data_archive_creator.add_reduce_vars_file("GEM", "")
+        # Send data to queue
+        data_loc = os.path.join(self.data_archive_creator.get_current_cycle_for_inst("GEM"),
+                                'GEM123.raw')
+        send_data_to_queue("1", "GEM", data_loc, "123", self.queue_connection)
+        # check that the file has been successfully reduced in the db
+        time.sleep(5)
```
ElliotAOram: This currently doesn't appear to be making any assertions, am I missing something?

ElliotAOram:
This is rather stale now and has proven by its magnitude that it should be split up to make this easier to test. #266 has been created to take care of some of the internal testing requirements but there may need to be additional PRs created for this. I will close this for now and reopen a cleaner version to take advantage of the various changes made since this PR was originally opened

PR #143, opened 2018-08-01
Data archive explorer

### LINKED ISSUE #139: Add common functionality for exploring data archive
At various stages of the project we want the ability to explore the data archive. This is mainly for the end of run and archive monitors, but could find use in other areas. For now we should add this to utils but with a view to move it to archive monitors after #11 if it has no use elsewhere. This should replace all of the search functionality in the `DataArchiveCreator`

ElliotAOram:
# Changes
The archive explorer is designed to be able to provide an easy way to obtain data from the isis data archive.
this allows us to remove a lot of the functionality and cached values that were previously stored in the data archive creator. This means that we can now use this searching functionality with both a real and a fake archive

# To test
* Ensure that the new unit tests for the archive explorer and file filter run and pass.
* Ensure the new files have 100% test coverage in coveralls
* [Creating a fake archive](https://github.com/ISISScientificComputing/autoreduce/wiki/Testing-and-validating#creating-a-fake-data-archive) locally. Create an `AchiveExplorer` and check it find all the value you expect. 

Fixes #139

### DIFF
--- a/utils/archive_explorer/tests/test_archive_explorer.py
+++ b/utils/archive_explorer/tests/test_archive_explorer.py
@@ -4,7 +4,7 @@ Test the archive explorer
 import datetime
 import os
 import shutil
-import time
+import tempfile
 import unittest
 
 from utils.data_archive_creator.data_archive_creator import DataArchiveCreator
@@ -24,10 +24,8 @@ class TestArchiveExplorer(unittest.TestCase):
         """
         Create the test output directory and the data archive directory
         """
-        cls.test_output_directory = os.path.join(os.path.dirname(os.path.realpath(__file__)),
-                                                 'test-output')
+        cls.test_output_directory = tempfile.mkdtemp()
         cls.archive_directory = os.path.join(cls.test_output_directory, 'data-archive')
-        os.makedirs(cls.test_output_directory)
 
     def setUp(self):
         """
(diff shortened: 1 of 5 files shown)


utils/archive_explorer/tests/test_archive_explorer.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+
+    test_output_directory = None
+
+    @classmethod
+    def setUpClass(cls):
+        """
+        Create the test output directory and the data archive directory
+        """
+        cls.test_output_directory = os.path.join(os.path.dirname(os.path.realpath(__file__)),
```
DavidFair: Could we instead use `tempfile.mkdtmp` for system tests [see](https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp). This avoids the potential of running into race conditions down the road

utils/archive_explorer/tests/test_archive_explorer.py
```diff
@@ -0,0 +1,151 @@
...
+        test_start_time = datetime.datetime.now()
+        actual = self.explorer.get_most_recent_run_since('GEM', test_start_time)
+        self.assertIsNone(actual)
+
+    def test_get_most_recent_run_single(self):
+        """ Test path valid most recent run """
+        expected = os.path.join(self.archive_directory, 'NDXGEM', 'Instrument', 'data',
+                                'cycle_18_2', 'GEM001.nxs')
+        test_start_time = datetime.datetime.now()
+        time.sleep(0.2)
```
DavidFair: Could we comment why this is needed

utils/archive_explorer/tests/test_file_filter.py
```diff
@@ -0,0 +1,140 @@
...
+class TestTimeFilterFiles(unittest.TestCase):
+    """
+    Test the functionality that filters files by modification time
+    """
+
+    def setUp(self):
+        """
+        Create test directory and populate it with some test files
+        """
+        self.test_output_directory = os.path.join(os.path.dirname(os.path.realpath(__file__)),
```
DavidFair: Ditto about tmp directory

ElliotAOram:
Note the `time.sleep` has been removed, but the underlying problem is not resolved. I have opened a separate issue to deal with it #148

utils/archive_explorer/tests/test_archive_explorer.py
ElliotAOram: Yeah this is a much nicer way to do this. I'll use this in future - thanks for the tip 😄

utils/archive_explorer/tests/test_archive_explorer.py
ElliotAOram: Can remove `time.sleep` in favour of timestamps but the underlying fix for this is too large for this issue. I'll create another

PR #156, opened 2018-08-06
Rename all queues to be pythonic and less confusing

### LINKED ISSUE #123: Pylint errors: QueueProcessors
Fix all the pylint errors in the following files and then update the score in travis to reflect this change:

```
./QueueProcessors/__init__.py
./QueueProcessors/QueueProcessor/__init__.py
./QueueProcessors/AutoreductionProcessor/__init__.py
./QueueProcessors/AutoreductionProcessor/timeout.py
```

ElliotAOram:
# Changes
Note this is a very significant change and needs to be thoroughly tested to ensure it does not break anything.

* Renamed `AutoreductionProcessor` -> `reduction_proecessor` - This is because this processor merely is the interface between the queue and the reduction step as such the `auto` part of the word should not be used as it is not automatic.
* Renamed `QueueProcessor` -> `flow_processor` - This has been renamed to `flow` as it is a processor designed to deal with project flow between queues e.g. `dataready` -> `reductionpending` -> `processing` -> `etc.`

# Test
* Ensure that the queue_processors are working. Best to do this with the manual submission scripts on the development env.
* Ensure the pylint issues have been resolved (see travis output)

**Note: I will update the score once the other - less volatile PRs have gone through. As such I will leave the in progress label on this for now**

Fixes #123


ElliotAOram:
After thinking about this, although this is a good direction for the queueprocessors, we need tests here before we should be considering a refactor. Closing this for archive

PR #158, opened 2018-08-08
Harmonise clients and Client settings

### LINKED ISSUE #157: Harmonise clients settings to single data structure
At the moment, the settings are inconsistent and sporadic, using different naming conventions for the same thing and independent log on functionality. To start with, we should move the settings to be a single class type (where some parts can be overridden if required. Then we should move to adapting client code to be generic also (this will be a separate issue)

ElliotAOram:
This pull request implements a structure for both clients and settings. These will be separated out into 2 PRs for easier review. 

Settings:
* The settings files now use the `ClientSettings` class this is a class that will:
  * Ensure the settings contain `[username, password, host, port]`
  * Ensure that all settings are strings
* The `ClientSettingsFactory` should be used to create new settings instances with `ClientSettingsFactory.create(settings_type=, username=, password=, host=,port=, kwargs=)`
  * This function ensure that a settings object of expected type is returned
  * This will also check that the correct kwargs are used to make the the settings object

Clients:
* The clients now use an abstract class to ensure that they have the same name for functionality e.g. `connect(), disconnect()`
* The test for the clients have been separated into different files
* The clients have been updated to use the new settings objects

# To Test
* Make sure all the clients work with the new settings:
  * Update settings.py to match test_settings.py (you can use `python setup,py test_settings` if you're happy for all your `settings.py` files to get overwritten)
  * Test that the webapp can still see the database (just load the webapp and check for data)
  * Test that the queue can still be used (use manual submission script) 

Fixes #157


ElliotAOram:
Don't merge until after release

ElliotAOram:
Fixed issue with int casting
starting the `queue_processor` seems to be a problem not attached to this PR (I just checked out dev and you can't start it there either. I'll look into this as top priority then we can try merging this afterwards

PR #159, opened 2018-08-08
Remove pylint checks for migrations files

ElliotAOram:
The migrations files are auto generated and hence should be ignored by pylint


ElliotAOram:
Only changes are to the pylint tests file so happy to self review/merge this

PR #163, opened 2018-08-15
Update activemq version

### LINKED ISSUE #161: Build steps should produce a failure state
If build steps fail, for example if activemq (or any other service) fail to install then we should fail immediately (return non zero build status) and print the installation log out to terminal. This is particularly for travis builds were instead of failing on a installation, the script continues and fails in testing when the tests for the service are run. It would be preferable to have an early exit instead.

ElliotAOram:
Updated the ActiveMQ version to ensure download does not fail (generate activemq executable name in settings file)

Fixes #161

### DIFF
--- a/build/commands/installs.py
+++ b/build/commands/installs.py
@@ -80,8 +80,9 @@ class InstallExternals(Command):
 
         for service in services_to_install:
             if install_service(service, BUILD_LOGGER) is False:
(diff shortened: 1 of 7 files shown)


ElliotAOram:
This is good to merge as the test coverage for this will be handled in a different PR

PR #164, opened 2018-08-16
DataArchiveCreator tests and documentation

### LINKED ISSUE #144: Add tests for the data archive creator
The `DataArchiveCreator` is currently un-tested. We need to as a minimum have unit test coverage for this. Even though this is designed as a test helper, it will most likely become a core part of development so it's best to ensure it does what we expect.

ElliotAOram:
Added tests and documentation to the `DataArchivecreator`
* Ensure that all the tests pass for the `DataArchiveCreator` and the test coverage has increased.
* Check the documentation for typos/errors

Fixes #144


ElliotAOram:
Only adding tests so this should not have any adverse affects

PR #166, opened 2018-08-20
Setup start QueueProcessors

ElliotAOram:
# Implementation
Allow `setup.py` to start the QueueProcessors.

This is not only useful as a developer using the system but should also help when it comes to tests that wish to start the `QueueProcessors`

# To test
* Run `ps aux | grep python` and locate the `process_id`s for the Autoreduction and QueueProcessor daemons. 
* For both processes perform: `kill -9 <proc_id>` where `<proc_id>` s the process id discovered in the above step.
* Run `ps aux | grep python` again and ensure that the neither of the QueueProcessors are running. 
* Run `python setup.py start -s queues` and ensure that the services are now running using `ps aux | grep python`

*Note: this should only work on linux - windows should simple state that the service can not be started on windows


ElliotAOram:
Note that this has increased the test coverage, but some of this is not well covered (e.g. the code has been inspected but not actually tested). This means that some functionality will have been checked that it does not fail - but not that is produces valid output

ElliotAOram:
This is too stale to pick up again properly from this point. Instead I will close this for now and comment in the issue that it has been closed and should be restarted

PR #167, opened 2018-08-20
Update modification time instead of sleep

### LINKED ISSUE #148: Remove need for sleeps in DataArchiveCreator (and tests that use it)
`DataArchiveCreator`, due to the way it has been implemented, has to use `time.sleep` to ensure that files are created before they are accessed. This issue mainly arises in tests when you do the following:
1. Get the current system clock time
1. Create file with `DataArchiveCreator`
1. Test if the file was created after the current system clock time we got in the first step

Note that without sleeps, the above is not always true. This is due to the way linux files are created via python and it does not guarantee they are synced with the system clock.

If instead we change the implementation to get the system clock time, create the file and change the file modification time to be that of the system clock, then we should have no issues with this going forward

ElliotAOram:
Small change to remove the need for sleeps in `DataArchiveCreator`

* Code review should be fine to test this

Fixes #148

### DIFF
--- a/utils/data_archive/data_archive_creator.py
+++ b/utils/data_archive/data_archive_creator.py
@@ -29,6 +29,7 @@ data_archive.delete_all_files()    # remove all files (not folders)
 data_archive.delete_archive()      # remove all files and folders
 
 """
+import calendar
 import os
 import shutil
 import time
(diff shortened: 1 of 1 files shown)


ElliotAOram:
Small change and only affects developer side so should be fine to merge

PR #181, opened 2018-08-30
Improve sql dump script and add minimal test case

ElliotAOram:
The SQL dump script was a bit too open before and made some assumptions. This is a cleaner class based script at should force the user to enter the data manually. Whilst this might be frustrating to have to enter this data manually, it should ensure that the user thinks about what they are doing (which is good as this deletes databases)


ElliotAOram:
This also matches the [documentation](https://github.com/ISISScientificComputing/autoreduce/wiki/Smoke-testing#database) correctly now

Issue #160, opened 2018-08-09
Move source code to a single package


ElliotAOram:
At some point we need to move our source code to a single package. By this I mean that the project structure should be:
* documentation
  * ...
* autoreduction
  * monitors
  * queue_processors
  * utils
  * Webapp
  * etc

This should not only make us more pythonic in structure, but also allow us the added bonus of being able to run static analysis over the whole repo more easily

ElliotAOram:
We should experiment with this at the beginning of the development period for 18.3. As it could have far reaching hidden implications we should do it at the beginning of dev cycle and have the whole cycle to find any issues

dtasev:
It's an ongoing slow discussion whether to go to a single package, or multiple repositories.

Regardless the structure above won't work well for Django, if we're going to a single package, we should make sure that the Django project is at the top level. That way we can get the ORM access everywhere and can remove all the orm.py stuff.

PR #192, opened 2018-09-28
Simplify imports

ElliotAOram:
### Changes
Simply the way we import some of our more commonly used classes by importing sub classes in some selected inits. These include:
* queue_proc_utils
* clients
* data_archive

### Test
* Perform a manual submission to validate that the queue is still working.
* Run the tests to ensure the imports are still working for the utils package

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ install:
 
   # ================== Test setup ==================== #
   - python setup.py test_settings
+  - cat utils/settings.py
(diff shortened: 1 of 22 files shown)


gc-code:
Are you sure you want to merge this directly into master?

ElliotAOram:
This has hidden repercussions that I was unaware of to begin with. As I don't understand it fully and this requirement will be eased by #160

Issue #205, opened 2018-10-16
End of run monitor - link up ICAT checker and service restart


ElliotAOram:
This is blocked by #203 and #204 

Once the above are complete we need to add a check in the End Of Run Monitor to:
1. Check the difference between the current run we are aware of in the EndOfRunMonitor and ICAT
2. If these have diverged : restart the End of Run Monitor.
3. Otherwise take no action

The code for doing the comparison should go [here](https://github.com/ISISScientificComputing/autoreduce/pull/207/files#diff-b0d6a5b80ff2b54c73a32a46e4c24cfdR35).
The result of this function should simple be:
* `True` (if the service is running okay and we do NOT require a restart)
* `False` (if the service has become out of sink and requires a restart)

Finally note that this health check runs in a separate thread and therefore should not interact directly with the end of run monitor

Issue #203, opened 2018-10-16
End of run monitor - ICAT checker


ElliotAOram:
Create a function (potentially in a separate file from EndOfRunMonitor) to query ICAT for the latest run for a given instrument. This should be easily callable from the EndOfRunMonitor and unit tested. We will need to use mocking to be able to validate it as we cannot use an ICAT connection on travis for security reasons.

PR #207, opened 2018-10-18
End of run monitor restart

ElliotAOram:
### Changes
* Add step into the end of run monitor windows service to allow for a restart of the service. 

This solution may have to be improved if we find that we are missing runs because of it. This could be a side effect of there being a file event (last_run.txt is updated) whilst the health check is taking place. If this is the case then we will have to have a separate Thread that monitors the service instead - this may infact be a separate windows service?
In order to do this the health check will have to be completely encapsulated from the end of run monitor and be purely external. If this can be incorporated into the implementation of #203 this would make the implementation of a separate windows service for monitoring much easier. 

### Test
* Point the end of run monitor to the data archive (overwrite `INST FOLDER` with  `INST_FOLDER = r'\\isis\inst$\NDX%s\Instrument'` in `monitors/settings.py`)
* Reinstall the end of run monitor - from command line (admin) type:
`python monitor_win_service.py install`
* In the Services tool for windows (type `Services` in start menu) locate the `Autoreduce Instrument Monitor` service.
* You will need to update the `Log On As` to an account with permission to view the archive
  * Right click the service and click properties
  * Navigate to the `Log On` tab
  * Check the `This account` section and put `clrc\<fed-id>` into the field and your password below
  * Click OK
* Right click the service again and click Start
* Once running go and check the `monitor.log` file located in `git-repo/logs/`
* You should see the standard start up text:
```
2018-10-18 09:57:26,992 Connecting to ActiveMQ...
2018-10-18 09:57:26,992 Connected to ActiveMQ
2018-10-18 09:57:27,005 Watching \\isis\inst$\NDXWISH\Instrument\logs
2018-10-18 09:57:27,019 Watching \\isis\inst$\NDXGEM\Instrument\logs
2018-10-18 09:57:27,026 Watching \\isis\inst$\NDXOSIRIS\Instrument\logs
2018-10-18 09:57:27,032 Watching \\isis\inst$\NDXPOLARIS\Instrument\logs
2018-10-18 09:57:27,039 Watching \\isis\inst$\NDXMUSR\Instrument\logs
2018-10-18 09:57:27,063 Watching \\isis\inst$\NDXPOLREF\Instrument\logs
```
* Wait around 15 seconds (current time out is set to 10 seconds but  interrupts from Wait for signal could cause slight delay)
* Refresh the log file (close and re-open)
* You should now see (obvious with the correct time):
```
2018-10-18 10:15:42,680 Performing Health check at 2018-10-18 10:15:42.681000
2018-10-18 10:15:42,680 No problems detected with service
```
* **Don't forget to the stop the service once you have finished testing else it will continue to poll the archive on your machine until it is told to stop!!!**


Anders-Markvardsen:
I observe "C:\Midlertiddig\autoreduce>c:\Python27\python.exe monitors\monitor_win_service.py install
Traceback (most recent call last):
  File "monitors\monitor_win_service.py", line 13, in <module>
    from monitors.health_check import HealthCheckThread
ImportError: No module named monitors.health_check"

Anders-Markvardsen:
Is it by design there doesn't seem to be n 'uninstall' option for python monitor_win_service.py.

How do you uninstall this service. For example 'sc delete AutoreduceInstrumentMonitor' from powershell does not have an effect.

I would prefer to remove eorm service entire for additional testing

Anders-Markvardsen:
above is testing from reducequeuedev

Anders-Markvardsen:
Notice python version on reducequeuedev is 2.7.14 whereas on my laptop 2.7.15. Mention this because of the unexplained behaviour we experienced yesterday with eorm refusing to start on my laptop yesterday and the remote possibility that this difference in version has something to do with this

Anders-Markvardsen:
For information: on reducequeuedev I get 

```
$ python setup.py test_settings
running test_settings
================== Migrate test settings ====================
error: [Errno 2] No such file or directory: 'C:\\EndOfRunMonitor-git\\WebApp\\autoreduce_webapp\\autoreduce_webapp\\test_settings.py'

```

Is this to because webapp is by purpose not checked out with this branch. If yes should there be a ref to this on wiki installations

Anders-Markvardsen:
on reducequeuedev manually copied test_settings.py to settings.py after pointing it to the archive and started eorm. It seems to work but stopped it because it e.g. picked up a osiris run 
```
2018-10-20 07:18:58,872 Data sent: {'instrument': 'OSIRIS', 'run_number': '00136739', 'data': '\\\\isis\\inst$\\NDXOSIRIS\\Instrument\\data\\cycle_18_2\\OSIRIS00136739.nxs', 'rb_number': '1820370', 'facility': 'ISIS'}
```
which then was not visible http://reducedev.isis.cclrc.ac.uk which make me worry that I may not be sending these to the production activemq rather than the dev activemq

will look at this later

Anders-Markvardsen:
The above makes sense from the code since health_check() always return true and health check only stops with stop signal and therefore will never print out "Main thread loop stopped". 

This may be OK for this PR?

Anders-Markvardsen:
@ElliotAOram see question above. Otherwise code and tests looks good.

ElliotAOram:
> Is it by design there doesn't seem to be n 'uninstall' option for python monitor_win_service.py.
> 
> How do you uninstall this service. For example 'sc delete AutoreduceInstrumentMonitor' from powershell does not have an effect.
> 
> I would prefer to remove eorm service entire for additional testing

I will investigate if that is possible but the implementation should probably be in a separate PR as this was an issue with this service before the changes made here

ElliotAOram:
> For information: on reducequeuedev I get
> 
> ```
> $ python setup.py test_settings
> running test_settings
> ================== Migrate test settings ====================
> error: [Errno 2] No such file or directory: 'C:\\EndOfRunMonitor-git\\WebApp\\autoreduce_webapp\\autoreduce_webapp\\test_settings.py'
> ```
> Is this to because webapp is by purpose not checked out with this branch. If yes should there be a ref to this on wiki installations

Is this is the reason an documentation about this should be added to the wiki

ElliotAOram:
> The above makes sense from the code since health_check() always return true and health check only stops with stop signal and therefore will never print out "Main thread loop stopped".
> 
> This may be OK for this PR?

I think this behaviour is acceptable for this PR but it should be checked that it doe not have side effects when we resolve #205

ElliotAOram:
@Anders-Markvardsen I have created #215 to address the uninstall issue for the windows service. Given the above comments, are you now happy with this?

PR #209, opened 2018-10-19
ICAT Checker for EoRM

### LINKED ISSUE #203: End of run monitor - ICAT checker

gc-code:
ICAT checker that accepts as input an element from the INSTRUMENTS list and returns the latest run number from that instrument according to ICAT. The list of instruments can be found in the test_settings.py of the monitors directory.

I also expanded logging so it includes the log level, file name and line number in both EoRM and the ICAT checker. I think this will make it easier to track down bugs.

A 'file_prefix' attribute has been added to the instrument dictionary because not all instruments prefix their files with the full name of the instrument. e.g. 'MERLIN' and 'MER'

## To Test

* Run the unit tests. No special configuration is required to do this.
* Run icat_monitor.py against a real ICAT. It should print out the latest run number for each instrument in the INSTRUMENTS list. This can be verified by looking at the last_run.txt of each instrument.

Fixes #203

### DIFF
--- a/monitors/end_of_run_monitor.py
+++ b/monitors/end_of_run_monitor.py
@@ -14,7 +14,9 @@ from monitors.settings import (INST_FOLDER, DATA_LOC, SUMMARY_LOC,
                                LAST_RUN_LOC, EORM_LOG_FILE, INSTRUMENTS)
 from utils.clients.queue_client import QueueClient
 
-logging.basicConfig(filename=EORM_LOG_FILE, level=logging.INFO, format='%(asctime)s %(message)s')
+logging.basicConfig(filename=EORM_LOG_FILE,
+                    level=logging.INFO,
+                    format='%(asctime)s %(levelname)s [%(filename)s:%(lineno)d] %(message)s')
 observer = Observer()  # pylint: disable=invalid-name
 
 
--- a/monitors/end_of_run_monitor.py
+++ b/monitors/end_of_run_monitor.py
@@ -11,10 +11,10 @@ from watchdog.events import FileSystemEventHandler
 from watchdog.observers import Observer
 
 from monitors.settings import (INST_FOLDER, DATA_LOC, SUMMARY_LOC,
-                               LAST_RUN_LOC, LOG_FILE, INSTRUMENTS)
+                               LAST_RUN_LOC, EORM_LOG_FILE, INSTRUMENTS)
 from utils.clients.queue_client import QueueClient
 
-logging.basicConfig(filename=LOG_FILE, level=logging.INFO, format='%(asctime)s %(message)s')
+logging.basicConfig(filename=EORM_LOG_FILE, level=logging.INFO, format='%(asctime)s %(message)s')
 observer = Observer()  # pylint: disable=invalid-name
 
 
--- a/monitors/icat_monitor.py
+++ b/monitors/icat_monitor.py
@@ -25,6 +25,8 @@ def get_run_number(file_name, instrument_prefix):
 def get_cycle_dates(icat_client):
     """
     What cycles could the last run have been in?
+    If the search space isn't constrained in some way then it takes far too long
+    to sort the list of data files. Narrowing down the dates is part of this.
     """
     date = datetime.datetime.today().strftime("%Y-%m-%d")
     logging.info("Getting nearest cycles to current date (%s)" % date)
@@ -44,7 +46,10 @@ def get_cycle_dates(icat_client):
 
 def get_instrument_run(icat_client, inst_name, cycle_dates):
     """
-    Returns the last run on the named instrument in ICAT
+    Returns the last run on the named instrument in ICAT.
+    Gets the list of investigations on the provided instrument within the
+    previously established cycle dates. The query then descends the investigation
+    tree until it reaches the files.
     """
     logging.info("Grabbing recent data files for instrument: %s" % inst_name)
     datafiles = icat_client.execute_query("SELECT df FROM InvestigationInstrument ii"
@@ -68,7 +73,7 @@ def get_instrument_run(icat_client, inst_name, cycle_dates):
 
 def get_last_runs():
     """
-    Retrieves the last run from ICAT on an instrument.
+    Retrieves the last run from ICAT for each instrument.
     """
     logging.info("Connecting to ICAT")
     icat_client = ICATClient()
@@ -77,7 +82,7 @@ def get_last_runs():
     # First, constrain the search space by getting recent cycle dates
     cycle_dates = get_cycle_dates(icat_client)
 
-    # Loop through the instruments, finding the latest run number for each
+    # Loop through the instruments, finding the last run number for each
     for instrument in INSTRUMENTS:
         run_number = get_instrument_run(icat_client, instrument['name'], cycle_dates)
         last_runs[instrument['name']] = run_number
--- a/monitors/icat_monitor.py
+++ b/monitors/icat_monitor.py
@@ -1,6 +1,5 @@
 """
-Monitor ICAT for the latest run on an instrument. If end of run monitor is out of sync
-then restart it.
+Monitor ICAT for the latest run on an instrument.
 """
 
 import datetime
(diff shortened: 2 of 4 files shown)


monitors/end_of_run_monitor.py
```diff
@@ -11,10 +11,12 @@
 from watchdog.observers import Observer
 
 from monitors.settings import (INST_FOLDER, DATA_LOC, SUMMARY_LOC,
-                               LAST_RUN_LOC, LOG_FILE, INSTRUMENTS)
+                               LAST_RUN_LOC, EORM_LOG_FILE, INSTRUMENTS)
```
ElliotAOram: Good readability change

monitors/end_of_run_monitor.py
```diff
@@ -11,10 +11,12 @@
...
 from utils.clients.queue_client import QueueClient
 
-logging.basicConfig(filename=LOG_FILE, level=logging.INFO, format='%(asctime)s %(message)s')
+logging.basicConfig(filename=EORM_LOG_FILE,
+                    level=logging.INFO,
+                    format='%(asctime)s %(levelname)s [%(filename)s:%(lineno)d] %(message)s')
```
ElliotAOram: Note that this format is being standardised in #206 [here](https://github.com/ISISScientificComputing/autoreduce/pull/206/files#diff-811c695e3e4c4625780187cf9cf2c1fbR5)
However as that is not merged yet this is fine to stay - we'll resolve merge conflict once we get round to it

monitors/icat_monitor.py
```diff
@@ -0,0 +1,107 @@
...
+
+def get_cycle_dates(icat_client):
+    """
+    What cycles could the last run have been in?
+    If the search space isn't constrained in some way then it takes far too long
+    to sort the list of data files. Narrowing down the dates is part of this.
+    """
+    date = datetime.datetime.today().strftime("%Y-%m-%d")
+    logging.info("Getting nearest cycles to current date (%s)", date)
+    last_cycle = icat_client.execute_query("SELECT c.startDate FROM FacilityCycle c"
```
ElliotAOram: You could reduce some code duplication here if you do something like:
```
find_cycle_query = "SELECT c.endDate FROM FacilityCycle c  WHERE {0} < c.startDate ORDER BY c.endDate {1} LIMIT 0,1"
last_cycle = icat_client.execute_query(find_cycle_query.format(date, 'DESC')
next_cycle = icat_client.execute_query(find_cycle_query.format(date, 'ASC')
```

ElliotAOram: Not for this PR, but I wonder if we find we have alot of common ICAT queries e.g. `find cycle` if we should store these together somewhere.

monitors/end_of_run_monitor.py
gc-code: Okay, it doesn't really matter. I just thought it would be good to have the level so you can easily search for 'ERROR'.

monitors/icat_monitor.py
```diff
@@ -0,0 +1,107 @@
...
+    if run_number:
+        logging.info("Found last run for instrument: %s", run_number)
+    return run_number
+
+
+def get_last_run(instrument):
+    """
+    Retrieves the last run from ICAT for an instrument
+    """
+    logging.info("Connecting to ICAT")
```
ElliotAOram: Coverage report shows that this is not unit tested.

monitors/icat_monitor.py
```diff
@@ -0,0 +1,107 @@
+"""
+Monitor ICAT for the latest run on an instrument. If end of run monitor is out of sync
```
ElliotAOram: `If end of run monitor is out of sync then restart it.` Is a misleading comment in this file as that will be done elsewhere

monitors/icat_monitor.py
gc-code: Notice that the two queries are slightly different. One is greater than, one less than. Also, the start date is used in one and the end date in the other for good reason. We don't want there to be a hard change in the dates when we cross into a new cycle, for example.

monitors/tests/test_icat_monitor.py
```diff
@@ -0,0 +1,82 @@
...
+import unittest
+import datetime
+from mock import Mock
+
+import monitors.icat_monitor as icat_monitor
+from monitors.settings import INSTRUMENTS
+
+
+# pylint:disable=too-few-public-methods
+class DataFile(object):
```
ElliotAOram: Is this the same form that is returned by icat?

monitors/tests/test_icat_monitor.py
```diff
@@ -0,0 +1,82 @@
+"""
```
ElliotAOram: Excellent tests with great use of mocking 👍

monitors/test_settings.py
```diff
@@ -11,10 +11,11 @@
...
-LOG_FILE = os.path.join(get_project_root(), 'logs', 'monitor.log')
-INSTRUMENTS = [{'name': 'WISH', 'use_nexus': True},
-               {'name': 'GEM', 'use_nexus': True},
-               {'name': 'OSIRIS', 'use_nexus': True},
-               {'name': 'POLARIS', 'use_nexus': True},
-               {'name': 'MUSR', 'use_nexus': True},
-               {'name': 'POLREF', 'use_nexus': True}]
+EORM_LOG_FILE = os.path.join(get_project_root(), 'logs', 'end_of_run_monitor.log')
+ICAT_MON_LOG_FILE = os.path.join(get_project_root(), 'logs', 'icat_monitor.log')
+INSTRUMENTS = [{'name': 'WISH', 'file_prefix': 'WISH', 'use_nexus': True},
```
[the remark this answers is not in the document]
gc-code: As I mentioned in the PR, MERLIN would be an example where the file prefix is MER. Just thinking about future expansion. For now it would be fine to keep the existing set up.

monitors/icat_monitor.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+logging.basicConfig(filename=ICAT_MON_LOG_FILE,
+                    level=logging.INFO,
+                    format='%(asctime)s %(levelname)s [%(filename)s:%(lineno)d] %(message)s')
+
+
+def get_run_number(file_name, instrument_prefix):
+    """
+    Extract the run number from a RAW or Nexus file
```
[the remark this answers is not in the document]
gc-code: I will get onto that.

monitors/test_settings.py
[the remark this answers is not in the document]
ElliotAOram: You're right apologies I missed this in the PR description

monitors/icat_monitor.py
ElliotAOram: Ah okay these are more subtle than I saw on first inspection

monitors/end_of_run_monitor.py
ElliotAOram: It is a good change. Just a heads that this may change again soon 😄

monitors/tests/test_icat_monitor.py
gc-code: Similar form. Python ICAT will return an object with the fields as its attributes. Obviously there will be more fields.

ElliotAOram: Okay cool this is good to know 👍 If we need to do more things like this in the future we should move this to be a test helper somewhere near the `ICATClient` but it's fine here for now as we don't have a reuse case just yet

monitors/icat_monitor.py
gc-code: Difficult to unit test because it encapsulates all the calls to the ICAT. Meaning that the ICAT client can't be effectively mocked (as far as I am aware).

ElliotAOram: I see where you are coming from. I will have a think about this

PR #220, opened 2018-10-26
ICAT checker and EoRM link

### LINKED ISSUE #205: End of run monitor - link up ICAT checker and service restart

gc-code:
Links the ICAT checker and end of run monitor. End of run monitor writes a CSV file with the last run number it has processed, which is then read by the health checker. The health checker then verifies that the runs present are the latest by checking with ICAT. If they aren't then the service is restarted.

I also added a test for icat_monitor.get_last_run.

Fixes #205

### DIFF
--- a/monitors/health_check.py
+++ b/monitors/health_check.py
@@ -35,13 +35,11 @@ class HealthCheckThread(threading.Thread):
         logging.info('Main Health check thread loop stopped')
 
     @staticmethod
-    def health_check():
+    def icat_check():
         """
-        Check to see if the service is still running as expected
-        :return: True: Service is okay, False: Service requires restart
+        Check the last run to be inserted into ICAT for each instrument
+        :return: True if the latest run in ICAT is the same or older than the one known by EoRM
         """
-        logging.info('Performing Health Check at %s', datetime.now())
-
         # Loop through the instrument list, getting the last run on each
         with open(EORM_LAST_RUN_FILE, 'rb') as last_run_file:
             last_run_reader = csv.reader(last_run_file)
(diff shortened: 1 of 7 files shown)


ElliotAOram:
Have you tested this for the case of EORM is writting to file and Health check is trying to read it? We might need a file lock for this

ElliotAOram:
Actually this should be fine providing all the read/write accesses to the file are performed inside a `try / except` worse case is we miss a health check if the timing is bad

monitors/health_check.py
```diff
@@ -31,13 +40,37 @@ def run(self):
...
+
     @staticmethod
     def health_check():
         """
         Check to see if the service is still running as expected
         :return: True: Service is okay, False: Service requires restart
         """
         logging.info('Performing Health Check at %s', datetime.now())
+        # Loop through the instrument list, getting the last run on each
+        with open(EORM_LAST_RUN_FILE, 'rb') as last_run_file:
```
ElliotAOram: This needs to be in a `try / except`

ElliotAOram:
Can we get test coverage over these lines (52-54): https://coveralls.io/builds/19743690/source?filename=monitors%2Fhealth_check.py#L52

I have a different PR that will get the rest of the test coverage for this file

gc-code:
> I wonder if a long term cleaner solution would be to (rather than checking EORM directly) we check the reduction database vs. ICAT for the difference.
> 
> The only issue this could raise is if the the message is in transit to the database at the time of checking - although this is fairly unlikely.
> 
> I believe this is the way the Nagios checks are designed (although to be on the safe side they have a tolerance of 3 runs de-sync)

I definitely agree with this. I have begun work on switching the health checker to the database, but I will keep this PR open just in case it proves too difficult to get done before next cycle.

A refactor of the code is required in order to reuse the database mapping created for the queue processor. This could complicate things.

PR #228, opened 2018-11-09
Resubmit missing runs from EoRM

ElliotAOram:
* Functionality in `ArchiveExplorer` to locate RB numbers for a given run.
* Function in `HealthChecker` to resubmit data that has been missed by EoRM
* Revert change on master from starting ICAT connection in `__init__`. There are occasions when you may want to initialise the client without the connection starting. Additionally it broke ICATClinet until tests

## To Test
* Code review
* Review of tests (these tests should cater for all use cases please consider if they do)

*As this uses ICAT this is going to be very difficult to set up a valid test environment for. However you can attempt the following:*
* Create a database that has a run for an instrument that is 3 ICAT e.g. `POLARIS113665`
* Create a fake archive and add a 4 data files  e.g. `POLARIS113665,POLARIS113666,POLARIS113667,POLARIS113668`
* Add entries in the `summary.txt` for all these files
* start the EndOfRunMonitor
* Observe in the logs and via activemq that all of these runs are resubmit


QueueProcessors/AutoreductionProcessor/autoreduction_processor.py
```diff
@@ -12,7 +12,8 @@
 from QueueProcessors.AutoreductionProcessor.autoreduction_logging_setup import logger
 # pylint:disable=no-name-in-module,import-error
 from QueueProcessors.AutoreductionProcessor.settings import MISC
-from settings import ACTIVEMQ
+# pylint:disable=relative-import
+from utils.settings import ACTIVEMQ
```
ElliotAOram: This should not be in this PR

monitors/icat_monitor.py
```diff
@@ -84,19 +84,58 @@ def get_last_run_in_dates(icat_client, instrument, cycle_dates):
...
+    """
+    Retrieve the location on disk of a file from ICAT
+    :param icat_client: ICAT client
+    :param instrument: Instrument name
+    :param run_number: File run number
+    :return: File location
+    """
+    datafiles = None
+    run_number_str = str(run_number)
+    for i in xrange(5):
```
ElliotAOram: Why 5?

monitors/health_check.py
```diff
@@ -70,12 +72,44 @@ def get_db_client():
...
+        return db_client
+
+    @staticmethod
+    def resubmit_run(icat_client, instrument, run_number):
+        """
+        Resubmit runs that are missing (have not been submitted by end of run monitor)
+        :param instrument: The instrument associated with the missing run
+        :param run_number: The run number to resubmit
+        :param limit: Where we expect the rb number to be located in the file
+        :param use_nxs: The expected extension of the data file
```
ElliotAOram: This docstring is now out of date

monitors/health_check.py
```diff
@@ -70,12 +72,44 @@ def get_db_client():
...
+        # Connect to the autoreduce queues
+        queue_client = QueueClient()
         try:
-            db_cli.connect()
+            queue_client.connect()
         except ConnectionException:
-            logging.error("Unable to connect to MySQL")
-        return db_cli
+            logging.error("Unable to connect to Queue")
+            return False
```
ElliotAOram: It might be worth opening a separate opening a separate issue to add functionality in the clients to perform a none throwing connect. I can imagine this is a fairly common issue that we'll potentially encounter more than just here.

utils/data_archive/archive_explorer.py
```diff
@@ -101,3 +101,21 @@ def get_most_recent_run_since(self, instrument, cut_off_time):
         if valid_files:
             return valid_files[-1]
         return None
+
```
ElliotAOram: Changes to archive explorer can be removed from this PR as it is no longer required due to querying ICAT

utils/data_archive/tests/test_archive_explorer.py
```diff
@@ -11,7 +11,7 @@
 from utils.data_archive.archive_explorer import ArchiveExplorer
 
 
-# pylint:disable=invalid-name
+# pylint:disable=invalid-name,missing-docstring,too-many-public-methods
```
ElliotAOram: Changes to Archive Explorer tests can be remove fro this PR as we are now using ICAT to query this info

monitors/test_settings.py
```diff
@@ -7,7 +7,9 @@
 from utils.project.structure import get_project_root
 
 # Config settings for cycle number, and instrument file arrangement
-INST_FOLDER = os.path.join(get_project_root(), 'data-archive', 'NDX%s', 'Instrument')
+ARCHIVE_BASE = get_project_root()
```
ElliotAOram: These changes may no longer be required now?

ElliotAOram:
@gc-code looks pretty good but we should remove the now redundant code I added which has been replaced with the ICAT querying. Additionally can you retarget this branch to merge with develop.

QueueProcessors/AutoreductionProcessor/autoreduction_processor.py
gc-code: This change is required in order for the build to pass.

ElliotAOram:
@gc-code changes look good and appear to work as described. Last thing to do is add test coverage for the parts we have added. the coverage report details this in the `New missed lines in diff` section.

QueueProcessors/AutoreductionProcessor/autoreduction_processor.py
ElliotAOram: I'm confused as to how this actually works. utils.settings doesn't have anything called `ACTIVEMQ` anymore it is now `ACTIVEMQ_SETTINGS`. I'll take a look at resolving this

PR #232, opened 2018-11-19
Allow mysql password with newer version of mysql

### LINKED ISSUE #223: Bug in mysql setup stage when using password
Bug observed on Windows 10 laptop.

To reproduce:

* Following Database setup steps on https://github.com/ISISScientificComputing/autoreduce/wiki/Setup
  * Update `build\settings.py` with password
  * run `python setup.py database`

This returns the error message

```
C:\Midlertiddig\autoreduction>python setup.py database
running database
==================== Building Database ======================
Setting up database on local host
Script did not complete. Check build.log for more details.
ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)
```

You can get a login by modifying mysql lines in `generate_database.py` to use `-pPASSWORD`. However it then fails at a later stage

Anders-Markvardsen:
Fixes #223

To test:

* Check that the Database setup works on https://github.com/ISISScientificComputing/autoreduce/wiki/Setup where mysql password is set 
* check code

### DIFF
--- a/build/database/generate_database.py
+++ b/build/database/generate_database.py
@@ -26,7 +26,7 @@ def run_sql_file(sql_file_location, logger):
     with open(sql_file_location, 'r') as input_file:
         password = ''
         if DB_ROOT_PASSWORD:
-            password = 'MYSQL_PWD=%s ' % DB_ROOT_PASSWORD
+            password = '-p%s ' % DB_ROOT_PASSWORD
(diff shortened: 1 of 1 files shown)


ElliotAOram:
Coverage has decreased for this, but only due to added more lines to a section of the code base that was not covered previously. as such I'm happy to merge this for the time being (the hope is to replace alot of these setup scripts with docker instead at a later stage)

PR #238, opened 2018-12-04
CEPH link in webapp

### LINKED ISSUE #236: Open CEPH directory from webapp
Something that could potentially be a large usability improvement is allowing users to open the folder where the reduced data is stored from the webapp. We would have to consider how this would work in an OS agnostic way and if it is a valuable enough change

ElliotAOram:
Add a link to the ceph data storage area for users to download their data. This link is currently available via the run summary page

### To test
* Open the webapp to the run summary page ensure that the link takes you to the right place (try this with several different runs) 

Fixes #236

### DIFF
--- a/QueueProcessors/AutoreductionProcessor/autoreduction_logging_setup.py
+++ b/QueueProcessors/AutoreductionProcessor/autoreduction_logging_setup.py
@@ -2,7 +2,7 @@
 import logging.handlers
 
 LOGGING_LEVEL = logging.INFO
-LOGGING_LOC = '/home/autoreduction/autoreduction/logs/autoreductionProcessor.log'
+LOGGING_LOC = '/home/isisautoreduce/logs/autoreductionProcessor.log'
 
 logger = logging.getLogger('AutoreductionProcessor')
 logger.setLevel(LOGGING_LEVEL)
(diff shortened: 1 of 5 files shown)


gc-code:
I've been looking into this and I don't think it will work as a link in the browser. We would have to provide it as a string that the user could copy and paste into their file explorer. Perhaps having it replace the existing reduced data path.

ElliotAOram:
@gc-code what would be wrong with having a link? If I'm not mistaken it is possible to link to CEPH storage directly in browser?

gc-code:
You're right, it does seem to be possible to view the files in a browser. It's just the particular link in your commit that didn't work because it used forward slashes instead of backslashes. The following link type seems to work across the major browsers: file://///data.analysis.stfc.ac.uk/rb

Site VPN access is still required, but I think the webapp is behind the site firewall anyway. I will implement this and see how it goes.

ElliotAOram:
@gc-code is ready to be reviewed again now?

gc-code:
Yes, I think it's the best we will be able to do with the security limitations on browsers.

ElliotAOram:
Thanks, I'll have a look today

PR #246, opened 2019-01-09
update travis to use pip install -e .

### LINKED ISSUE #247: Update dependencies
Currently autoreduction uses specific versions of alot of the packages it pip installs. We should attempt to update these to be most recent versions

ElliotAOram:
Allow pytest-cov to dictate what version of pytest we need. These 2 dependencies are closely tied so we should just let pytest-cov decided which dependency it requires. 

Additionally this removes the specified versions for dependencies. We should instead try to remain up to date and deal with issues as they arise

Fixes #247

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,9 @@ before_install:
   - mysql < build/database/test_db_setup.sql
 
 install:
+  # Install autoreduction
   - pip install -e .
+  # Install test suite dependencies
   - pip install -r requirements.txt
 
   # ================== Test setup ==================== #


.travis.yml
```diff
@@ -21,7 +21,7 @@ before_install:
   - mysql < build/database/test_db_setup.sql
 
 install:
-  - python setup.py install
+  - pip install -e .
```
Anders-Markvardsen: Could it be useful in include a short comment on who pip install -e . does as different from pip install -r requirements.txt?

Anders-Markvardsen:
looks good with avoiding specific mention of versions in travis, setup and requirement files, but wonder if it would be useful to have a comment in these to developers to "look at our install pages for packages version Autoreduction has been tested with"?

Issue #28, opened 2018-04-18
System test for autoreduction


ElliotAOram:
We should add a system test that uses all parts of the system. This will form the basis of regression testing and should prove useful for future plans to restructure the code base.

The test will not be able to test the full end to end code execution as the EndOfRunMonitor is a windows only service. Instead we should just send data to the queue in the expected formats. The test will then exercise the queue flows, data reduction and database submission. Due to this constraint, it is expected that the system tests will not run on windows. 

This test should be able to be run on both travis and locally

ElliotAOram:
blocked by #80

ElliotAOram:
This branch will have several merge conflicts to resolve after #86 these should be fixed first, then work can resume. I believe the first test case for GEM is almost working and it should be easy enough to adapt these for the other instruments. @DavidFair come talk to me when you move on to this and we can have a look at it's current state together

PR #267, opened 2019-02-19
DataArchiveCreator reduce.py and reduce_vars.py

### LINKED ISSUE #266: DataArchiveCreator to create reduce.py and reduce_vars.py
Issue raised by: [Developer]

### What?
It should be possible to add a reduction script to the fake data archive via the `DataArchiveCreator`. This is to replicate the `reduce.py` file that is maintained by instrument scientists at ISIS.

### Where?
[`DataArchiveCreator`](https://github.com/ISISScientificComputing/autoreduce/blob/master/utils/data_archive/data_archive_creator.py)

### How?
This is a sub issue that will reduce the size and scope of the System test pull request for autoreduction

Open a python interpreter and do the following:
```
import sys
sys.path.append(r'path/to/autoreduction/repo/utils/data_archive/data_archive_creator')
from data_archive_creator import DataArchiveCreator

# Create data archive
dac = DataArchiveCreator(r'path/to/location/to/create/archive')
dac.make_data_archive(['GEM'], 19, 19, 1)

# Create reduce script
dac.add_reduce_script('print(\'This is reduce.py\')')
dac.add_reduce_vars_script('print(\'This is reduce_vars.py\')')
```
At this point you should be able to see both the `reduce.py` and `reduce_vars.py` files present in the `data-archive\NDXGEM\user\scripts\autoreduce\` directory. Both these scripts should have the text specified above.

Final we should check the file are removed correctly:
```
dac.delete_all_files()
```
Ensure at this point the files are no longer present

ElliotAOram:
### Summary of work
* Added ability to create `reduce.py` and `reduce_vars.py` to a fake data archive via the use of the `DataArchiveCreator`
* Added the ability to search for `reduce.py` and `reduce_vars.py` in a data archive via the `ArchiveExplorer`

### How to test your work
* Code review
* Check the instructions in #266

### Additional comments
This is blocking continuation of #28 

Fixes #266

### DIFF
--- a/utils/data_archive/archive_explorer.py
+++ b/utils/data_archive/archive_explorer.py
@@ -101,3 +101,19 @@ class ArchiveExplorer(object):
         if valid_files:
             return valid_files[-1]
         return None
+
+    # ============================= Script Directories ===================================== #
+    def get_script_directory(self, instrument):
+        """ :return: archive_directory/NDXGEM/user/scripts/autoreduce"""
+        return self._file_path_exists(os.path.join(self.get_user_directory(instrument), 'scripts',
+                                                   'autoreduction'))
+
+    def get_reduce_file(self, instrument):
+        """ :return: archive_directory/NDXGEM/user/scripts/autoreduce/reduce.py"""
+        return self._file_path_exists(os.path.join(self.get_script_directory(instrument),
+                                                   'reduce.py').format(instrument))
+
+    def get_reduce_vars_file(self, instrument):
+        """ :return: archive_directory/NDXGEM/user/scripts/autoreduce/reduce.py"""
+        return self._file_path_exists(os.path.join(self.get_script_directory(instrument),
+                                                   'reduce_vars.py'))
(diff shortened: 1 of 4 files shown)


gc-code:
It is not currently possible to initialise a DataArchiveCreator object on an existing fake archive, the archive would need to be deleted and recreated. This is because it attempts to create the archive directory in the constructor. It would be useful to be able to add reduction scripts to an existing fake archive, as opposed to having to start all over again and create the instrument directories. This can be done by either removing directory creation from the constructor or implementing a test to see if the directory already exists.

ElliotAOram:
@gc-code I Agree
However, whilst this is a good idea I think it should be dealt with in a separate PR for 2 reasons:
1. It doesn't relate to this PR closely enough to be contained in it
2. It's actually slightly more work than an simple change as this will invalidate some of the other functions of the DAC - such as the ability to delete all the files in the archive. As this is done by holding a list of files rather than discovery. 

Created a new issue for this #268

gc-code:
OK

PR #275, opened 2019-02-21
Order database runs by run number

### LINKED ISSUE #273: Healthcheck must assess difference based on run_number
Issue raised by: [developer]

### What?
Currently the healthcheck in the End of Run Monitor queries the database for the most recent run by filtering date. This however, can cause a issue if you resubmit an older run manually. This will mean that health will resubmit all of the runs between the one resubmitted and the current one in ICAT.

### Where?
[This query needs to be updated](https://github.com/ISISScientificComputing/autoreduce/blob/master/monitors/health_check.py#L60)

### Reproducible?
Yes - Simply re-run an old file and wait for health check to kick in the re-run

* Submit an old run (atleast 5 run's previosu to the current) using the manual_subission.py script
* Wait for the health_check script to re-run. 
* Ensure that no new runs are submitted by health check

gc-code:
### Summary of work
Modified the health checker last run query to order by run number rather than created date. This causes problems when an old run is resubmitted, leading the health checker to wrongly assert that the database is behind ICAT.

### How to test your work
Resubmit an old run and check to make sure the health checker uses the latest run number.

Fixes #273

### DIFF
--- a/monitors/health_check.py
+++ b/monitors/health_check.py
@@ -63,7 +63,7 @@ class HealthCheckThread(threading.Thread):
             .join(db_cli.reduction_run().instrument) \
             .filter(db_cli.reduction_run().run_version == 0) \
             .filter(db_cli.instrument().name == inst) \
-            .order_by(db_cli.reduction_run().created.desc()) \
+            .order_by(db_cli.reduction_run().run_number.desc()) \
             .first()
         conn.commit()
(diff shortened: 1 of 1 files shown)


ElliotAOram:
Blocked by #276

ElliotAOram:
No longer blocked

PR #286, opened 2019-03-11
System test

### LINKED ISSUE #28: System test for autoreduction

ElliotAOram:
### Summary of work
* Added utility functions to start and stop the `QueueProcessors` and `activemq service`
* Updated test_settings.py files for various
* Added end to end test for `WISH` and test for `WISH` failure with invalid user script
* Changed autoreduction processor to use the same python binary that is executing the code to run sub processes
  * This is an error we had previously avoided by luck. In the case that a virtual environment is being used to run autoreduction python code, we should not be starting processing using "python" as the interpreter. This is because this will use the default system installation of python not the virtual environment were the code is running as such we can not grantee that autoreduction and it's dependencies are install on system python. 
* Fixed other tests to start and stop activeMQ as required

## How to test your work
**This can only be tested on a linux machine**
### Test the system test locally
* Set up the testing database `LOCALLY` - **DO NOT DROP THE DATABASE ON THE DEV OR PRODUCTION MACHINES**
```
$ mysql -u root 
> database drop autoreduction;
> exit
$ cd path/to/git/repo/for/autoreduction
$ mysql -u root < build/database/test_db_setup.sql
$ python setup.py database
$ mysql -u root < build/database/populate_reduction_viewer.sql
```
* migrate the test_settings
```
python setup.py test_settings
```
* Attempt to run the system tests. Note: these will unless you have sudo access for your current session. 
	* Ensure you have sudo access for the current session
```
$ sudo ls
```
* Attempt to run the tests again - they should pass!

### Test code changes on development environment
* log in to the compute node for the development environment 
```
$ git fetch-p
$ git checkout 28-system-test
$ git pull origin 28-system-test
```
* Once the code has been updated attempt to manually submit some runs using `scripts/manual_submission_script/manual_submission.py`
* Ensure runs successfully complete

Fixes #28


ElliotAOram:
This is now ready for review

ElliotAOram:
@gc-code I think I have addressed all the issues now
One thing to make sure of is that you re-run `python setup.py test_settings` after switching to the system test branch. I forgot this in the original PR description but it will result in the test failing as test_setting.py is no longer the same as develop/master

systemtests/test_end_to_end.py
```diff
@@ -0,0 +1,236 @@
...
+            self.database_client.disconnect()
+            self._delete_reduction_directory()
+            del self.data_archive_creator
+            # Done in tearDown to ensure run is removed even if test fails early
+            self._remove_run_from_database(self.instrument, self.run_number)
+
+        @classmethod
+        def tearDownClass(cls):
+            # Stop external services
+            external.stop_activemq()
```
gc-code: Don't stop ActiveMQ when finished. I would assume that it is already running like the rest of the test suite and remove both start and stop. I would like to be able to run the entire test suite together. I can see that you have tried adding start and stop to other test classes, but I am still getting lots of connection refused errors on AMQ.

systemtests/test_end_to_end.py
```diff
@@ -0,0 +1,236 @@
...
+
+        def setUp(self):
+            """ Start all external services """
+            # Get all clients
+            self.database_client = DatabaseClient(MYSQL_SETTINGS)
+            self.database_client.connect()
+            self.queue_client = QueueClient(ACTIVEMQ_SETTINGS)
+            self.queue_client.connect()
+            # Create test archive and add data
+            self.data_archive_creator = DataArchiveCreator(os.path.join(get_project_root()))
```
gc-code: Should look for existing data archive in case the tests failed last time.

systemtests/test_end_to_end.py
ElliotAOram: I would argue that the better end solution is to start and stop activemq for all tests that require it as this ensures tests are run in an isolated manner and do not interplay with one another. However, given the growing size of this PR it's probably best to do this in a separate issue now. I'll remove it from this PR.

gc-code: That sounds like a good idea. The AMQ start/stop should be handled by a separate issue.

systemtests/test_end_to_end.py
ElliotAOram: Similar to above this should be handled more globally by a wrapper class. However, (also like above) this large change should be in a different issue. For the time being I will add a check for this to see if an archive already exists

ElliotAOram:
@gc-code did you get this running locally on your ansible server? I had managed to fix this on Travis but hadn't validated it on my ansible instance yet.

gc-code:
Yes, I got it running fine.

PR #290, opened 2019-03-14
Empty Web application validation

### LINKED ISSUE #177: Webapp validation for empty
If the database has no data in `reduction_viewer_reductionrun` then it is very easy to get errors on the website. Some simple validation should correct this so that we either do not display links to places that would be invalid or, re-direct users to some message about no valid data could be found.

This is quite a rare state for the webapp, but it should be empty at the beginning of every cycle. 

Once this issue is resolved we can update the [smoke testing: database](https://github.com/ISISScientificComputing/autoreduce/wiki/Smoke-testing#database) wiki page to remove the need to add runs after flushing the database.

ElliotAOram:
### Summary of work
* Ensure that the Status records are made by the web app if they do not exist
* Add a mailto link to our support email in the case of there being no runs available

### How to test
* Ensure you are on a local machine or if testing on the dev environment no one is using it / needs any data on it
* Delete the local database:
```
$ mysql -u root -p
> drop database autoreduction;
> exit
$ mysql -u root < build/database/test_db_setup.sql
$ python setup.py database
```
to use the script to wipe the database:
```
$ python scripts/database_management/reset_database_post_cycle.py
```
Then follow the options to first backup, and then wipe the database

* Now load the webapp and ensure it does not error
* Instead you should see a page telling you there are no runs and given you an email link to the autoreduction support email

Fixes #177

### DIFF
--- a/WebApp/autoreduce_webapp/reduction_viewer/utils.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/utils.py
@@ -42,7 +42,7 @@ class StatusUtils(object):
         name or create one if it doesn't yet exist
         """
         # pylint:disable=no-member
-        status = Status.objects.get(value=status_value)
+        status = Status.objects.get_or_create(value=status_value)
         return status
 
     def get_error(self):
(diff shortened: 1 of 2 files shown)


ElliotAOram:
Agreed something doesn't seem right with this one. I'll take a better look again

ElliotAOram:
This work is invalidated by the work completed in #298

PR #293, opened 2019-03-19
Order instruments on run page alphabetically

ElliotAOram:
### Summary of work
* Make instruments list in alphabetical order

### How to test your work
* Load the webapp
* See that the front page (run list) has the instruments in alphabetical order

*No associated issue*


gc-code:
Works as described.

PR #300, opened 2019-03-29
New End of Run Monitor

### LINKED ISSUE #299: Development of new EoRM
Issue raised by: developer

### What?
To develop a new end of run monitor in accordance with the specification [here](https://github.com/ISISScientificComputing/autoreduce-documents/blob/master/brainstorming/prototype_eorm.svg).

The existing EoRM code should be reused to the extent possible.

The code should be easy to manually test, not requiring excessive setup like the current EoRM.  Modify the last run for an instrument and check that all the runs between the two values are correctly submitted.

Run the unit tests and ensure they all pass.

gc-code:
### Summary of work

Implementation of the new end of run monitor design.

### How to test your work

See updated README file for details on how to setup locally.

Unit tests can be run without any setup. Just type ```pytest test_new_end_of_run_monitor.py```

### Additional comments

I have made the cycle folder a setting that has to be manually adjusted. The reason I did this is to minimise the knowledge of the archive that's written into the code itself, this helps with testing because we don't have to go about setting up fake archives. At the same time, I thought it would be too cumbersome to have to adjust the cycle folder on every instrument in the CSV file, so I put it in settings.py. Let me know what you think is best.

I experienced difficulty with the EoRM logging config being overwritten by that of the queue client. That's why I have set up a logger in a slightly different way.

Fixes #299

### DIFF
--- a/monitors/README.md
+++ b/monitors/README.md
@@ -1,4 +1,32 @@
-# Monitors *(Windows Only)*
+# New End of Run Monitor *(Any Platform)*
+
+This End of Run Monitor can be run on any platform either standalone on the command line or
+as a scheduled task. All it requires is a CSV file of the following format:
+
+```
+WISH,1234,wish_lastrun.txt,wish_summary.txt,/archive/wish/data,.nxs
+GEM,1234,gem_lastrun.txt,gem_summary.txt,/archive/gem/data,.nxs
+```
+
+In order, the fields are: instrument name, last run number observed on the instrument, location
+of this instrument's lastrun.txt, location of this instrument's summary.txt, data location
+(parent of cycle folder), file extension of data to reduce. The last observed run is updated
+whenever EoRM submits a new run.
+
+There are two settings in settings.py for this script:
+
+* LAST_RUNS_CSV - Location of the CSV file described above
+* CYCLE_FOLDER - Folder for the current cycle in the instrument data directory.
+
+## Production Configuration
+
+In production, it is recommended to run this script as a Cron job on Linux or using the task
+scheduler on Windows. The period should be a minute or less so that runs are processed
+as soon as they arrive.
+
+The cycle folder in settings.py must be updated at the beginning of each new cycle.
+
+# Original End of Run Monitor *(Windows Only)*
 
 Scripts used to check for new runs from the instruments at ISIS.
 
--- a/monitors/README.md
+++ b/monitors/README.md
@@ -8,10 +8,10 @@ WISH,1234,wish_lastrun.txt,wish_summary.txt,/archive/wish/data,.nxs
 GEM,1234,gem_lastrun.txt,gem_summary.txt,/archive/gem/data,.nxs
 ```
 
-In order, the fields are: instrument name, last run number observed on the instrument, location
-of this instrument's lastrun.txt, location of this instrument's summary.txt, data location
-(parent of cycle folder), file extension of data to reduce. The last observed run is updated
-whenever EoRM submits a new run.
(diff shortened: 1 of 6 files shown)


monitors/README.md
```diff
@@ -1,4 +1,32 @@
-# Monitors *(Windows Only)*
+# New End of Run Monitor *(Any Platform)*
+
+This End of Run Monitor can be run on any platform either standalone on the command line or
+as a scheduled task. All it requires is a CSV file of the following format:
+
+```
+WISH,1234,wish_lastrun.txt,wish_summary.txt,/archive/wish/data,.nxs
```
ElliotAOram: Is there a reason we are using a CSV file here and not a settings.py file? This seems like it would be more inline with the rest of the code base

monitors/README.md
```diff
@@ -1,4 +1,32 @@
...
+* LAST_RUNS_CSV - Location of the CSV file described above
+* CYCLE_FOLDER - Folder for the current cycle in the instrument data directory.
+
+## Production Configuration
+
+In production, it is recommended to run this script as a Cron job on Linux or using the task
+scheduler on Windows. The period should be a minute or less so that runs are processed
+as soon as they arrive.
+
+The cycle folder in settings.py must be updated at the beginning of each new cycle.
```
ElliotAOram: This is fine for the time being but in future we should assess this automatically and cache the data - just so we can remove another manual process to remember each cycle

monitors/new_end_of_run_monitor.py
```diff
@@ -0,0 +1,215 @@
...
+        return line_parts
+
+    def read_rb_number_from_summary(self, run_number):
+        """
+        Loads the summary file and reads off the experiment RB number
+        :param run_number: Run number to lookup
+        :return: Experiment RB number
+        """
+        # Detect run number as a substring
+        with open(self.summary_file, 'rb') as summary_fp:
```
ElliotAOram: What's the significance of `fp` suffix here?

monitors/README.md
gc-code: So that the last run can be updated each time a new run arrives, this would be much more difficult to do in a settings.py file. If I just stored the last run number in a text file then we would have the issue of matching up keys across two files, which I think is more complicated than just storing everything in CSV.

monitors/new_end_of_run_monitor.py
gc-code: Stands for 'file pointer' from the C world. I wanted to distinguish it from the file name which is contained in summary_file.

gc-code: I'll change this because it's not very clear.

monitors/tests/test_new_end_of_run_monitor.py
```diff
@@ -0,0 +1,215 @@
...
+from monitors.settings import (CYCLE_FOLDER, LAST_RUNS_CSV)
+from monitors.new_end_of_run_monitor import (InstrumentMonitor,
+                                             get_prefix_zeros,
+                                             FileNotFoundError,
+                                             InstrumentMonitorError,
+                                             update_last_runs,
+                                             main)
+
+# Test data
+SUMMARY_FILE = ("WIS44731Hayden,Waite,"
```
ElliotAOram: Can you change the names on this test data - Assuming this is actually a real person right now?

ElliotAOram:
Nice work on this. I think we should start out using this next cycle on the development env. then on maintenance day (providing it has a stable first half of cycle) move it to production 👍

gc-code:
Sounds like a good plan.

PR #304, opened 2019-04-17
Handle ENGINX summary format

### LINKED ISSUE #303: Handle ENGINX format summary file in new EoRM
Issue raised by: developer

### What?

The new end of run monitor is currently unable to operate on instruments where the summary file truncates each run number. Presumably this is done for human readability. This also means that the same run number can occur multiple times within the summary file.

### Where?

Summary files for ENGINX, OSIRIS, POLARIS and POLREF have been found to miss the first few digits off of each run number.

### How?

Iterate over the summary entries from the bottom of the file. Extract the run number from each row and look for it as a sub-string in the full run number.

### Reproducible?
Yes

Test that the new end of run monitor works on the instrument list above.

gc-code:
### Summary of work
Now searches from the bottom of the summary file. It also looks for the summary run number as a sub-string in the full run number.

### How to test your work
Attempt to submit runs for ENGINX or OSIRIS using the new end of run monitor.

Fixes #303

### DIFF
--- a/monitors/new_end_of_run_monitor.py
+++ b/monitors/new_end_of_run_monitor.py
@@ -60,6 +60,39 @@ def get_prefix_zeros(run_number_str):
     return zeros
 
 
+def is_integer(int_str):
+    """
+    Test to see if the provided string is an integer
+    :param int_str: Integer as a string
+    :return: True if the string represents an integer
+    """
+    try:
+        int(int_str)
+        return True
+    except ValueError:
+        return False
+
+
+def extract_run_number_from_summary(first_part):
+    """
+    Look for the run number in the provided summary entry
+    :param first_part: First part of the summary entry
+    :return: Run number as printed in the summary file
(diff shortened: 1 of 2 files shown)


monitors/new_end_of_run_monitor.py
```diff
@@ -60,6 +60,39 @@ def get_prefix_zeros(run_number_str):
     return zeros
 
 
+def is_integer(int_str):
```
ElliotAOram: In a separate issue later on we should refactor this and other similar functions to somewhere that more generic so that other areas of the code base can potentially make use of very generic functions such as this

ElliotAOram:
Tests run locally on Windows and Linux

ElliotAOram:
Travis did run by webhook failed hence no travis on this PR: https://travis-ci.org/ISISScientificComputing/autoreduce/builds/521248547

PR #305, opened 2019-05-02
Fixed bug with timed reruns only reaching 'Queued' state

### LINKED ISSUE #272: Reruns only get Queued
Issue raised by: [developer]

### What?
When a run fails, a re-run will automatically be re-queued. This re-queued run appears in the database, but never actually gets run instead it gets stuck in the `Queued` state

### Where?
This is most likely in the `QueueProcessors`. It will take a bit of tracking down to find exactly what is going on. I suggest looking at the log files for extra information on this. As well as searching in the code for when failed runs get rerun.

### Reproducible?
Yes - this is consistently happening on production

It might be possible to deliberately force a failed run on the development environment. Whilst not all the runs have been showing the re-run behaviour [some have](http://reducedev2.isis.cclrc.ac.uk:8080/runs/ENGINX/302612/1/). Once it has been assessed how to do this, then with the new change, resubmitting a run that will fail. This should result in a new run being submitted that does actually run correctly and not remain in the queued state indefinitely.

gc-code:
### Summary of work
Fixed bug where timed reruns would trigger a Python exception due to an import error. This would lead to the run getting stuck in the queued state.

### How to test your work
Force a timed rerun by adding the following line to post_process_admin.py _send_error_and_log

```
self.data['retry_in'] = 1
```

Then submit a run that fails. You will see on the master branch that it just remains in the queued state, whereas on this branch the run goes through and repeatedly errors.

### Additional comments

Simple fix after finding the relevant log lines and a way to reproduce.

Fixes #272

### DIFF
--- a/QueueProcessors/QueueProcessor/queue_processor.py
+++ b/QueueProcessors/QueueProcessor/queue_processor.py
@@ -444,7 +444,6 @@ class Listener(object):
             MessagingUtils().send_pending(new_job, delay=retry_in * 1000)
         except Exception as exp:
             logger.error(traceback.format_exc())
-            new_job.delete()  # pylint: disable=no-member
             raise exp
 
 def setup_connection(consumer_name):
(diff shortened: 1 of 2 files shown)


gc-code:
I removed the line `new_job.delete()` in the exception handling code because it also produced an error. See the issue for more information.

PR #308, opened 2019-05-08
New Webapp interface

### LINKED ISSUE #298: Implement new Webapp user interface
Issue raised by: [developer]

### What?
Once the design work has been completed and agreed upon in #297, We must implement the proposed changes.

### Where?
Front end facing web app

* Open the webapp (this should take you to the index page by default)
* Observe that the front page matches that of the new design
* Ensure all buttons work and lead where you would expect them to

*Add extra criteria once design is complete

ElliotAOram:
### Summary of work
* Added Debug toolbar - this should only be rendered in debug mode 
* Added landing page `/overview`
* Set `/overview` to be the first page after redirect or accessing index
* Created new instrument pages that show variables, run details and status
* Added custom Pagination for run navigation

#### Test Index
* Boot the web app and ensure the landing page is `/overview`

#### Test Instrument pages
* Go to an instrument page
* Try everything that you can break things:
  * Ensure links work
  * Ensure pagination works
  * Ensure filtering works
  * Ensure drop down menus work for instrument variables
  * Ensure drop down menus work for experiments

Fixes #298


WebApp/autoreduce_webapp/utilities/pagination.py
```diff
@@ -0,0 +1,156 @@
...
+    Specific implementation for rendering the pagination if sorting by Run Number
+    """
+
+    def set_start_and_end(self):
+        """
+        Custom function to set the start and end variables and
+        construct the display name for the page
+        """
+        self.start = self.records[0].run_number
+        self.end = self.records[-1].run_number
```
gc-code: I like that runs are now sorted by run number instead of date, but reruns are printed in reverse order, with the latest rerun appearing at the bottom. Is there a way that reruns can be sorted such that the latest appears first?

WebApp/autoreduce_webapp/utilities/pagination.py
```diff
@@ -0,0 +1,156 @@
...
+        self._set_next_and_previous()
+        self.current_page = self.page_list[self.current_page_index - 1]
+        self._create_display_list()
+
+    def _validate_current_page(self):
+        """
+        Ensure that the current page specified is valid
+        Update the page to min/max if outside of expected range
+        """
+        page_count = len(self.query_set) / self.items_per_page
```
gc-code: There seems to be an off by one error somewhere around here that stops me from accessing the last page in a list. On my VM webapp, I set the page size to 10 and attempt to click the final run range in the list, the link just returns me to the penultimate run range.

setup.py
```diff
@@ -16,6 +16,7 @@
 from build.commands.start import Start
 
 setup_requires = ['Django',
+                  'django-debug-toolbar',
```
gc-code: The webapp is now dependent on the debug toolbar being installed to run. Is it possible to have this as an option for developers, or is it integrated into the code itself?

WebApp/autoreduce_webapp/templates/instrument_summary.html
```diff
@@ -1,48 +1,175 @@
...
+            {% if filtering == 'run' %}
+                <div class="row">
+                    {% if paginator %}
+                    <div class="col-md-12 text-center">
+                        <div class="btn-group" role="group">
+                            {% if paginator.has_previous %}
+                                <button title="First Page" class="btn btn-default" onclick="changePage(1)"><i class="fa fa-step-backward"></i></button>
+                                <button title="Previous Page" class="btn btn-default" onclick="changePage({{ paginator.previous_page_index }})"><i class="fa fa-chevron-left"></i></button>
+                            {% endif %}
+                            {% for page in paginator.display_list %}
```
gc-code: It seems like some of this processing could be removed from the template. A method called get_visible_page could be implemented or similar.

WebApp/autoreduce_webapp/utilities/pagination.py
ElliotAOram: I believe this should be fixed with the changes I just added. I was relying too heavily on python integer division - now using float division and `math.ceil`.

gc-code: This is now fixed.

WebApp/autoreduce_webapp/templates/instrument_summary.html
ElliotAOram: Which code are you thinking can be removed? I'm not sure that I completely follow.

Given that the display list is length of 1 + page_tolerance * 2 where page tolerance defaults to 3, it should never be processing more than a list of length 9. 

I think it could potentially reduce readability more than it's worth to performance  by changing this - unless I misunderstand what you are considering changing?

WebApp/autoreduce_webapp/utilities/pagination.py
ElliotAOram: This was actually put in by mistake. I agree that it is nicer, however we still need the date option as this is important for re-runs. e.g. if I re-run a run I expect it to be at the top of the page. 
I'll make the above amendment secondary to sort by version number (ascending), but I'll also add an option to sort by date too

ElliotAOram:
@gc-code with the exception of the date filtering, this is now ready to be reviewed again

gc-code:
All looking good to me. Once you resolve my last comment I will change my review to approve.

Anders-Markvardsen:
Suggestions for change:

1. When 'Filter by ...' RB number the Sort by options should be Date and Number (not Run number)
2. When 'Filter by ...' Run Number is suggested that likewise Sort by is Date and Number
3. To me it looks odd that 'Items per page' only visible when Filter by = Run Number
4. I suggest Sort by goes before Item per page because I feel the former is more important than the latter to the user
5. On GEM it is confusing what Items per page means versus the Pagination. There clearly two items but the Pagination says 1-1
6. A a nice to have MUSR Show Variables indicate with arrow down that it can be expanded when it can't
7. When click on MUSR Configure New Jobs I get error

```
IntegrityError at /instrument/MUSR/variables/
(1451, 'Cannot delete or update a parent row: a foreign key constraint fails (`autoreduction`.`reduction_variables_runvariable`, CONSTRAINT `reduction_variables__variable_ptr_id_9e17ca11_fk_reduction` FOREIGN KEY (`variable_ptr_id`) REFERENCES `reduction_variables_varia)')
```

ElliotAOram:
6. won't be completed before this release as it will involve some javascript trickery
7. is an issue relating to test database set up.

ElliotAOram:
Ready for review again

ElliotAOram:
@gc-code did you check if a hard refresh of the page fixed the formatting issue on the top row of the run table?

ElliotAOram:
Discussed point 2. No longer appears to be an issue

> A few things to improve:
> 
> * The top row of the run table appears in smaller font than the rest.
> * ENGINX is very slow to load on my VM webapp. Switching between pages takes about 5 seconds when the page size is set to 10. This may be because there are a large number of runs in error and processing state for ENGINX. It takes substantially longer to load than the old webapp. I believe this is something you have introduced since I last tested the new webapp.
> * Run number isn't as meaningful in the pagination control when filtering by date. Could you modify it so that it displays the date range of the runs on the page?

gc-code:
Points 1 and 2 can be considered solved.

ElliotAOram:
@gc-code @Anders-Markvardsen this can be reviewed again

gc-code:
Can you please remove the reduction variables to make load times faster. I would like to see that before reviewing again.

gc-code:
After your recent commits, when I select a page in the new webapp other than the first page, I get an 'instrument not found' error and no runs are shown.

It would also be good if the 'See instrument variables' button could be made flush with the other buttons, rather than appearing right aligned below the status box.

gc-code:
Setting the sort combo box to a value resolves the first issue. This leads me to believe the 'sort' URL parameter being empty may be the source of the problem. The URL looks like this when a run page fails as described above:

http://example.com/instrument/WISH/?page=3&filter=run&pagination=50&sort=

ElliotAOram:
@gc-code above changes should resolve the issues you were seeing

gc-code:
It is working well but I don't like the fact that popups appear all over the site when there are errors. I would rather click on individual runs to find out the cause of an error. If there are enough errors then they bury the run table, making it difficult to get to the information I want to see.

ElliotAOram:
@gc-code given that the popups scenario is also true for the current webapp, might I suggest that change belongs in a separate PR? Unless this is something different?

PR #313, opened 2019-05-13
Enginx event mode

### LINKED ISSUE #301: ENGINX: Event mode
Issue raised by: [user]

### What?
EnginX currently do not support running in Event mode. As such they would like to ensure event files do not run and instead autoreduction provides warning like: "Run skipped because Event mode". However this should still be on the webapp to ensure there is no confusion with missing runs in the job queue

### Where?
* Web application to render "Skipped" response - Reduction job page
* Capture mantid output for if Event file - AutoreductionProcessor
* Add new status type "Skipped" - Database 

### How?
Requirement raised during Feedback meeting with EnginX

* Submit an EnginX event data file to autoreduction
* See that the file is given the "Skipped" in the database
* See that the run in rendered with the Skipped tag in the web application from the run summary page (list of all runs) as well as from the single run page.

ElliotAOram:
### Summary of work
* Ensure that a mechanism exists in the queue processors to enable runs to be skipped
* Ensure that this is updated correctly in the database
* Ensure that skipped runs are displayed correctly in the webapp

### How to test your work
* Checkout this branch on the development environment for the compute node. 
* Submit a run to the development environment that is an Event mode EnginX file (`304652.nxs-304662.nxs`)
* Ensure the run is processed and the result is a run that is in the "Skipped" Status.
* Ensure that this renders correctly in the webapp 

Fixes #301


ElliotAOram:
Slight drop in test coverage which in this case I believe is okay. The reason being is that I have added code to a currently untested function (due to it's complexity and length). The additions I have made are relatively safe and attempting to add a unit test around this function would extend this PR way beyond scope of the issue.

There are already several issues created around testing the queues and testing this functionality will be included once we do those issues.

gc-code:
Could the creation of the 'reduction_skipped' queue be included in the test setup? I am attempting to test on my VM queue processor and it is failing because the queue does not exist. If we are looking to test this in the future then it would need to be a part of the test setup.

ElliotAOram:
What is the error you are getting? As I understand it if a queue in acrivemq does not exist if you send data to it it will create it. Is this a problem when you subscribe?

ElliotAOram:
@gc-code I don't see this on the development environment (I have just done a manual submission of EnginX 304652) from my local machine to the dev environment. http://reducedev2.isis.cclrc.ac.uk:8080/runs/ENGINX/304652/0/

Would you be able to take another look to confirm you can still reproduce this? And if so may you can get me some log information so I can work out where it is getting stuck? Thanks

PR #321, opened 2019-06-06
Credentials file for services

### LINKED ISSUE #320: Harmonise settings entry point for webapp and utils
Issue raised by: [developer]

### What?
We should not have to enter user/password/host/etc. for external services such as MySQL, ICAT, ActiveMQ into both the `utils/settings.py` and the webapp `settings.py`. Instead it would be better to have a single configuration file that does both. I can't think of a situation when these settings would not be the same at a given time.  

### Where?
* New credentials file in utils directory
* May require a intermediate file to process the settings if they are just key/value pairs
* Import these and update test_settings files for WebApp / utils accordingly

* Enter valid credentials into the `utils/services.credentials`
  * I would suggest trying this with local settings first then development settings
*  Run the webapp: `python Webapp/autoreduce_webapp/manage.py runserver`
* Check it is using the expected settings
* Use manual submission to submit a run
* Ensure this run appears in the webapp (the 2 are using the same settings)

ElliotAOram:
### Summary of work
* Add an `.ini` file that holds all the credentials for services
* update build script to move test `.ini` file

### How to test your work
* Run `python setup.py test_settings` from the root directory of the project
* Enter valid credentials into the `utils/credentials.ini`
  * I would suggest trying this with local settings first then development settings
* Run the webapp: python Webapp/autoreduce_webapp/manage.py runserver
* Check it is using the expected settings
* Use manual submission to submit a run
* Ensure this run appears in the webapp (the 2 are using the same settings)

Fixes #320

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 
 # Ignore settings files
 settings.py
+credentials.ini
 
 # Ignore distutils installation
 dist/
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,10 +32,11 @@ install:
   - python setup.py externals -s activemq,icat
   - python setup.py database
   - sudo python setup.py start
-
+  - cat build.log
 script:
   # ================ Assert Test Setup =============== #
   - pytest build/tests/test_db_generation.py
+  - exit(1)
 
   # ================ Static Analysis ================= #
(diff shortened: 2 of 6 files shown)


gc-code:
Change target to develop please.

PR #330, opened 2019-06-13
Take last RB number from summary

### LINKED ISSUE #329: New EoRM: Sometimes uses incorrect RB number
Issue raised by: developer

### What?
EoRM sometimes reads the wrong RB number on instruments where run numbers are truncated. I think the best approach to solving this will be just to use the last RB number like the current EoRM.

### Where?
New EoRM

### How?
When testing on the development environment I observed that some RB numbers are being reported incorrectly.

### Reproducible?
Yes

RB numbers are reported correctly.

gc-code:
### Summary of work
Now just takes last RB number from the summary file.

### How to test your work
Run on the development node and ensure all RB numbers are reported correctly. It should be clear whether it is running correctly after a few days.

Fixes #329

### DIFF
--- a/monitors/new_end_of_run_monitor.py
+++ b/monitors/new_end_of_run_monitor.py
@@ -60,19 +60,6 @@ def get_prefix_zeros(run_number_str):
     return zeros
 
 
-def is_integer(int_str):
-    """
-    Test to see if the provided string is an integer
-    :param int_str: Integer as a string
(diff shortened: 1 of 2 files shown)


gc-code:
Coverage has only decreased due to a reduction in the volume of code.

ElliotAOram:
I'll check this again on Monday morning to see how it did over the weekend

PR #339, opened 2019-07-11
Read RB number from the Nexus file

### LINKED ISSUE #338: New EoRM: Sometimes misses runs on ENGINX
Issue raised by: developer

### What?

The new end of run monitor can occasionally miss runs on ENGINX. The reason for this is that ENGINX's large summary file can be in the middle of writing when EoRM goes to read it. This means that it reads a random chunk from the summary file and submits it as the RB number. The submitted chunk is unlikely to be an integer so the run is just ignored as a calibration experiment.

A quick improvement for this would be to reduce the size of the ENGINX summary file so that it is faster to write.

A better solution may be to read the RB number from the Nexus file using the nexusformat library.

### Where?
New EoRM

### Reproducible?
No

Run the new EoRM for a cycle and see that it submits every run successfully on ENGINX.

gc-code:
### Summary of work

The new EoRM now loads RB numbers from the Nexus file itself, falling back on the summary file if necessary. The summary file will definitely still be used on instruments that produce HDF4 Nexus files. I considered it not worthwhile to load HDF4 files given that all instruments that use HDF4 are fine with the summary file. Existing libraries that load HDF4 are also quite difficult to setup because they rely on a C++ library underneath.

### How to test your work

There is now a wiki entry for testing EoRM locally: https://github.com/ISISScientificComputing/autoreduce/wiki/Testing-the-new-End-of-Run-Monitor

There are two additional dependencies to be installed in the following order:
 * numpy
 * nexusformat

I have also pulled the new code onto the development node.

Fixes #338

### DIFF
--- a/monitors/new_end_of_run_monitor.py
+++ b/monitors/new_end_of_run_monitor.py
@@ -7,6 +7,7 @@ import csv
 import logging
 import os
 import json
+from nexusformat.nexus import nxload
 from filelock import (FileLock, Timeout)
 
 from monitors.settings import (LAST_RUNS_CSV, CYCLE_FOLDER)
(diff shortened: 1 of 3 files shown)


gc-code:
This is ready to be reviewed.

ElliotAOram:
GitHub wouldn't let me merge this despite having a clean merge locally. I've merged in develop so it should now merge cleanly. Just to be sure we'll wait for travis

PR #352, opened 2019-08-01
Ansible Docker install

gc-code:
### Summary of work
I have added a Docker installation to the ansible-compute scripts. 

### How to test your work
Install Autoreduction on a fresh VM using the ansible-compute scripts, then check that Docker has been installed successfully.


ElliotAOram:
Added a couple of extras to start the docker daemon and ensure the user account (`isisautoreduce`) has permissions to use docker without the need for sudo

PR #361, opened 2019-08-09
325 MPL background

### LINKED ISSUE #325: Set MPL background in autoreduction code that uses Mantid
Issue raised by: [developer]

### What?
In order to ensure we do not get plotting errors from Mantid, we must set the following:
```
os.environ['MPLBACKEND'] = 'Agg'
```
This has been done as a temporary measure in the EnginX script, however a more permanent solution for all instruments is required. This would most likely be setting this variable before running any mantid code.

### Where?
* Autoreduction processor - `post_process_admin.py`
* Remove this from EnginX script once successfully validated

### How?
Noticed as an error on production

### Reproducible?
Yes - if you remove this line from EnginX script and run it you get a tkinter error. However, please DO NOT DO THIS as the production service uses these scripts

* Remove the `os.environ['MPLBACKEND'] = 'Agg'` line from the EnginX script
* Resubmit any EnginX run
* Ensure the run completes successfully without error

JackEAllen:
### Summary of work
Moving os.environ['MPLBACKEND'] = 'Agg' from Enginx script into Dockerfile, setting the variable on build.

### How to test your work
* Dockerfile builds and runs as expected
* Successfully run Mantid Docker

Fixes #325


QueueProcessors/AutoreductionProcessor/post_process_admin.py
```diff
@@ -27,6 +27,10 @@
 import traceback
 from contextlib import contextmanager
 
+
+# Ensure that Mantid does not attempt to plot to display
+os.environ['MPLBACKEND'] = 'Agg'
```
ElliotAOram: You should be able to remove this from here now as this is handled in the docker file

docker_reduction/mantid/tests/input/load_script.py
```diff
@@ -8,21 +8,115 @@
 Used to test that a file can be loaded into mantid from the input directory and
```
ElliotAOram: Can you put this file back to the original test script that was in here (I am guessing the tests will fail due to this)

docker_reduction/mantid/tests/input/reduce_vars.py
```diff
@@ -0,0 +1,30 @@
+standard_vars = {
```
ElliotAOram: This file doesn't need to committed either for the time being. We are not testing it and it will not be required in most cases

PR #365, opened 2019-08-15
Check mysql version

ElliotAOram:
### Additional comments
 

Fixes #xyz


ElliotAOram:
This was only for testing hence can be closed

PR #409, opened 2019-10-29
MySQL package to run scripts

### LINKED ISSUE #408: Use MySQL package to run SQL scripts
Issue raised by: [developer]

### What?
currently in the build steps for the project we use Popen (e.g. opening a new system process) to run the MySQL scripts. This is not a nice way to do thing especially when you start adding sudo into the mix on linux systems. It would make more sense to remain pythonic and instead create a connection to the database using the database client and then run the queries from there. 

### Where?
* `build/database/*`
* `utils/clients/database_client`

### How?
discovered while attempting to upgrade to python 3

* Ensure that the database is still built correctly on travis

ElliotAOram:
### Summary of work
* Use database client to run sql scripts rather than Popen

### How to test your work
* Ensure the database is set up correctly on travis

Fixes #408


ElliotAOram:
Note that build package needs to be run separately. I have added a comment to the code to explain this. I think perhaps we could just move all the tests to be run before pylint? But this only avoids the problem rather than fixing it. For now I think it is fine to ignore the issue

ElliotAOram:
Minor decrease in overall test coverage due to the addition of build module to test coverage

PR #457, opened 2020-02-13
Remove imp and email functionality

ElliotAOram:
### Summary of work
* Remove the use of deprecate imp in queue processor
* Remove email functionality from queue processor as it was never used and doesn't appear to work. Furthermore, I don't think we want this functionality


ElliotAOram:
test coverage increase due to removed code. System tests cover the change to module loading (removal of imp)

ElliotAOram:
Tested extensively on dev

PR #458, opened 2020-02-13
Skip calibration RBs

### LINKED ISSUE #451: Ensure that we are skipping RB 0

### What?
It would appear that RB 0 is no longer being skipped. I suggest this is because it existed in the EndOfRunMonitor code which was dropped in favour of the new ingestion solution. 
As such we should look to add this functionality back in 

### Where?
Queue Processor? 

### How?
Noticed with pre cycle runs 

### Reproducible?
Yes

* Manually submit runs that are of RB 0 or below and ensure they turn up in the database as skipped. 
* The database and the web app should report these as skipped and should display a sensible warning message

ElliotAOram:
### Summary of work
Ensure that RB0 or less AND String RBs are skipped and added to the skipped queue (not just never run)

### How to test your work
Manually submit an RB 0, ensure it shows up in the webapp as skipped and has an explanation 

Fixes #451

### DIFF
--- a/queue_processors/queue_processor/queue_processor.py
+++ b/queue_processors/queue_processor/queue_processor.py
@@ -208,12 +208,7 @@ class Listener:
         # Make sure the RB number is valid
         error_message = is_valid_rb(rb_number)
         if error_message:
-            logger.warning("Skipping non-integer RB number: %s", rb_number)
-            self._data_dict['message'] = 'Reduction Skipped: {}. Assuming run number to be ' \
-                                         'a calibration run.'.format(error_message)
-            skipped_queue = ACTIVEMQ_SETTINGS.reduction_skipped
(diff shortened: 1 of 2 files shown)


ElliotAOram:
Need to do my own manual testing + release notes before merge

queue_processors/queue_processor/queue_processor.py
```diff
@@ -216,6 +205,17 @@ def data_ready(self):
         self._data_dict['reduction_script'] = reduction_script
         self._data_dict['reduction_arguments'] = arguments
 
+        # Make sure the RB number is valid
```
JackEAllen: Nit: Would this be worth placing in a separate method?

ElliotAOram:
@JackEAllen this is ready for re-review. Thanks

JackEAllen:
Looks good! 👍

PR #461, opened 2020-02-18
Remove "Items per page" when filtering by RB

### LINKED ISSUE #441: Webapp remove items per page on RB filtering
Issue raised by: [developer]

### What?
When you filter by RB number on a web app instrument page (the page where you can see all the reduction jobs), the items per page filtering option should disappear or be disabled as it should not work for RB filtering.

At present the input box for runs per page disappears, but not the title 

### Where?
`WebApp/autoreduce_webapp/reduction_viewer/views.py` is the controller
`WebApp/autoreduce_webapp/templates/instrument_summary.html` is the view.

* Open the webapp
* Select an instrument
* Select Filter by RB Number from the drop down menu
* Click apply filter
* Ensure that the `items per page` option is greyed out or displayed

JEThompson97:
### Summary of work
Added a conditional on the "Items per page" header

_Before:_
image

_After:_
image

### How to test your work
- Open the webapp
- Select an instrument
- Select Filter by RB Number from the drop down menu
- Click apply filter
- Check that "Items per page" has displayed

### Additional comments
We might want to have it such that "Items per page" disappears/reappears when the selection is made (rather than only after filtre is applied)

Fixes #441

### DIFF
--- a/WebApp/autoreduce_webapp/templates/instrument_summary.html
+++ b/WebApp/autoreduce_webapp/templates/instrument_summary.html
@@ -65,9 +65,11 @@
                 <div class="col-md-3">
                     <h3>Sort by</h3>
                 </div>
+                {% if filtering == 'run' %}
                 <div class="col-md-3">
                     <h3>Items per page</h3>
                 </div>
+                {% endif %}
             </div>
             <div class="row">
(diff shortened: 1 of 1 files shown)


ElliotAOram:
This can actually be closed as the bug is fixed as a side effect #465

PR #467, opened 2020-02-18
Change WebApp form on filtre selection

### LINKED ISSUE #464: Allow WebApp to change on filtre selection

### What?
Currently, the WebApp form only changes after the user has selected _"Apply filters"_.
This means that if, for example, the user selects _"Run number"_ - then then have to _"Apply filters"_ before they can select an _"Items per page"_ value:

_Step 1) User selects "Filter by: Run number", then "Apply filters"_
image

_Step 2) User selects an "Items per page" value_
image

This 2-step process can be 1 step if the form can adjust dynamically

### Where?
`WebApp/autoreduce_webapp/reduction_viewer/views.py` is the controller
`WebApp/autoreduce_webapp/templates/instrument_summary.html` is the view.

### How?
We noticed it could be an enhancement during #461

### Reproducible?
Yes

- Open the webapp
- Select an instrument

**Check 1:**
- Select _Filter by: RB_
- Check whether _Items per page_ disappears

**Check 2:**
- Select _Filter by: Run number_
- Check whether _Items per page_ reappears

JEThompson97:
### Summary of work
JavaScript for instrument_summary created to resubmit the form when the "Filter by" selection changes.
This causes the python conditions (injected within the HTML) to rerun and thereby redetermine the page structure based on the current field values.

### How to test your work
- Open the webapp
- Select an instrument
- Try both _Filter by_ values
- Ensure the page reloads with the expected form elements

### Additional comments
- This solution was chosen over adjusting the HTML within the JavaScript based on the "Filter by" value because this would result in duplicated functionality for hiding/showing elements.
- A side effect of this solution choice is that the "Filter by" value is always applied instantly and automatically on selection.

Fixes #464 

### DIFF
--- a/WebApp/autoreduce_webapp/static/javascript/instrument_summary.js
+++ b/WebApp/autoreduce_webapp/static/javascript/instrument_summary.js
@@ -10,8 +10,7 @@
 
     var init = function init(){
         document.addEventListener("DOMContentLoaded", function(){
-            var thisElem = $(this);
-            setDefaultCursor(thisElem);
+            setDefaultCursor($(this));
         });
 
         var select = document.getElementById('filter_select');
--- a/WebApp/autoreduce_webapp/static/javascript/instrument_summary.js
+++ b/WebApp/autoreduce_webapp/static/javascript/instrument_summary.js
@@ -1,0 +1,11 @@
+(function(){
+    var init = function init(){
+        var select = document.getElementById('filter_select');
+        select.onchange = function(){
+            document.getElementById('filter_options').submit();
+        };
+
+    };
+
+    init();
+}())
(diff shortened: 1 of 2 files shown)


JEThompson97:
Notes on 45ae6a2

**1. Cursor only changing on movement:**
When I was testing on Chrome: following form submission, the cursor would only change back to default after being moved (a small issue that may have gone unnoticed but still..)
After a while of debugging, I found suggestions that this issue is due to Chrome Dev tools being open. When I tried on Edge - the issue was resolved! However, on Chrome, even with Dev tools closed, it still persists...
**2. Setting cursor on element, not just body:**
Rather than just change the cursor on the `document.body`, I found a [suggestion](https://stackoverflow.com/questions/36597412/mouse-cursor-set-using-jquery-css-not-changing-until-mouse-moved) to do so for `this` element as well, so I have implemented it this way.

Issue #380, opened 2019-10-07
Ensure 'Started by' field is correctly assigned


ElliotAOram:
Issue raised by: [developer]

### What?
The database currently has a `started by` field, however it is seemingly never correctly populated. We need to look into ensuring that user names (or failing this 'autoreduction' when run automatically) is used to populate these fields. 

### Where?
* QueueProcessors should be propagating this as they handle the saving of state for each step in the workflow of autoreduction. I would guess this should be an additional field passed to the database as part of a `ReductionRun` object.
* This should be added to the database object at the `/dataready` queue 
* Furthermore, this may require updates in the webapp as when a run is resubmitted by a user from here, the webapp will need to pass the user information (along with the all of the meta data about the run) to ActiveMQ in the message. 

### How?
Noticed after a run was resubmitted by no one knew who had triggered it

### How to test the issue is resolved
* Run a run using the manual submission script `scripts/manual_operations/manual_subission.py` and see that something is populated in the `started by` field of the database. -> In this case we expect it to be `Manual submission from CLI` (or similar)
* Login to the webapp and resubmit a run. Ensure that your user name is present in the started by field. -> In this case we expect `user name`.
* Add the changes to the development environment and wait for a run to be submitted automatically. -> In this case we expect `Autoreduction`.

JEThompson97:
We found the pre-existing "started_by" field to be an _Integer_ field.
@ElliotAOram believes this is intended to take a user number. However, it is _not_ a Foreign Key field.
This means that we are able to populate the field with a user_number for when a user submits a job, and use "control code" for manual and automatic submission.

We've chosen the following control codes:
- -1 = manual submission
- 0 = automatic

PR #477, opened 2020-02-25
Delete WebApp\autoreduce_webapp\tests\utils.py

JEThompson97:
### Summary of work
This PR deletes the `utils.py` file.
Upon making changes to the file as part of #380, @ElliotAOram noted that the file may no longer be in use.

On investigation, I found that none of the file's methods were being called anywhere outside of the file itself.
My presumption is that this means it is not being used anywhere and is therefore dead code.


ElliotAOram:
Thanks for finding/removing this

Issue #478, opened 2020-02-25
Display started by on Web application


ElliotAOram:
Issue raised by: [developer]

### What?
With the resolution of #380 we now are correctly asserting who has started runs (including those that are started automatically and manually by developers/users). As such it would be good to display this information on the webapp.
This should be in an extra field in the run summary page next to the other meta data about the run and should read as follows:

#### if started by user (started_by > 0)
* `Started by: user name (first name, last name)` e.g. `Started by: Joe Bloggs`
#### if started by automation (started by 0)
* `Started by: Autoreduction service`
#### if started by manual submission script (started by -1)
* `Started by: Development team`
#### if started by Null
* There are some cases (older runs / oddities) where the started_by field in the database entry might be null. These should not have the `started by` field in the run summary page at all. 

### Where?
view - `WebApp/autoreduce_webapp/templates/run_summary.html`
controller - `WebApp/autoreduce_webapp/reduction_viewer/views.py` `def run_summary(...)`

### How?
Extension task to #380 

### How to test the issue is resolved
* Set your `credentials.ini` to point to the development environment
* start the webapp
* Navigate to (<webapp_address>/runs/GEM/90107/1/)
	* Ensure that the user is stated as `Started by: Development team`
* Navigate to (<webapp_address>/runs/GEM/90107/)
	* Ensure that the user is stated as `Started by: Autoreduction service`
* Navigate to (<webapp_address>/runs/GEM/90107/3/)
    * Ensure that the user is stated as `Started by: Elliot Oram`
* Navigate to (<webapp_address>/runs/GEM/90050/)
    * Ensure that the there is no `Started by:` field present in the meta data table

PR #483, opened 2020-02-26
Implement SFTP Client

### LINKED ISSUE #470: IMPLEMENT : SFTP Client

### What?
In order to access the data from CEPH, we need an SFTP Client. We should continue with the patterns we have been using for the other autoreduction clients (`queue_client`, `database_client`, etc.).  These can be found in the `utils/clients` and they implement the `utils/clients/abstract_client.py`.
* The SFTPClient must be implemented along side the other clients using the same pattern
* The SFTPClient must have a `ClientSettings` (`utils/clients/settings/client_settings.py`) implementation and matching entry in the SettingsFactory (`utils/clients/settings/client_settings_factory.py`)
* The SFTPClient will need an entry in the `utils/test_credentials.ini` 
* The SFTPClient must have an accompanying test suite with full coverage (where possible)

image

### Where?
New code should be created in `utils/clients`

### How?
This is part of the implementation of the visualisation feature

* It should be possible to use the SFTPClient to get a file from any server
The following test script could act as a good test (assuming you are using the correct credentials.ini file & replacing the values of `SERVER_FILE_PATH ` and `LOCAL_FILE_PATH`):
```
from utils.clients.sftp_client import SFTPClient
sftp = SFTPClient()
sftp.connect()
SERVER_FILE_PATH = 'some/path/to/file'
LOCAL_FILE_PATH = 'some/path/to/file'
sftp.retrieve(SERVER_FILE_PATH, LOCAL_FILE_PATH)
```
You should then be able to find the server file in the given local file path

JEThompson97:
### Summary of work
SFTP Client implemented using pysftp, extending abstract_client.py.
Framework based on existing client classes (e.g. database_client, queue_client, icat_client).

### Additional comments
Currently this is a **work in progress**

Fixes #470 

### DIFF
--- a/utils/clients/settings/client_settings_factory.py
+++ b/utils/clients/settings/client_settings_factory.py
@@ -49,6 +49,8 @@ class ClientSettingsFactory:
             settings = self._create_icat(**kwargs)
         elif settings_type.lower() == 'queue':
             settings = self._create_queue(**kwargs)
+        elif settings_type.lower() == 'sftp':
+            settings = self._create_sftp(**kwargs)
         return settings
 
     def _create_database(self, **kwargs):
@@ -75,6 +77,14 @@ class ClientSettingsFactory:
         icat_kwargs = ['authentication_type']
         self._test_kwargs(icat_kwargs, kwargs)
         return ICATSettings(**kwargs)
+
+    def _create_sftp(self, **kwargs):
+        """
+        :return: SFTP compatible settings object
+        """
+        sftp_kwargs = []  # TODO: figure out what the sftp_kwargs should be
+        self._test_kwargs(sftp_kwargs, kwargs)
+        return SFTPSettings(**kwargs)
 
     def _test_kwargs(self, expected, actual):
         """
@@ -149,3 +159,13 @@ class ActiveMQSettings(ClientSettings):
         self.reduction_skipped = reduction_skipped
         self.all_subscriptions = [data_ready, reduction_started,
                                   reduction_complete, reduction_error, reduction_skipped]
+
+
+class SFTPSettings(ClientSettings):
+    """
+    SFTP settings object
+    """
+
+    # TODO: figure out what to populate this class with
+    def __init__(self, **kwargs):
+        super(SFTPSettings, self).__init__(**kwargs)
--- a/utils/clients/settings/client_settings_factory.py
+++ b/utils/clients/settings/client_settings_factory.py
@@ -34,9 +34,9 @@ class ClientSettingsFactory:
                                                 reduction_complete, reduction_error
         :return: A ClientSettings object
         """
-        if settings_type.lower() not in ['database', 'icat', 'queue']:
-            raise ValueError("Factories creation settings type must be one of: 'database', "
-                             "'icat', 'queue'")
+        if settings_type.lower() not in ['database', 'icat', 'queue', 'sftp']:
+            raise ValueError("Factories creation settings type must be one of:"
+                             "'database','icat', 'queue', 'sftp'")
         kwargs['username'] = username
         kwargs['password'] = password
         kwargs['host'] = host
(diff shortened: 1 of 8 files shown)


utils/clients/settings/client_settings_factory.py
```diff
@@ -34,9 +34,9 @@ def create(self, settings_type, username, password, host, port, **kwargs):
                                                 reduction_complete, reduction_error
         :return: A ClientSettings object
         """
-        if settings_type.lower() not in ['database', 'icat', 'queue']:
-            raise ValueError("Factories creation settings type must be one of: 'database', "
-                             "'icat', 'queue'")
```
ElliotAOram: in raise value can we change this to be 
```
valid_types = ['database', 'icat', 'queue', 'sftp']
if settings_type.lower() not in valid_types:
	raise ValueError(f"Factories creation settings type must be one of: {','.join(valid_types)}")
```

JEThompson97: > in raise value can we change this to be
> 
> ```
> valid_types = ['database', 'icat', 'queue', 'sftp']
> if settings_type.lower() not in valid_types:
> 	raise ValueError(f"Factories creation settings type must be one of: {','.join(valid_types)}")
> ```

I've added this change In ef9252c, but have put `valid_types` into Class scope so that it can be accessed by the test method (rather than the test method making it's own `valid_types` list)

JEThompson97:
**Please note:**
I've referenced the wrong issue for commits 0d2f030 and e2da684.
(#478 instead of #470 -which it should be)

JEThompson97:
Regarding the Travis build errors:
- I _believe_ the `test_valid_connection` issue is caused because the code can't access real credentials (naturally, because we don't upload them to github).
- The rest of the issues are pylint complaining that I'm accessing variables within `sftp_client` directly, but I need to do this in order to mock them.

Do correct me if I'm wrong about either of the above.

ElliotAOram:
@JEThompson97 You are correct regarding the `test_valid_connection` issue and it's cause. As such you will have to mock this. This may involve mocking the connection AND the command you run in `_test_connection` to test the connection. In short this test case therefore me checking that `_test_connection` returns True when the connection doesn't throw anything.

You're right on the second issue as well. You will have to add `# pylint:disable=protected-access` to avoid this - there is no way around this in testing

utils/clients/sftp_client.py
```diff
@@ -0,0 +1,102 @@
+# ############################################################################### #
+# Autoreduction Repository : https://github.com/ISISScientificComputing/autoreduce
+#
+# Copyright &copy; 2019 ISIS Rutherford Appleton Laboratory UKRI
```
ElliotAOram: Update to 2020

utils/clients/sftp_client.py
```diff
@@ -0,0 +1,102 @@
...
+        Disconnect from the SFTP server
+        """
+        if self._connection is not None:
+            self._connection.close()
+        self._connection = None
+
+    def _test_connection(self):
+        """
+        Test whether there is a connection to the SFTP server
+        :return: True if there is a connection
```
ElliotAOram: Can you add the `:raises:` tag in the docstring here

utils/clients/sftp_client.py
```diff
@@ -0,0 +1,102 @@
...
+
+        try:
+            self._connection.pwd
+        except AttributeError:
+            raise ConnectionException("SFTP")
+        return True
+
+    def retrieve(self, server_path, local_path=None, override=True):
+        """
+        Retrieves file from the given server_path and downloads it to the given local_path
```
ElliotAOram: again `:raises:` tag. This is not used in the rest of the code base currently but is something we are looking to add as we go

utils/clients/tests/test_sftp_client.py
```diff
@@ -0,0 +1,118 @@
+# ############################################################################### #
+# Autoreduction Repository : https://github.com/ISISScientificComputing/autoreduce
+#
+# Copyright &copy; 2019 ISIS Rutherford Appleton Laboratory UKRI
```
ElliotAOram: 2020

ElliotAOram:
Just need to move pysftp package to `setup.py` from `requirements.txt`

Issue #474, opened 2020-02-25
Design : Plotting Meta language


ElliotAOram:
Issue raised by: [project requirement]

### What?
In order to allow for a generic way to plot that can be implemented easily and is extensible. The current suggestion is to have a plotting meta language (perhaps yaml/xml/json based), that the plotting factory can easily interpret.

* In the first instance we should check to see if something like this already exists.
  * If it does and can be adapted for our needs we should use it
* Failing this we will have to write our own meta language for graphing and it must define:
  * plot type
  * axis units
  * ...

### Where?
This will produce documentation along with an example file of the meta language

### How to test the issue is resolved
* Ensure that the meta language is extensible
* Ensure that the meta language covers all/most of the criteria for a plot
* Ensure that relevant documentation is provided
* Ensure that the design UML is updated to include the meta language and an interpreter to go between the plotting factory and the meta language

PR #490, opened 2020-03-09
Create Interpreter for plotting meta language

### LINKED ISSUE #475: Implement: Plotting meta language and interpreter

### What?
Once we are happy with the output for #474 we should look into implementing the interpreter for the plotting meta language. This will be responsible for the decoding the meta language into something that plotly can understand. This may well form a large part of the plotting factory but we can assess at a later stage if this is infact the same as the plotting factory.

The interpreter must be able to: 
* Read a meta data file
* Return a valid plotly construct for the plot
* Inform if any error or unrecognised values exist in the provided meta data file

### Where?
New code in the plotting module

* Ensure that you can provide a variety of plotting meta language files to the interpreter and get the desired result. This should include but not be limited to the example file generated from #474 
* Ensure that if you provide invalid files you receive sensible error messages
* Ensure that this is unit tested and run in travis
* Ensure that the documentation created in #474 is still up to date
* Ensure that API style documentation exists for the plotting meta language

JEThompson97:
### Summary of work
- This PR creates an interpreter to read plot type files as a _plotly Layout dictionary_.
- Due to design decisions made by the team during the development of this interpreter - its functionality is very minimal.
- It simply loads a plot type file from a given location (.yaml expected by .txt works too) and interprets the data as a python dict.
- If the highest level key of the dict is 'figure', as it is currently in the `example.yaml` file (see #474), then this level is removed and only its value is returned. Otherwise, the whole dict is returned.

image

### How to test your work
- Run the accompanying `test_interpreter.py` file and ensure all tests pass
- Run the interpreter with a variety of plotting meta language files to the interpreter and get the desired result. This should include but not be limited to the example file.

### Additional comments
- Since the python `yaml` library loads yaml data as a python dictionary natively, no post-load interpretation is required.

Fixes #475 


JEThompson97:
Please note: 6d90811 references the wrong commit (#470 when it should be: #475)

JEThompson97:
**The reason for the Travis failure:**
The Interpreter class doesn't have any internal state, so it's method's do not reference itself. Pylint therefore wants to refactor its methods into class-less functions. Doing so would entirely remove the Interprete class in favour of just an interpret method.
Not sure what we want to do here?

requirements.txt
```diff
@@ -1,3 +1,4 @@
+PyYAML
```
ElliotAOram: Does this not need to be in setup.py?

ElliotAOram:
I see why pylint wants this approach, however, as I can see that the code might grow and I want to keep this code encapsulated together I would like to keep it as a class. 

As such I would use the `@staticmethod` decorator on the methods that do not reference `self`.

requirements.txt
JEThompson97: Neither PyCharm nor Travis has complained..
I've just checked - it's the same case for PySFTP - it's in the requirement.txt file but not the setup.py
I can add them both if you'd like? I'm not too familiar with this aspect of the code

ElliotAOram: Yes could you please add them both to `setup.py`
In theory hard dependencies on the project (e.g. Autoreduction will not work without these) are in `setup.py`. soft dependencies live in `requirements.txt` (e.g. packages that help with testing the project like Mock)

JEThompson97: Sure, I'll add both to `setup.py`.
Would you like me to remove them from `requirements.txt` too? Or leave them there?

ElliotAOram: Yes can you remove this from `requirements.txt`

JEThompson97: I've done so for this and #483, but the Travis build seems to fail as a result

PR #491, opened 2020-03-11
Skip when beam off

### LINKED ISSUE #452: Skip runs when the beam is off

### What?
It should be possible to find when the beam has low power (<10uAmp) and hence we should skip reduction. 

### Where?
Queue Processor? 

### How?
Noticed with pre cycle runs 

### Reproducible?
Yes

* Manually submit runs have low uAmp (beam off) and ensure they turn up in the database as skipped. (MARI26905-26909)
* The database and the web app should report these as skipped and should display a sensible warning message

ElliotAOram:
### Summary of work
* Added validation module that performs various validation checks on the data
* Added `testing_data` module that houses some `.nxs` files used in testing
* Added to the data archive creator functionality to ensure that it can handle copying "real" files (not just mocked empty files) into a fake data archive. 
* Ensured the system tests still pass by using a valid data file (note these cannot be mocked as the queue processor code is started as an  external service from within the system test) 

### How to test your work
* Code review
* Checkout this branch on the development compute node and restart the queue processors
* Resubmit some runs with 0 beam current
  * Ensure these are skipped
* Resubmit some runs with 0 or less as an RB number ( or a string RB)
  * Ensure these are skipped
* Resubmit normal runs we expect to be reduced
  * Ensure these actually get reduced 

Fixes #452

*Release notes have been updated*


ElliotAOram:
In order to fix system tests it should be possible to provide a custom file location path. at the moment we make assumptions on this being ISIS only

queue_processors/queue_processor/validate_job.py
```diff
@@ -0,0 +1,60 @@
...
+        return "Calibration file detected (RB Number less than or equal to 0)"
+    except ValueError:
+        return "Calibration file detected (RB Number is a string)"
+
+
+def check_beam_current(run_file_location):
+    """
+    Ensure that that there is value for beam current in the file else we can assume the beam is off
+    :param run_file_location: The location of the run file to inspect
+    :return: An error message if one is generated or None if run has beam current
```
JEThompson97: Is it worth more accurately saying
"if run has beam current **greater than 0.1 (10uAmp)** "
or "if run has **significant** current"
?

testing_data/__init__.py
```diff
@@ -0,0 +1,13 @@
+# ############################################################################### #
+# Autoreduction Repository : https://github.com/ISISScientificComputing/autoreduce
+#
+# Copyright &copy; 2020 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+# ############################################################################### #
+"""
+Paths to be data used for testing the project that is not easy to mock / fake e.g. meta data complete nexus file
```
JEThompson97: "Paths to ~~be~~ data"

queue_processors/queue_processor/tests/test_validate_job.py
```diff
@@ -0,0 +1,90 @@
...
+import unittest
+from mock import patch
+
+from queue_processors.queue_processor import validate_job as validate
+from testing_data import VALID_NEXUS, ZERO_BEAM_NEXUS
+
+
+class TestValidateJob(unittest.TestCase):
+    """
+    Test cases for functions that valid an autoreduction job should run
```
JEThompson97: "Test cases for functions that **a** valid ~~an~~ autoreduction job should run"

queue_processors/queue_processor/tests/test_validate_job.py
```diff
@@ -0,0 +1,90 @@
...
+    def test_validate_job_valid(self):
+        """ Assert none is returned if all validation checks pass"""
+        self.assertIsNone(validate.validate_job(rb_number=100,
+                                                file_location=self.valid_nxs_location))
+
+    @patch('queue_processors.queue_processor.validate_job.is_valid_rb')
+    @patch('queue_processors.queue_processor.validate_job.check_beam_current')
+    def test_validate_job_one_invalid(self, mock_check_beam, mock_is_valid_rb):
+        """
+        Test that just a single expected error message if one of the validation functions fail
```
JEThompson97: "Test that just a single expected error message **is returned** if one of the validation functions fail"

ElliotAOram:
@JEThompson97 I think this ready to be reviewed again 👍

ElliotAOram:
Thanks @JEThompson97 I'll have a look at point 1.

The second behaviour is fine. RBNumbers are sometimes passed as strings for calibrations and this is what is expected in autoreduction.

ElliotAOram:
There are various problems with this validation:
* How do we know what acceptable beam is?
  * As Franz pointed out some experiments do not need beam to run e.g. measure cosmic radiation.
  * If the final beam value is 0 does not mean we should necessarily discount the whole experiment and visa versa with the first value. 
  * We can not rely on averages due to beam fluctuation
I am leaning towards the side of not performing this check. The number of times we see this is fairly minimal and I think the chance that we see valid data as invalid and therefore do no reduce it is too high.

PR #492, opened 2020-03-16
SOAP to replace WebScraper

### LINKED ISSUE #488: Replace web scraper by soap call to User office scheduler
Issue raised by: [developer]

### What?
The webscraper for cycle dates is not the cleanest solution to ingest cycle and maintenance dates. These can be obtained via the user office scheduler API. 
This will require there to be SOAP calls to this service. 

### Where?
Replacement of the web scraper

### How?
Discussion with business apps led to us discovering this was possible

TBC

JEThompson97:
### Summary of work
- This PR creates a `BusAppsIngestionClient` to interface the scheduler and uows APIs in order to retrieve cycle data.
- It also creates a `SchedulerDataProcessor` to process (combine, sort, clean, refactor) the raw data retrieved through the Client into a more helpful format.
___
* Before I add finishing touches (see Additional comments below), could I please get a review of the work done thus far? Specifically:
  * The way I've implemented the functionality (I've left `TODO` comments to discuss design decisions)
  * The names of variables, methods, classes

### How to test your work
- Review code
- Run `\autoreduction\utils\clients\busapps_ingestion_client.py`, examine output

### Additional comments
Work still to do (after a review of work thus far):
- Create tests
- Make cycle-data classes serialisable
- Write docstrings

Fixes #488 


scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,98 @@
+import re
```
ElliotAOram: Needs copyright statement

ElliotAOram:
@JEThompson97 will finish looking through this tomorrow

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class SchedulerDataProcessor:
+    """
+    Class to process data received from the Scheduler API
+    """
+    def __init__(self):
+        self._earliest_possible_date = datetime(2000, 1, 1, tzinfo=None)
+        # TODO: confirm what is to be accepted (vs what is invalid)
+        #   Current regex =  4 digits | '/' | 1 or more digit/letter(s) | end
```
ElliotAOram: I think this regex is fine 👍

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,159 @@
...
+        pre_processed = self._pre_process(raw_cycle_data, raw_maintenance_data)
+        return self._process(pre_processed)
+
+    @staticmethod
+    def print_start_dates(data):
+        """
+        Print each 'start' attribute within a list, prefixed by an index.
+        :param data: The list of items to print the start dates of
+        """
+        for idx, item in enumerate(data):
```
ElliotAOram: nit: just use `index` here to be fully verbose

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,159 @@
...
+            else:
+                clean_list.append(item)
+        print(f"New length: {len(clean_list)}")
+        return clean_list
+
+    # TODO: question - I've used the method of combining the lists as this was suggested,
+    #  but I'm not exactly sure why we do this instead of keeping them separate?
+    #  Is it a speed optimisation perhaps? I wonder if it makes us more prone to bugs/confusing code..
+    @staticmethod
+    def _combine_lists(first, second):
```
ElliotAOram: The reason to combine the lists is to make it easier/more efficient to process the maintenance days vs. the cycles. More specifically we start with:
```
maintenance_days = [suds.maintenance_day, ...]
cycles = [suds.cycles, ...]
```
from this point, we need to do the following steps to reach the desired output:
* Create Cycle / MaintenanceDay objects corresponding to each of the suds object
* Populate the Cycle objects with the maintenance days object that exist within their date range

By combining the lists first, we can sort both the cycles and maintenance_days by date. This means the list now looks like the following:
```
all_items = [cycle, maintenance_day, maintenance_day, cycle, maintenance_day, cycle, ...]
```
When the list is in this form it is easier and more efficient to convert the suds objects to our own custom data objects. The reason for this as we can loop through the full list and structure all the data in once because of the following assumption:
* In the sorted list, a cycle object is followed by the maintenance days that exist within its date range. 

Therefore we can create a cycle object then create maintenance day objects for all the following list items **until** we hit a new cycle object where we then can create a new cycle object and adding following maintenance days to it. 

tldr; it means that the complexity of this data structuring is `O(n)` as it's a single loop. If you store these lists separately, then its `O(n^2)` as you have to search through the maintenance day list every time you add the maintenance days to the cycles. It is marginally more efficient if both lists are sorted, but still not as fast as the first implementation.

ElliotAOram: Having said all of this, we are talking about the efficiency over a rather small list (20+ items). I wonder if the readability would be better for separate lists where the process would be:
* sort both lists and keep them separate
* for each cycle:
  * create a new cycle object
  * look through the sorted maintenance day list
  * check to see if the date of the maintenance day falls within the cycle (should be the first n items as the list is sorted)
  * if maintenance day date is within cycle:
    * create maintenance day obj
    * add to cycle obj
    * remove maintenance suds obj from sorted list
  * else:
   * move to next cycle item

ElliotAOram: okay before you do anything else can you implement the two sorted lists implementation and then we can compare the difference and decide which we want to go with? I agree that the current implementation is probably less obvious than using 2 lists.

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,159 @@
...
+    def _sort_by_date(self, data):
+        """
+        Sorts a list by a date field (specified by the _sort_by_field variable)
+        :param data: The list to be sorted
+        :return: The list sorted by the specified field
+        """
+        print("sorting..")
+        return sorted(data, key=lambda date: date[self._sort_by_field])
+
+    # TODO: question - can a maintenance day ever occur outside of a cycle period?
```
ElliotAOram: They shouldn't and I think we should trust that this is validated at the business apps end. Although could you add a comment to say that we are making this assumption?

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,159 @@
...
+        :param data: A pre-processed list containing all valid cycle and maintenance day data items.
+        :return: A list of Cycle objects containing a list of 0 or more MaintenanceDay objects.
+        :raises RuntimeError:
+            If a maintenance day is encountered in the data before a cycle.
+            This will only occur if the data has not been properly pre-processed.
+        """
+        cycle_list = []
+        idx = 0
+        while idx < len(data):
+            # TODO: I've avoided using try/except for expected behaviour, unlike the draft implementation
```
ElliotAOram: This is fine. Either implementation is acceptable in python

ElliotAOram:
Looks pretty good on the whole. Client particularly looks to be very good 👍 Have left a few discussion points. If you want to talk to me about them, please get in touch

scripts/scheduler_ingest.py
JEThompson97: Very well summarised, thanks for this.
I see what you mean about keeping the complexity as o(n). What you've gone on to pseudocode there with the bullet points was actually what I was going to suggest! That method seems more readable, easier to add more helpful exceptions, and (if I add some well-placed conditions) should maintain o(n).

As you say, I'll implement the alternative and we'll go from there.

JEThompson97: *Interesting update:*
I've implemented your pseudo code (about to commit+push) and within it **included a check** that the latest **maintenance day ends before** the current **cycle's end date**.

On running - it's picked up an issue with the data I'm working with which the original method didn't pick up on
(see _output from the original_ method below with issue highlighted)

image

ElliotAOram: Oh yes that's quite strange. Those maintenance days don't exist inside any cycle... is this for cycle 20_2. I wonder if it's an issue with the scheduling for it being off with all the current covid-19 stuff?

ElliotAOram: Checking the beam status page, neither of the out of cycle maintenance day dates are on their table either. https://www.isis.stfc.ac.uk/Pages/Beam-Status.aspx
We should validate this with BusApps to see where these dates have come from. Although I suggest that this is low priority as the dates are for 20_1 which is not going to be going ahead.

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,213 @@
...
+    def _clean_data(self, data):
+        """ Removes invalid items from the data.
+        :param data: The list of items to be validated
+        :return: The list of items that were passed in, but with all invalid items removed """
+        print("\nBEGIN Data clean")
+        print(f"Original length: {len(data)}")
+        clean_list = []
+        for item in data:
+            if 'name' in item and not re.search(self._cycle_name_regex, item['name']):
+                print(f"\nItem removed due to strange cycle name "
```
ElliotAOram: Item removed ~due to strange~ as cycle name did not match the expected regex

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,213 @@
...
+                      f"({item['name']}).\nFull item: {item}")
+            elif item['start'].replace(tzinfo=None) < self._earliest_possible_date:
+                print(f"\nItem removed due to impossible date "
```
ElliotAOram: Item removed as ~due to impossible~ date lies outside range of interest

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,213 @@
...
+    def _process_from_separate(cycle_data, maintenance_data):
+        """ Converts pre-processed cycle data items and maintenance day data items
+        into a single list of Cycle objects. Each cycle object contains a list
+        of MaintenanceDay objects - all of which fall within the given cycle period.
+        Maintenance Days which fall outside of cycle periods are highlighted and discarded.
+        :param cycle_data: A pre-processed list containing all valid cycle data items.
+        :param maintenance_data: A pre-processed list of all valid maintenance day data items.
+        :return: A list of Cycle objects containing a list of 0 or more MaintenanceDay objects. """
+        cycle_list = []
+        unused_maintenance_data = maintenance_data.copy()
```
ElliotAOram: `unused_maintenance_data` might be better as just `maintenance_data`. The `unused` suggests that they are not going to be used at all (atleast that's how I read it)

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,213 @@
...
+                                             f"end: {unused_maintenance_data[1].end}"
+                    else:
+                        cycle_after_string = f"NONE (No later cycle dates)"
+                    print(f"WARNING - Encountered maintenance day outside of cycle dates "
+                          f"(assuming cycle_data and maintenance_data "
+                          f"were ordered as earliest-to-latest).\n"
+                          f"Maintenance Day= start: {m_day['start']}, end: {m_day['end']}\n"
+                          f"Closest Cycle before= start: {cycle_obj.start}, end: {cycle_obj.end}\n"
+                          f"Closest Cycle after= {cycle_after_string}\n"
+                          f"This Maintenance Day has therefore been discarded.\n")
```
ElliotAOram: This warning is very useful, but makes the code difficult to read as it breaks it up alot if you are trying to follow the logic. As such could you refactor it into a function for readability?

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,213 @@
...
+                    # if the next m_day is EARLIER than the current cycle END
+                    cycle_obj.add_maintenance_day(m_day['start'], m_day['end'])
+                    unused_maintenance_data.pop(0)
+                else:
+                    # if the next m_day is LATER than the current cycle END
+                    break
+            cycle_list.append(cycle_obj)
+        return cycle_list
+
+    def _process_as_single_list(self, data):
```
ElliotAOram: This can now be remove as other implementation is better

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,213 @@
...
+                print(f"\nItem removed due to impossible date "
+                      f"({item['start']}).\nFull item: {item}")
+            else:
+                clean_list.append(item)
+        print(f"New length: {len(clean_list)}")
+        print("END Data clean\n")
+        return clean_list
+
+    @staticmethod
+    def _combine_lists(first, second):
```
ElliotAOram: This can now be removed

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,213 @@
...
+
+    @staticmethod
+    def print_start_dates(data):
+        """ Print each 'start' attribute within a list, prefixed by an index.
+        :param data: The list of items to print the start dates of """
+        for idx, item in enumerate(data):
+            print(f"{idx}: {item['start']}")
+
+    def _pre_process_as_single_list(self, raw_cycle_data, raw_maintenance_data):
+        """ Makes adjustments to the data without converting it to a different data type.
```
ElliotAOram: This can now be removed

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,213 @@
...
+    def __init__(self):
+        self._earliest_possible_date = datetime(2000, 1, 1, tzinfo=None)
+        # Current regex =  4 digits | '/' | 1 or more digit/letter(s) | end
+        self._cycle_name_regex = r"\d{4}/\w*$"
+        self._maintenance_specific_key = 'facility'
+        self._datetime_fields = ["start", "end"]
+        self._sort_by_field = "start"
+
+    def convert_raw_to_structured_as_single_list(self, raw_cycle_data, raw_maintenance_data):
+        """ Converts raw data received from the Scheduler API into a list of Cycle objects,
```
ElliotAOram: This can now be removed

utils/clients/busapps_ingestion_client.py
```diff
@@ -0,0 +1,125 @@
...
+            "invalid_scheduler_client": TypeError("The Scheduler Client does not exist"
+                                                  "or has not been initialised properly"),
+            "invalid_session_id": ConnectionException("BusApps Ingestion")
+        }
+
+    def create_uows_client(self):
+        """ Creates a User Office Web Service Client is one does not exist """
+        if self._uows_client is None:
+            self._uows_client = Client(self.credentials.uows_url)
+            # print(type(self._uows_client))
```
ElliotAOram: remove commented print statement

utils/clients/busapps_ingestion_client.py
```diff
@@ -0,0 +1,125 @@
...
+                                             "or has not been initialised properly"),
+            "invalid_scheduler_client": TypeError("The Scheduler Client does not exist"
+                                                  "or has not been initialised properly"),
+            "invalid_session_id": ConnectionException("BusApps Ingestion")
+        }
+
+    def create_uows_client(self):
+        """ Creates a User Office Web Service Client is one does not exist """
+        if self._uows_client is None:
+            self._uows_client = Client(self.credentials.uows_url)
```
ElliotAOram: to be verbose and to avoid confusion, I wonder if we would be better changing the import to:
`import suds` and then referencing these clients as `suds.Client`. I only say this Autoreduction has it's own understanding of a `Client` and seeing it written here it's not immediately obvious what the client is

utils/clients/busapps_ingestion_client.py
```diff
@@ -0,0 +1,125 @@
...
+            # print(type(self._uows_client))
+
+    def create_scheduler_client(self):
+        """ Creates a Scheduler Client is one does not exist """
+        if self._scheduler_client is None:
+            self._scheduler_client = Client(self.credentials.scheduler_url)
```
ElliotAOram: Same comment as above re `suds.Client`

utils/clients/busapps_ingestion_client.py
```diff
@@ -0,0 +1,125 @@
...
+        return self._scheduler_client.service.getCycles(sessionId=self._session_id)
+
+    def ingest_maintenance_days(self):
+        """ Ingests maintenance day dates from the Scheduler client.
+        :return: Maintenance day data as a list of 'sudsobject's """
+        return self._scheduler_client.service.getOfflinePeriods(sessionId=self._session_id,
+                                                                reason='Maintenance')
+
+
+# TODO: Please note - all of the code below is for testing purposes only.    # pylint:disable=fixme
```
ElliotAOram: This code can now be removed

utils/clients/busapps_ingestion_client.py
```diff
@@ -0,0 +1,125 @@
...
+"""
+Client class for ingesting BusApps data via SOAP
+"""
+from suds import Client, WebFault
+from utils.clients.abstract_client import AbstractClient
+from utils.clients.connection_exception import ConnectionException
+from utils.test_settings import BUSAPPS_SETTINGS
+
+
+class BusAppsIngestionClient(AbstractClient):
```
ElliotAOram: What do you think about changing the name to `CycleIngestionClient` and then specify in the class doc string and file doc string that the ingest is via BusinessApps? My only thought is others looking at the code base will have no concept of busapps (e.g. those from other facilities)

utils/clients/tests/test_busapps_ingestion_client.py
```diff
@@ -0,0 +1,181 @@
...
+    def create_client(self, to_mock=None, credentials=None):
+        """ Returns a client with 0 or more mocked instance variables
+        :param to_mock: A list of client instance variables to mock
+        :param credentials: The credentials to initialise the client with
+                            (test credentials used if none supplied)
+        :return: A client with 0 or more mocked instance variables """
+        if not credentials:
+            credentials = self.test_credentials
+        client = BusAppsIngestionClient(credentials)
+        if to_mock and "_uows_client" in to_mock:               # pylint: disable=protected-access
```
ElliotAOram: Should just be able to put the disable for pylint on the top line of these functions and it will do it for all the below

utils/clients/tests/test_busapps_ingestion_client.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def test_invalid_init(self):
+        """ Test initialisation raises TypeError when given invalid credentials """
+        with self.assertRaises(TypeError):
+            BusAppsIngestionClient("invalid")
+
+    def test_default_init(self):
+        """ Test initialisation values are set """
+        client = BusAppsIngestionClient()
+        self.assertIsNone(client._uows_client)          # pylint: disable=protected-access
```
ElliotAOram: same as above re. pylint issues

utils/clients/tests/test_busapps_ingestion_client.py
```diff
@@ -0,0 +1,181 @@
...
+        self.assertIsNone(client._scheduler_client)     # pylint: disable=protected-access
+        self.assertIsNone(client._session_id)           # pylint: disable=protected-access
+
+    # TODO: Note - EO mentioned I shouldn't need to mock the Client as we assumed   # pylint: disable=fixme
```
ElliotAOram: Looks good 👍

utils/clients/tests/test_busapps_ingestion_client.py
JEThompson97: This is the best tip ever.
I'm so tired of copying&pasting `pylint:disable` comments, haha

scripts/scheduler_ingest.py
JEThompson97: For readability, I feel inclined to avoid using the same name as the method argument `maintenance_data` (the method which this variable is a local copy of).

Would `maintenance_data_copy` work perhaps?

JEThompson97:
Note on `_unexpected_maintenance_day_warning` tests:

- I spent a while trying to add fine-grained checks within the tests for `_unexpected_maintenance_day_warning`.
- I was trying to check the method was being called with the expected objects as arguments, but it began to complicate the tests a fair bit.
- As such, I've simplified the tests for these to just check that the method is called once.

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,252 @@
...
+    Class to represent a generic time period
+    """
+    def __init__(self, start, end):
+        self.start = start
+        self.end = end
+
+    def __eq__(self, other):
+        return self.as_dict() == other.as_dict()
+
+    # TODO: Note to EO [#2] - I ended up not actually using     # pylint:disable=fixme
```
ElliotAOram, line 25: I think if we are not using them lets remove them for now. They are not hard to re-implement and I don't want to bloat the code base with functionality that we don't yet require and have not immediate plan to use

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,252 @@
...
+        dict_.update({"name": self.name,
+                      "maintenance_days": self.maintenance_days})
+        return dict_
+
+    def add_maintenance_day(self, start, end):
+        """ Adds a maintenance day object to this cycle
+         :param start: The start date of the maintenance day being added
+         :param end: The end date of the maintenance day being added """
+        self.maintenance_days.append(MaintenanceDay(start, end))
+        # if start < self.start:
```
ElliotAOram, line 71: Can remove this now?

scripts/scheduler_ingest.py
```diff
@@ -0,0 +1,252 @@
...
+
+        print(f"WARNING - Encountered maintenance day outside of cycle dates "
+              f"(assuming cycle_data and maintenance_data "
+              f"were ordered as earliest-to-latest).\n"
+              f"Maintenance Day:\n\tStart = {md_data['start']}\n\tEnd = {md_data['end']}\n"
+              f"Closest Cycle before:{cycle_before_string}"
+              f"Closest Cycle after:{cycle_after_string}"
+              f"This Maintenance Day has therefore been discarded.\n")
+
+# test_cycle_values = {
```
ElliotAOram, line 222: Can be removed now?

ElliotAOram:
Looks really good now! 👍

PR #502, opened 2020-03-31
Add prepare_data

### LINKED ISSUE #473: Implement : Prepare data

### What?
* Implement the functionality for extracting and preparing data for plotly.
* This must match the data formats set out in #472.
* This code must match the below design and have an accompanying test file
  * In most likely is not required to be a class - a script should suffice

image

### Where?
New file inside the plotting module 

* Ensure that a file supplied in the correct format yields and valid numpy array return (#472 will define correct format)
* Ensure that a incorrect file type is rejected and raises an exception
* Ensure that incorrect input data (even if the file type is correct) is rejected and raises an exception
* Ensure all of the above error cases are accounted for in the test file

JEThompson97:
### Summary of work
- This PR adds a `prepare_data.py` component, along with tests.
- `prepare_data.py` primary function is to load a CSV file from a given path, converts it into a `pandas.DataFrame` and returns this.
- Formats are outlined below:

#### Expected input [example]
```
# X , Y , E 
1
1.0, 2.0, 3.0 
4.0, 5.0, 6.0
1.0, 2.0, 3.0
```
_Note:_ Line 1 above specified by instance variable `PrepareData.expected_first_row`

#### Output [example]
```
|Spectrum | X     | Y   | E   |
| :---    | :---: |:---:| ---:|
| 1       |       |     |     |
| 1       | 1.0   | 2.0 | 3.0 |
| 1       | 4.0   | 5.0 | 6.0 |
| 2       | 1.0   | 2.0 | 3.0 |
```

### How to test your work
* Review code
* Run tests: `\plotting\plot_meta_language\tests\test_prepare_data.py`

Fixes #473

### DIFF
--- a/plotting/prepare_data.py
+++ b/plotting/prepare_data.py
@@ -15,69 +15,54 @@ import csv
 class PrepareData:
 
     def __init__(self):
-        self.expected_file_first_line = "# X , Y , E"
+        self.expected_file_first_row = "# X , Y , E"
         self.columns = ["Spectrum", "X", "Y", "E"]
 
     # TODO: consider pd.read_csv VS individual lines processed in turn
     #   Due to the file size, reading line by line might be more efficient
-    def prepare_data(self, input_path, output_path):
-        with open(input_path, 'r', ) as input_file:
-            reader = csv.reader(input_file)
+    def prepare_data(self, path):
+        with open(path, 'r', ) as input_file:
             first_row = input_file.readline()
-            self._check_first_row(first_row)
+            self._check_first_row(first_row)    # Using input_file to treat row as 1 string item
 
             second_row = input_file.readline()
-            try:
-                spectrum = int(second_row)
-            except ValueError:
-                error_message = (f"Unexpected second row of file.\n"
-                                 f"Expected: Integer (first spectrum number)\n"
-                                 f"Actual:\t\t{second_row}\n")
-                raise RuntimeError(error_message)
+            spectrum = self._check_first_row(second_row)
 
+            reader = csv.reader(input_file)
             processed_data = []
-            for i, row in enumerate(input_file):
-                float_list = self.raw_row_to_float_list(row)
+            for i, row in enumerate(reader):
+                float_list = [float(item) for item in row]  # Ensures
                 if len(float_list) is 1:
-                    spectrum = int(row)
+                    spectrum = int(row[0])
                 else:
                     processed_data.append([spectrum]+float_list)
-                    # print([spectrum]+float_list)
-                # if i > 10000:
-                #     break
 
             return pd.DataFrame(processed_data, columns=self.columns)
 
-
-
-            # with open(output_path, 'w') as output_file:
-            #     writer = csv.writer(output_file, delimiter=",")
-            #     for i, row in enumerate(input_file):
-            #         writer.writerow(self.raw_row_to_float_list(row))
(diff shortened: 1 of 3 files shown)


JEThompson97:
_Note:_ One assertion from my tests works locally but not on Travis.
I've commented it out for now to allow Travis to pass.

The issue on Travis is that it thinks the size of the final DataFrame is 0 (it should be one). Not sure how to investigate why it thinks this..?

JackEAllen:
> _Note:_ One assertion from my tests works locally but not on Travis.
> I've commented it out for now to allow Travis to pass.
> 
> The issue on Travis is that it thinks the size of the final DataFrame is 0 (it should be one). Not sure how to investigate why it thinks this..?

Thanks @JEThompson97  I will have a look into it. Do you have the exact error/warning message Travis produced? I can't find the previous travis build where this fail occurred and it may help in shedding light as to what the problem is.

JEThompson97:
@JackEAllen Here's a link (+screenshot if the link doesn't work)

https://travis-ci.org/github/ISISScientificComputing/autoreduce/builds/669224965?utm_medium=notification&utm_source=github_status

image

JackEAllen:
@JEThompson97 I think it's something to do with the use of builtins.open, I believe what's happening is that when you mock prep.prepare_data() into a dataframe as such: 
```
with patch("builtins.open", mock_open(read_data=self.valid_data)):
        data_frame = prep.prepare_data("")
```
Travis is seeing this as an empty dataframe instead of returning as intended:
```
   Spectrum    X    Y    E
0         2  3.1  3.2  3.3
```
I can't think of any quick fixes for this off of the top of my head. Could you look into this and see if you are able to fix it easily? 

If you struggle, may be worth leaving out for the time being.

@ElliotAOram  Given this problem is only with the Travis build and not when running locally, should no fix be found, do you know if we can make Travis ignore this mock?

JackEAllen:
Other than the above issue, everything looks good  for merging 👍

JEThompson97:
@JackEAllen Thanks for looking into this!
I'm going to try playing around with how I've made the mock and see if it resolves it any. Will keep you in the loop

ElliotAOram:
As there are some posts online about difficult mocking patch open in travis, let's just add a testing file and actually open it to be sure

JEThompson97:
@ElliotAOram 
Rather than upload a standalone test file, I've told the `setUp` and `tearDown` methods to create and remove one respectively.
Travis seems to be OK with it if you are :)

plotting/prepare_data.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        processed_data = []
+        with open(path, 'r') as input_file:
+            self._check_first_row(input_file.readline())
+            spectrum = self._check_second_row(input_file.readline())
+
+            reader = csv.reader(input_file)
+            for row in reader:
+                float_list = [float(item) for item in row]
+                if len(float_list) is 1:     # pylint:disable=literal-comparison
```
ElliotAOram, line 44: Should use `len(float_list) == 1:`
`is` should be used for direct comparison by reference. see SO article for a pretty good explanation https://stackoverflow.com/questions/47778889/pylint-complains-about-comparing-a-string-to-a-literal-with-is
The example here is with strings, but the same applies

ElliotAOram: ps. this is also why pylint is complaining

PR #503, opened 2020-03-31
Database variable redundancy

### LINKED ISSUE #285: Remove data redundancy in the database
Issue raised by: [developer]

### What?
The database table `reduction_varaiables_variable` has a massive amount of duplicated data. It appears that the variables used to run a particular reduction job are stored even if they are the same as the variables used before. Given that we already have a linking table (`reduction_variables_runvariable`) this is needless duplication and is not only filling up our database, but also bad design. Instead of mindlessly adding new variables we should instead check if for pre-existing variables that are identical to the current variables.

### Where?
Will have to investigate where, as well as ensure there is not a good reason why, this information is duplicated - not reused. 

### How?
Discovered while writing a script to remove records manually and failing the foriegn key constraint in `reduction_variables_runvariable`. Subsequently looking at the database and discovering the duplication. 

### Reproducible?
This is currently the state live on production

Ensure that when a job is submitted with the same parameters as a different job, the information is not duplicated in the database:
1. Submit a run using manual submission
2. Check the variables have been added to the reduction_variables_variable table in the database
3. Submit the same run
4. Check that the variables have `NOT` been added to the database but instead the `reduction_variables_runvariable` table links the reduction_id to the variables from the previous run

Note: This should be in place before next cycle starts (It will be too difficult to attempt to rectify this in the current database)

ElliotAOram:
# Do not merge still in progress
### Additional comments
 

Fixes #285

### DIFF
--- a/queue_processors/queue_processor/queueproc_utils/tests/test_variable_utils.py
+++ b/queue_processors/queue_processor/queueproc_utils/tests/test_variable_utils.py
@@ -21,14 +21,19 @@ from queue_processors.queue_processor.orm_mapping import RunJoin, Variable, Inst
 class TestVariableUtils(unittest.TestCase):
 
     def setUp(self):
-        self.valid_variable = Variable(name='test',
-                                       value='value',
-                                       type='text',
-                                       is_advanced=False,
-                                       help_text='help text')
+        self.valid_variable = Variable(name='bool_variable',
+                                       value='True',
+                                       type='boolean',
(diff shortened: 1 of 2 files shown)


ElliotAOram:
This will never work with current DB implementation

ElliotAOram:
In order to properly resolve this issue we will need to make a change in the database as well as the changes in the this PR. as such I will not delete the branch for this PR just yet

PR #506, opened 2020-04-14
Replace 'MagicMock's with '@patch' where appropriate

### LINKED ISSUE #505: Replace MagicMocks with @patch

### What?
- Most of our test code uses the `@patch` decorator to mock out classes, method, etc.
- However, some use `MagicMock`s instead to do the same thing.
- After a discussion with @ElliotAOram, we decided that for consistency, it'd be good to use `@patch` in all instances where possible.
  * `@patch` is favoured as it nicely separates the mocking setup from the test code
- Therefore, we should refactor instances of `MagicMock` into `@patch`.

### Where?
- Test classes

### How?
- Use of an alternative mocking method

- All instances of `MagicMock` should be removed except for instances where `@patch` cannot provide the same functionality.
- All tests should still pass following the change.

JEThompson97:
### Summary of work
- For each instance of `MagicMock` in the autoreduction code base: I assessed whether it was appropriate and whether it could be replaced with a `@patch` decorator.
- For those that could/were appropriate to be replaced, this PR implements the replacement.

- Cases where `MagicMock`s remain are where they are used appropriately to contain specific mock functionality, used across multiple tests.

### How to test your work
- Review that MagicMocks are only used where seemingly appropriate.
- Ensure all tests are still passing.

### Additional comments
- If the reviewer(s) agree: The remaining instances of `MagicMock` can be used as an exemplar appropriate way to use them in our code base.

Fixes #505


JEThompson97:
Not sure how to address the coveralls failure - the only "new code" I've added here is _copyright_ comment blocks..

ElliotAOram:
coveralls failure doesn't seem to add up.  as such I'm willing to ignore it as it is such a smaller % difference.

PR #510, opened 2020-04-17
Basic AMQ message class

### LINKED ISSUE #507: Implement Basic AMQ message class
Issue raised by: [developer]

### What?
In order to more formally represent the data passed between Queues in the system we should ahve a class responsible for serialisation and validation of the data. 

The first stage is to implement the class and the tests, this should include the functionality for the (un)serialisation but does not need to include the validation at this stage as this is not present in the current code base (we'll add the validation once we have managed to integrate it into the code base).

To start with we should simply make and test the class. 

See design below:
image

### Where?
New class

Check that the class is implemented according to the design and has related tests

ElliotAOram:
### Summary of work
Added a class that can handle the containment of an AMQ message and act as a wrapper to the meta data stored within. The class can also be serialised and un-serialised (whcich must be done before sending.

### How to test your work
* Code review
* Assess quality of unit tests

### Additional comments
Note: this is not current implemented in the places it will be used in the code base. This is ONLY a review of the containing class for AMQ messages. Integration will be done in a  separate PR and most likely done one area at a time.

Fixes #507 

### DIFF
--- a/message/job.py
+++ b/message/job.py
@@ -6,54 +6,39 @@
 # ############################################################################### #
 """ Represents the messages passed between AMQ queues"""
 import json
+import attr
 
 
 # pylint:disable=too-many-instance-attributes
+@attr.s
 class Message:
     """
     A class that represents an AMQ Message, these can be both serialised and un-serialised
     for sending messages to and from AMQ
     """
-
-    def __init__(self, serialised_object=None, overwrite=True):
-        """
-        If a serialised object is supplied then the class will be populated
-        :param serialised_object: a serialised json object (json.dumps)
-        :param overwrite: Should we overwrite existing variables?
-        """
-        self.description = None
-        self.facility = None
-        self.run_number = None
-        self.instrument = None
-        self.rb_number = None
-        self.started_by = None
-        self.file_path = None
-        self.overwrite = None
-        self.run_version = None
-        self.job_id = None
-        self.reduction_script = None
-        self.reduction_arguments = None
-        self.reduction_log = None
-        self.admin_log = None
-        self.return_message = None
-        self.retry_in = None
-        if serialised_object:
-            self.un_serialise(serialised_object, overwrite)
-
-    def to_dict(self):
-        """
-        Create a dictionary of all the class member variables
-        :return: member variables dictionary
-        """
-        return {key: value for key, value in self.__dict__.items()
-                if not key.startswith('__') and not callable(key)}
+    description = attr.ib(default=None)
+    facility = attr.ib(default=None)
+    run_number = attr.ib(default=None)
+    instrument = attr.ib(default=None)
+    rb_number = attr.ib(default=None)
+    started_by = attr.ib(default=None)
+    file_path = attr.ib(default=None)
+    overwrite = attr.ib(default=None)
+    run_version = attr.ib(default=None)
+    job_id = attr.ib(default=None)
+    reduction_script = attr.ib(default=None)
+    reduction_arguments = attr.ib(default=None)
+    reduction_log = attr.ib(default=None)
+    admin_log = attr.ib(default=None)
+    return_message = attr.ib(default=None)
+    retry_in = attr.ib(default=None)
 
     def serialise(self):
         """
         serialised member variables as a json dump
         :return: json dump of a dictionary representing the member variables
         """
-        return json.dumps(self.to_dict())
+        return json.dumps(attr.asdict(self))
 
     def un_serialise(self, serialised_object, overwrite=True):
         """
@@ -62,7 +47,7 @@ class Message:
         :param overwrite: Should we overwrite any existing variables?
         """
         un_serialised_dict = json.loads(serialised_object)
-        self_dict = self.to_dict()
+        self_dict = attr.asdict(self)
         for key, value in un_serialised_dict.items():
             if key in self_dict.keys():
                 current_value = self_dict[key]
--- a/message/job.py
+++ b/message/job.py
@@ -13,7 +13,7 @@ import attr
 @attr.s
 class Message:
     """
-    A class that represents an AMQ Message, these can be both serialised and un-serialised
+    A class that represents an AMQ Message, these can be both serialized and deserialized
     for sending messages to and from AMQ
     """
     description = attr.ib(default=None)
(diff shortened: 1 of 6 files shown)


ElliotAOram:
Should look into the use of `attrs` to get as dict for free. http://www.attrs.org/en/stable/examples.html#converting-to-collections-types

message/job.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        Create a dictionary of all the class member variables
+        :return: member variables dictionary
+        """
+        return {key: value for key, value in self.__dict__.items()
+                if not key.startswith('__') and not callable(key)}
+
+    def serialise(self):
+        """
+        serialised member variables as a json dump
```
JEThompson97, line 38: Very minor: There's some inconsistency in whether capitals are used at the beginning of (i) descriptions (ii) param: (iii) return:

message/job.py
```diff
@@ -0,0 +1,71 @@
+# ############################################################################### #
+# Autoreduction Repository : https://github.com/ISISScientificComputing/autoreduce
+#
+# Copyright &copy; 2020 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+# ############################################################################### #
+""" Represents the messages passed between AMQ queues"""
```
JEThompson97, line 7: ~~""" Represents the messages passed between AMQ queues"""~~
"""
Represents the messages passed between AMQ queues
"""

message/job.py
```diff
@@ -0,0 +1,71 @@
...
+import json
+
+
+# pylint:disable=too-many-instance-attributes
+class Message:
+    """
+    A class that represents an AMQ Message, these can be both serialised and un-serialised
```
JEThompson97, line 16: A class that represents an AMQ **Message.**
Messages can be both serialised and unserialised
    for sending messages to and from AMQ

_(2 sentences rather than 1)_

ElliotAOram:
When we add validation to this we can also use attrs to help us: http://www.attrs.org/en/stable/examples.html#validators

message/tests/test_job.py
```diff
@@ -0,0 +1,97 @@
...
+    def test_to_dict(self):
+        """
+        Test the class can be correctly converted to a dictionary
+        """
+        actual = self.msg.to_dict()
+        for key, value in actual.items():
+            self.assertIn(key, self.key_list)
+            self.assertIsNone(value)
+
+    def test_serialise(self):
```
JEThompson97, line 66: As I understand it, this tests that _empty_ variables remain _empty_ when they are serialised.
Is it worth testing this where the variables values are _not None_?

message/tests/test_job.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        Test the class can be serialised and retain the variables values
+        """
+        serialised = self.msg.serialise()
+        actual = json.loads(serialised)
+        for key, value in actual.items():
+            self.assertIn(key, self.key_list)
+            self.assertIsNone(value)
+
+    def test_un_serialise_overwrite_true(self):
```
JEThompson97, line 76: Should there be a test for _unserialise_overwrite__**false**?
Where some variables were None, some were not, and a check that only those that were None are updated

message/job.py
```diff
@@ -0,0 +1,56 @@
...
+    def serialise(self):
+        """
+        serialised member variables as a json dump
+        :return: json dump of a dictionary representing the member variables
+        """
+        return json.dumps(attr.asdict(self))
+
+    def un_serialise(self, serialised_object, overwrite=True):
+        """
+        un serialise an object and use its values to populate member variables in this class
```
JEThompson97, line 45: Although it's questionable whether "unserialise" is a real word, to me it looks more right than "un-serialise" or "un serialise".
Might just be me

message/job.py
```diff
@@ -0,0 +1,56 @@
...
+    retry_in = attr.ib(default=None)
+
+    def serialise(self):
+        """
+        serialised member variables as a json dump
+        :return: json dump of a dictionary representing the member variables
+        """
+        return json.dumps(attr.asdict(self))
+
+    def un_serialise(self, serialised_object, overwrite=True):
```
JEThompson97, line 43: _(I wrote a lengthy message about using a different name to `un_serialise` but GitHub unfortunately deleted it..
Could we discuss this one over zoom instead?)_

ElliotAOram:
@JEThompson97 this is ready for re-review. I'm thinking it might be worth removing the serialized, deserialize functions to a separate class and having them as functionality on the queue client given that they are simply calling `json.loads/dumps` but I think for now we should leave it here and work out how to best refactor this once it's in use? 
Either that or, have a abstract class that all messages inherit fro that has the simple json style serialization and this can then be overridden if we ever have a message that has more complex serialization requirements. Although I currently don't see a case where we would need another type of message in a single implementation of the autoreduction system. I believe the only time the message may differ is if the Autoreduction implementation was used at another facility (but most likely they would still only have one message).

message/tests/test_job.py
```diff
@@ -0,0 +1,121 @@
...
+    def test_populate_from_dict_overwrite_true(self):
+        """ Overwritten from a dictionary values"""
+        actual = self.empty_msg
+        actual.populate(self.populated_dict, overwrite=True)
+        self.assertEqual(attr.asdict(actual), self.populated_dict)
+
+    def test_populate_from_dict_overwrite_false(self):
+        """ Do not overwrite from dictionary """
+        self.populated_dict['job_id'] = 123
+        self.populated_dict['rb_number'] = 33333
```
JEThompson97, line 94: These changes to populated_dict - do they persist into tests which follow this?
If so, wouldn't that make the ordering of tests affect their behaviour? (although it might not actually affect whether they pass/fail, the values tests use will be depend on the order they're run, which seems a little iffy)

JEThompson97: You've accounted for this below for `test_populate_from_serialized_overwrite_false`, where you've used a copy instead of the persistent class variable

message/tests/test_job.py
```diff
@@ -0,0 +1,121 @@
...
+                           'reduction_log': None,
+                           'admin_log': None,
+                           'return_message': None,
+                           'retry_in': None}
+        self.populated_msg = Message(run_number=11111, rb_number=22222, description='test message')
+        self.populated_dict = copy.deepcopy(self.empty_dict)
+        self.populated_dict['run_number'] = self.populated_msg.run_number
+        self.populated_dict['rb_number'] = self.populated_msg.rb_number
+        self.populated_dict['description'] = self.populated_msg.description
+        self.populated_keys = ['run_number', 'rb_number', 'description']
```
JEThompson97, line 47: ### This comment links to the one below

You've 'setUp' various Messages and dicts which you've used in various combinations across the tests and in some cases made copies where modifications occur.

I'm just wondering why these aren't instead instantiated on-demand as and when needed by individual tests. This would avoid the need to make copies, and guarantees tests do not impact each other (e.g. modifying a variable in a test - see other comment below where I believe this occurs)

This would just mean refactoring the instantiations above into methods e.g. `create_populated_msg`, and calling these within the tests as needed (replacing any copies). `See \utils\clients\tests\test_sftp_client.py` for an example of this.

Happy to discuss

message/tests/test_job.py
ElliotAOram: These won't persist as setUp is called before each test. However, I should still use a copy here for safety. I'll also split up the setup code into case and class specific

message/tests/test_job.py
ElliotAOram: Your suggestion is a better way of doing it given that not all tests require all the setup elements.  will make this change

message/tests/test_job.py
JEThompson97: Ahh I didn't realise that about `setUp`, good to know!

ElliotAOram:
@JEThompson97 ready for review again

message/tests/test_job.py
```diff
@@ -21,101 +20,132 @@ class TestMessage(unittest.TestCase):
...
-                           'return_message': None,
-                           'retry_in': None}
-        self.populated_msg = Message(run_number=11111, rb_number=22222, description='test message')
-        self.populated_dict = copy.deepcopy(self.empty_dict)
-        self.populated_dict['run_number'] = self.populated_msg.run_number
-        self.populated_dict['rb_number'] = self.populated_msg.rb_number
-        self.populated_dict['description'] = self.populated_msg.description
-        self.populated_keys = ['run_number', 'rb_number', 'description']
+    @staticmethod
+    def _empty():
```
JEThompson97, line 24: Could perhaps do with a bit of docstring for `_empty()` and `_populated()`

message/tests/test_job.py
```diff
@@ -21,101 +20,132 @@ class TestMessage(unittest.TestCase):
...
         actual.populate(source=serialized, overwrite=False)
-        self.assertEqual(actual.rb_number, self.populated_msg.rb_number)
+        self.assertEqual(actual.rb_number, original.rb_number)
         self.assertEqual(actual.job_id, 123)
 
     def test_invalid_serialized(self):
         """ Test raise exception for invalid serialized object """
         serialized = 'test'
-        self.assertRaises(ValueError, self.empty_msg.populate, serialized)
+        empty_msg, _ = self._populated()
```
JEThompson97, line 150: Should empty_msg be assigned from _empty()?
i.e. `empty_msg, _ = self._empty()` 

(Not that this affects the test much, since whether or not the message is populated vs empty shouldn't have an effect on this specific test)

ElliotAOram:
@JEThompson97  thanks again for comments. Should be ready for another look

Issue #517, opened 2020-04-23
Remove the use of raw stomp access to AMQ


ElliotAOram:
Issue raised by: [developer]

### What?
In several places identified by @JEThompson97 we are still using raw stomp code to access activeMQ. As we move towards a harmonised system for Messages #512 it is important we also harmonise the way we access ActiveMQ. 

### Where?
The places we are using raw stomp access are:
* [x] `queue_processors/autoreduction_processor/autoreduction_processor.py` - #517 
* [x] `queue_processors/autoreduction_processor/post_process_admin.py` - #525 
* [x] `WebApp\autoreduce_webapp\autoreduce_webapp\queue_processor.py` - #538 

### How to test the issue is resolved
* Ensure all use of raw stomp connections is removed from the code base other than in the `QueueClient`.
* Ensure the functionality of each of the components above still works which should be done by:
  * Resubmitting a run from the WebApp
  * Ensuring that run successfully reaches the autoreduction processor
  * Ensure the run is successfully reduced and finishes in the completed status in the autoreduction database
* Ensure the unit tests are updated if required

PR #518, opened 2020-04-23
Remove use of raw stomp in autoreduction_processor.py

ElliotAOram:
### Summary of work
* Changed the AMQ subscription in the autoreduction_processor to be done via the QueueClient. 

### How to test your work
* Perform a manual submission on the development environment (this code is currently on there but might be worth checking before you do this).
* Ensure that the job does not get stuck in the queued state.

### Additional comments
* Before merging I might look into cleaning this up a little. There is currently functionality in the QueueClient for subscription but I wasn't able to get this to work initially in this code. As such I will look into if I can get it to work or remove it. 

EDIT: Will wait till I have done the rest of the stomp removing until I make any decisions about this.

Part of #517 

### DIFF
--- a/queue_processors/autoreduction_processor/autoreduction_processor.py
+++ b/queue_processors/autoreduction_processor/autoreduction_processor.py
@@ -19,6 +19,7 @@ import stomp
 from queue_processors.autoreduction_processor.autoreduction_logging_setup import logger
 # pylint:disable=no-name-in-module,import-error
 from queue_processors.autoreduction_processor.settings import MISC, ACTIVEMQ
+from utils.clients.queue_client import QueueClient
 
 
 class Listener:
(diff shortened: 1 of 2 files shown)


queue_processors/autoreduction_processor/autoreduction_processor.py
```diff
@@ -137,33 +136,22 @@ def __init__(self):
...
-                                 header={'activemq.prefetchSize': '1'})
-            logger.info("[%s] Subscribing to %s", self.consumer_name, queue)
-        logger.info("Successfully subscribed to all of the queues")
-
-        # activemq_client = QueueClient()
-        # activemq_client.connect()
+        activemq_client = QueueClient()
+        connection = activemq_client.connect()
 
         # Create event listener
```
JEThompson97, line 148: Interesting that this code was already here but commented out...

ElliotAOram: Yes, they were there from my first attempt a long time ago to get this right

PR #525, opened 2020-04-28
Remove raw stomp use in post process admin

ElliotAOram:
### Summary of work
* Remove raw stomp use in `post_process_admin.py`
* Change tests to use QueueClient in mocks
* This also clears up the issue from #518 which caused the subscription in the `autoreduction_queue_processor` to fail.

### How to test your work
* Ensure unit tests pass
* Code is currently on development node (might be worth checking before running) so it should be possible to resubmit a run and if it completes with no issue then this passes

Part of #517


ElliotAOram:
This PR should also resolve the recently emerged pylint issue

queue_processors/autoreduction_processor/test_settings.py
```diff
@@ -14,19 +14,6 @@
 
 from utils.project.structure import get_project_root
 
-# ActiveMQ
-ACTIVEMQ = {
-    "brokers": "127.0.1.1:61613",
-    "amq_queues": ["/queue/ReductionPending"],
-    "amq_user": "admin",
-    "amq_pwd": "admin",
-    "postprocess_error": "/queue/ReductionError",
```
JEThompson97, line 23: Interesting that previously 2 keys mapped to the same queue (`/queue/ReductionError`)
Was this for readability reasons?

ElliotAOram: Not sure. Most likely an oversight.

PR #529, opened 2020-04-29
Add Database Software Table

### LINKED ISSUE #528: Add and integrate software table into autoreduction
Issue raised by: [developer]

### What?
As part of the new database design we need to add a table to capture the software used for reduction and the version. As well as adding the database table, we will also need to populate it from the queue processors. This issue should cover both adding the table and ensure it is populated. 

See design image here (where `Job` is current ReductionRun`): 
image

### Where?
Model - `WebApp/autoreduce_webapp/reduction_viewer/models.py`
Integration in QP - `queue_processors/autoreduction_processor/post_process_admin.py` - `reduce.py`

### How?
Part of  move to new database schema

* Will require a unit tests to ensure:
  * That the correct version of Mantid is found
  * That the software (+version) is added (if it does not exist)  Software table
  * That the existing Software object is returned if it already exists
  * That the reduction job is associated with the software it uses

ElliotAOram:
# Summary of work
* Add the Software table to the database

# TODO
* Hook up to queue processor
* Ensure we use reference to original
* Display software version in the webapp (optional extra to this issue)
* Add release notes

### How to test your work
* Check out this code on the development nodes
* Submit a run
* Ensure the run ends up in the database and the software field is populated
* Repeat this, and ensure a new entity for software is NOT added instead the new reduction job references the existing (created by the first) software record.

Fixes #528 

### DIFF
--- a/WebApp/autoreduce_webapp/reduction_viewer/models.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/models.py
@@ -46,6 +46,17 @@ class Status(models.Model):
         return u'%s' % self.value
 
 
+class Software(models.Model):
+    """ Represents the software that is used for data reduction """
+
+    name = models.CharField(max_length=100, blank=False)
+    version = models.CharField(max_length=20, blank=False)
+
+    def __unicode__(self):
+        """ :return: unicode string of: name=value """
+        return u'%s=%s' % (self.name, self.version)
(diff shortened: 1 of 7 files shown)


ElliotAOram:
Can't reliably write tests for integration of this into QP until we switch ORM

PR #533, opened 2020-04-30
Manual Submission to check DB before calling ICAT

### LINKED ISSUE #526: Allow manual submission to find data file location from DB
Issue raised by: [developer]

### What?
Currently the `manual_submission.py` script use ICAT to find the data location. In the case that autoreduction has already reduced the file, the data location will be in the database. As such, to avoid unrequired calls to ICAT (and as a alternative for when ICAT is down or does not contain a data file we are looking for), we should expand manual_submission to first (before trying ICAt) look in the autoreduction database for the existing data file location

### Where?
`scripts/manual_operations/manual_submission.py`

### How?
Noticed this while ICAT was down for maintenance

* Change the ICAT login credentials to be incorrect
* Run the manual_submission script on a run that already existing in the database
* Ensure the run is resubmitted without error

JEThompson97:
### Summary of work
- This PR changes the _data file location & RB-number retrieval_ functionality (fka `get_data_file`) of the `manual_submission` script (MSS) to first check the Auto-reduction database (i.e. check whether the file has already been reduced) before resorting to calling ICAT for this information.
- This avoids unnecessary ICAT calls and enhances autoreduction's (AR) independence from ICAT

### How to test your work
- Ensure the following cases are true (which can be checked by the print statements produced):
1. When a file exists in AR, MSS retrieves the necessary data from there
2. When a file does not exists in AR, MSS calls ICAT
3. When a file exists in neither AR or ICAT, MSS exits gracefully as usual

### Additional comments
- I've used raw SQL because we plan to change our ORM soon.
  * It is easier to transition from raw SQL to a new ORM than from an existing ORM.

Fixes #526 

### DIFF
--- a/scripts/manual_operations/manual_submission.py
+++ b/scripts/manual_operations/manual_submission.py
@@ -15,6 +15,9 @@ import argparse
 
 # The below is only a template on the repo
 # pylint: disable=import-error, no-name-in-module
+from icat import ICATSessionError
+
+from utils.clients.connection_exception import ConnectionException
 from utils.clients.icat_client import ICATClient
 from utils.clients.queue_client import QueueClient
 from utils.clients.database_client import DatabaseClient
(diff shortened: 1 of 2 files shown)


JEThompson97:
@ElliotAOram Ready for review whenever

PR #539, opened 2020-05-01
django orm outside of webapp

### LINKED ISSUE #535: Change ORM to Django only
Issue raised by: [developer]

### What?
As part of the database re-design we also are looking to replace the SQLAlchemy ORM with just Django ORM.
This task could be massive, so I suggest this issue should cover just the creation of the Django ORM interface to be used outside of the webapp.
Follow up issues should include:
* Integrate Django ORM to DatabaseClient
* Replace SQLAlchemy use throughout autoreduction
  * Probably break this up into modules

### Where?
Need to think about where the new ORM should live. Perhaps a new `model` or `database` module?

* Test that the new ORM is tested and works externally to the webapp
  * This will most likely be assessed through code review and the accompanying tests

ElliotAOram:
### Summary of work
* Add functionality to access Django ORM
* Add `model` package we should also refactor `Message` into here (#540)
* Add test cases 

### How to test your work
* Ensure test cases and travis run
* You could try and write a small script to check this is working:
```
from model.database import DjangoORM
orm = DjangoORM()
orm.connect()
gem = orm.data_model.Instrument.objects.filter(name='GEM').first()
print(gem)
```

### Additional comments

Fixes #535


ElliotAOram:
Ready for review

model/database/orm.py
```diff
@@ -0,0 +1,100 @@
...
+            sys.path.append(path)  # pragma: no cover
+
+    @staticmethod
+    def setup_django():
+        """
+        Use the WebApp settings to initialise a django instance that can be used for model access
+        """
+        try:
+            # import here to avoid failure without calling add_webapp_path first
+            # pylint:disable=import-outside-toplevel
```
JEThompson97, line 53: I wonder if these pylint disables which span across multiple levels ought to be applied at a class level?

model/database/orm.py
```diff
@@ -0,0 +1,100 @@
...
+        if path not in sys.path:
+            # Can't properly replicate this in a testing environment due to travis
+            # constraint on sys path access and mocking would break underlying dependency
+            # but it's simple enough we don't need to worry about testing it
+            sys.path.append(path)  # pragma: no cover
+
+    @staticmethod
+    def setup_django():
+        """
+        Use the WebApp settings to initialise a django instance that can be used for model access
```
JEThompson97, line 49: Maybe a note to say that "add_webapp_path should always be called before this method" or similar (if I'm remembering correctly that that is the case)

model/database/orm.py
```diff
@@ -0,0 +1,100 @@
...
+        self.setup_django()
+        try:
+            if not self.data_model:
+                self._get_data_model()
+            self.data_model.Instrument.objects.first()
+            if not self.variable_model:
+                self._get_variable_model()
+            self.variable_model.Variable.objects.first()
+        # pylint:disable=bare-except
+        except:
```
JEThompson97, line 98: Would a comment to explain why you've used a _bare except_ be useful here for the benefit of future devs

model/database/tests/test_orm.py
```diff
@@ -0,0 +1,90 @@
...
+        # pylint:disable=protected-access
+        model = orm._get_variable_model()
+        actual = model.Variable.objects.filter(name='bool_variable').first()
+        self.assertIsNotNone(actual)
+        self.assertEqual(actual.name, 'bool_variable')
+        self.assertEqual(actual.type, 'boolean')
+
+    def test_connect(self):
+        """
+        Test: The orm is exposed and populated as member variables
```
JEThompson97, line 74: _nit_
~~The orm is exposed and populated as member variables~~
> The **DjangoORM instance** is exposed and populated as member variables

(or similar) just to be explicit we're referring to object and not the underlying ORM

model/database/tests/test_orm.py
```diff
@@ -0,0 +1,90 @@
...
+        When: add_webapp_path is called more than once
+        """
+        DjangoORM.add_webapp_path()
+        expected = sys.path
+        DjangoORM.add_webapp_path()
+        self.assertEqual(expected, sys.path)
+
+    def test_get_data_model(self):
+        """
+        Test: That the data model can be accessed
```
JEThompson97, line 46: _nit_
> Test: ~~That the~~ **The** data model can be accessed

Just for consistency with other _Test:_ comments

model/database/tests/test_orm.py
```diff
@@ -0,0 +1,90 @@
...
+        orm.connect()
+        # pylint:disable=protected-access
+        model = orm._get_data_model()
+        actual = model.Instrument.objects.filter(name='GEM').first()
+        self.assertIsNotNone(actual)
+        self.assertEqual(actual.name, 'GEM')
+
+    def test_get_variable_model(self):
+        """
+        Test: That the variable model can be accessed
```
JEThompson97, line 59: See above _nit_ comment ^

model/database/tests/test_orm.py
```diff
@@ -0,0 +1,90 @@
...
+        orm = DjangoORM()
+        self.assertTrue(orm.connect())
+        self.assertIsNotNone(orm.data_model.Instrument.objects.first())
+        self.assertIsNotNone(orm.variable_model.Variable.objects.first())
+
+    @patch('model.database.orm.DjangoORM.data_model')
+    def test_failed_connect(self, mock_data):
+        """
+        Test: False is returned
+        When: Connect fails to retrieve data (raises an exception)
```
JEThompson97, line 86: _nit_
> When: ~~Connect~~  connect fails to retrieve data (raises an exception)

To match the method's casing (as per comment convention)

model/database/orm.py
ElliotAOram: Probably not in this case. I generally take an approach depending on the severity of the disable. In this case (toplevel imports) is quite serious and as such I want to ensure that we are only disabling this at the point of use. 

Other things such as line too long and docstrings are less important so can be used at class level

Issue #520, opened 2020-04-23
Ensure all AMQ interactions via QueueClient are tested


JEThompson97:
Issue raised by: [developer]

### What?
- Part of #512
- Now that instances of AMQ interaction have been identified (#519), one of the next steps is to ensure all interactions via QueueClient are tested.

### Where?
- Specifically, tests need to be make for QueueClient usage in the following locations:
  * queue_processor.py
  * messaging_utils.py

### How?
After having identified untested AMQ interaction via QueueClient as part of #519 

### How to test the issue is resolved
- Ensure every instance of AMQ interaction via QueueClient has tests around it

PR #541, opened 2020-05-06
[DNM] Update calls to QueueClient.send to use Message class instead of json dumps

### LINKED ISSUE #512: Integrate AMQ Message into autoreduction
Issue raised by: [developer]

### What?

### Where?
* monitors/run_detection.py
* queue_processor
* autoreduction_processor
* webapp

### How?

* All test must pass

JEThompson97:
# DO NOT MERGE please
This PR is being replaced by smaller version
___
### Summary of work
- This PR updates all calls to `QueueClient.send`, specially the `message` argument
- Previously, usages of QueueClient.send populated the message argument with a json.dump of the data to be sent
- This has been replaced with a new instance of the new Message class (#507) across all usages and tests for said usages

#### Other small additional changes:
- **post_process_admin usage of 'send'**
  * During this work I found that post_process_admin was calling send on `QueueClient._connection` (the stomp instance) rather than on QueueClient itself (which inturn calls stomp after connecting and extracting the message content).
  *As such, I updated post_process_admin to use the QueueClient rather than the conneciton
- **queue_processor tests**
  * I also found that a 2 additional usages of `QueueClient.send` which I had not spotted in previous work (#520) in `queue_processor`.
  * I updated these instances to use the Message class like the rest
  * I added a test for 1 of these (c57d6da) but the other usage is in the 100-line `data_ready` method, and so I could not elegantly create a test for this (doing so would be a large task, involving mocking out huge sections of code)

### How to test your work
- Ensure all tests pass
- Code review

### Additional comments
- This is a large PR which spans many important areas of the code base...

Fixes #512

### DIFF
--- a/WebApp/autoreduce_webapp/reduction_variables/utils.py
+++ b/WebApp/autoreduce_webapp/reduction_variables/utils.py
@@ -18,6 +18,8 @@ import sys
 import chardet
 import importlib.util as imp
 
+from message.job import Message
+
 sys.path.append(os.path.join("../", os.path.dirname(os.path.dirname(__file__))))
 os.environ["DJANGO_SETTINGS_MODULE"] = "autoreduce_webapp.settings"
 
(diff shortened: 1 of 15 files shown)


ElliotAOram:
@JEThompson97 the easiest way to get an idea of what might be going wrong in the system tests is to print the log files after the system tests run.

JEThompson97:
@ElliotAOram Thanks for the heads up, I'll give that a try in future
In this case specifically, the test failures pointed to something about the way I had integrated the tests which needs refactoring.
I'm in the process now of trying to implement the best solution

PR #542, opened 2020-05-07
Integrate Message class to manual_submission

JEThompson97:
### Summary of work
- This PR begins the task of Integrating the Message class into the code base (#512)
- Specifically, this PR does the following:
  * Updates `QueueClient.send` to accommodate Message instances being received as the message argument (as well as json dumps which have been use up until now)
  * Adds logging to the Message class such that when `populate` encountered a key which does not match an attribute, a warning is logged
  * Refactors the `manual_submission` call to `QueueClient.send` to use the Message class instead of a json dump (+ tests)
  * Adds a test for a usage of `QueueClient.send` that was missed.
    * note: there is another usage in the same class but it is within the 100-line `data-ready` method.  Due to it's size, adding tests to this method should be separate chunk of work.

### How to test your work
- Ensure all tests pass
- Code review

Part of #512

### DIFF
--- a/message/job.py
+++ b/message/job.py
@@ -8,8 +8,15 @@
 Represents the messages passed between AMQ queues
 """
 import json
+import logging
+
 import attr
 
+from utils.project.static_content import LOG_FORMAT
+from utils.project.structure import get_log_file
+
+logging.basicConfig(filename=get_log_file('job.log'), level=logging.INFO,
+                    format=LOG_FORMAT)
 
 # pylint:disable=too-many-instance-attributes
 @attr.s
@@ -35,12 +42,14 @@ class Message:
     return_message = attr.ib(default=None)
     retry_in = attr.ib(default=None)
 
-    def serialize(self):
+    # Note: added indent to allow pretty-printing
+    #   Although I can work around needing this if not appropriate to have as argument
+    def serialize(self, indent=None):
         """
         Serialized member variables as a json dump
         :return: JSON dump of a dictionary representing the member variables
         """
-        return json.dumps(attr.asdict(self))
+        return json.dumps(attr.asdict(self), indent=indent)
 
     @staticmethod
     def deserialize(serialized_object):
@@ -71,3 +80,8 @@ class Message:
                 if overwrite or self_value is None:
                     # Set the value of the variable on this object accessing it by name
                     setattr(self, key, value)
+            else:
+                warning_message = (f"Unexpected key encountered during Message population: '{key}'."
+                                   f"Skipping this key...")
+                logging.warning(warning_message)
+                print(warning_message)  # Note: for debug purposes
--- a/message/job.py
+++ b/message/job.py
@@ -42,8 +42,6 @@ class Message:
     return_message = attr.ib(default=None)
     retry_in = attr.ib(default=None)
 
-    # Note: added indent to allow pretty-printing
-    #   Although I can work around needing this if not appropriate to have as argument
     def serialize(self, indent=None):
         """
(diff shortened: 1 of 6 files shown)


message/job.py
```diff
@@ -35,12 +42,14 @@ class Message:
     return_message = attr.ib(default=None)
     retry_in = attr.ib(default=None)
 
-    def serialize(self):
+    # Note: added indent to allow pretty-printing
```
ElliotAOram, line 45: if defaults to None this is fine, but just be sure to add it in the param docstring and give explanation

queue_processors/queue_processor/tests/test_queue_processor.py
```diff
@@ -42,3 +44,27 @@ def test_setup_connection(self, mock_sub_ar, mock_connect, mock_client):
...
+
+class TestListener(unittest.TestCase):
+    """
+    Exercises the Listener
+    """
+    def setUp(self):
+        self.rb_number = -1
+        self.reason = "test"
+
+    def test_construct_and_send_skipped(self):
```
ElliotAOram, line 57: 👍

utils/clients/tests/test_queue_client.py
```diff
@@ -102,16 +103,29 @@ def test_serialise_message(self):
 
     # pylint:disable=no-self-use
     @patch('stomp.connect.StompConnection11.send')
```
ElliotAOram, line 105: Can we change this to mock `utils.clients.queue_client.QueueClient.send`

utils/clients/tests/test_queue_client.py
```diff
@@ -102,16 +103,29 @@ def test_serialise_message(self):
...
         """
         client = QueueClient()
-        client.send('dataready', 'test-message')
+        client.send('dataready', 'raw_json_dump')
         (args, _) = mock_stomp_send.call_args
         self.assertEqual(args[0], 'dataready')
-        self.assertEqual(args[1], 'test-message')
+        self.assertEqual(args[1], 'raw_json_dump')
+
+    @patch('stomp.connect.StompConnection11.send')
```
ElliotAOram, line 117: same as above re-patch statement

ElliotAOram: It should be a like for like conversion (code shouldn't have to change other than what it is patching

queue_processors/queue_processor/tests/test_queue_processor.py
```diff
@@ -42,3 +44,27 @@ def test_setup_connection(self, mock_sub_ar, mock_connect, mock_client):
...
+        # pylint:disable=protected-access
+        """
+        Test: _data_dict['message'] is given a value,and the _data_dict is sent via the QueueClient
```
JEThompson97, line 60: Note to self: add space
> ..value, and..

utils/clients/tests/test_queue_client.py
JEThompson97: Yeah, I see what you mean, I've started testing inner workings of job.Message (which are already covered by test_job).
Will resolve in next commit.

utils/clients/tests/test_queue_client.py
JEThompson97: I'm a bit confused by this. `QueueClient.send` is what this test is testing, right? So by patching it out, wouldn't this test nothing except that "send is called when send is called"?

ElliotAOram: Yep you're right ignore me!

PR #544, opened 2020-05-07
Replace storage and use of ._connection outside of QueueClient

### LINKED ISSUE #543: autoreduction_processor uses QueueClient._connection (stomp) directly
Issue raised by: [developer]

### What?
- While working on #512, I found that `autoreduction_processor\post_process_admin` was calling send on `QueueClient._connection` (the stomp instance) rather than on QueueClient itself (which inturn calls stomp after connecting and extracting the message content).
- On further investigation, I found `autoreduction_processor\autoreduction_processor` similarly stores and uses QueueClient._connection stomp directly.
- post_process_admin.py and autoreduction_processor.py should be adjusted to use QueueClient rather than the connection

### Where?
`\queue_processors\autoreduction_processor\post_process_admin.py`
`\queue_processors\autoreduction_processor\autoreduction_processor.py`

### How?
- Discovered while working on #512

- Ensure PostProcessAdmin no longer uses `_connection.send`
 at any point, but rather uses`QueueClient.send`

JEThompson97:
### Summary of work
- This PR adjusts `post_process_admin.py` and `autoreduction_processor.py` to only use QueueClient methods, and not save and use the underlying stomp connection.
- It contains several notes questions of how to develop this PR further.

### How to test your work
- Review and respond to `Note:` comments
- Review code thus far

Fixes #543

### DIFF
--- a/queue_processors/autoreduction_processor/autoreduction_processor.py
+++ b/queue_processors/autoreduction_processor/autoreduction_processor.py
@@ -152,12 +152,12 @@ class Consumer:
         activemq_client = QueueClient()
         activemq_client.connect()
 
-        # activemq_client.subscribe_amq(consumer_name=self.consumer_name,
-        #                               listener=Listener())
-        activemq_client._connection.subscribe(destination='/queue/ReductionPending',
-                                              id='1',
-                                              ack='auto',
-                                              header={'activemq.prefetchSize': '1'})
+        activemq_client.subscribe_amq(consumer_name=self.consumer_name,
+                                      listener=Listener())
+        # activemq_client._connection.subscribe(destination='/queue/ReductionPending',
+        #                                       id='1',
+        #                                       ack='auto',
+        #                                       header={'activemq.prefetchSize': '1'})
 
 
 def main():  # pragma: no cover
--- a/queue_processors/autoreduction_processor/autoreduction_processor.py
+++ b/queue_processors/autoreduction_processor/autoreduction_processor.py
@@ -136,28 +136,17 @@ class Consumer:
     """ Class used to setup the queue listener. """
     def __init__(self):
         """ Initialise consumer. """
-        # Note: The consumer name of autoreduction_processor.py is "queueProcessor"
-        #   while the consumer name of queue_processor.py is "Autoreduction_QueueProcessor".
-        #   This seems like it needs changing...
-        self.consumer_name = "queueProcessor"
+        self.consumer_name = "autoreduction_processor"
 
     def run(self):
         """
         Connect to ActiveMQ via the QueueClient and listen to the
         /ReductionPending queue for messages.
         """
-        # Note: To reviewer - I'm not very similar with subscriptions, consumers, etc.
-        #   I'm not 100% confident I've changed the code below
-        #   (from connection to client) correctly
         activemq_client = QueueClient()
         activemq_client.connect()
-
         activemq_client.subscribe_amq(consumer_name=self.consumer_name,
                                       listener=Listener())
-        # activemq_client._connection.subscribe(destination='/queue/ReductionPending',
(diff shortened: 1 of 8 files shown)


queue_processors/autoreduction_processor/autoreduction_processor.py
```diff
@@ -136,24 +136,28 @@ class Consumer:
...
-        # Create event listener
-
-        # connection.set_listener('Autoreduction', listener)
-        connection.subscribe(destination='/queue/ReductionPending',
-                             id='1',
-                             ack='auto',
-                             header={'activemq.prefetchSize': '1'})
+        activemq_client.connect()
+
+        activemq_client.subscribe_amq(consumer_name=self.consumer_name,
```
ElliotAOram, line 155: I have had trouble with this function in the past. Could be the issue you're seeing?

JEThompson97: I had a suspicion this was the problem, so I replaced the code with the commented-out directly-using-stomp version below (which was the code before) - see 6db37df - but the Travis issue persisted..

I'm yet to continue my investigation of this issue, been working on other things this morning. I'll let you know after I've done so.

JEThompson97:
@ElliotAOram Ready for review (at long last)

queue_processors/autoreduction_processor/autoreduction_processor.py
```diff
@@ -136,24 +136,28 @@ class Consumer:
     """ Class used to setup the queue listener. """
     def __init__(self):
         """ Initialise consumer. """
+        # Note: The consumer name of autoreduction_processor.py is "queueProcessor"
```
ElliotAOram, line 139: Can you change these to be the correct way round and consistent with package names (e.g. `queue_processor` and `autoreduction_processor`

queue_processors/autoreduction_processor/autoreduction_processor.py
```diff
@@ -136,24 +136,28 @@ class Consumer:
...
-        # connection.set_listener('Autoreduction', listener)
-        connection.subscribe(destination='/queue/ReductionPending',
-                             id='1',
-                             ack='auto',
-                             header={'activemq.prefetchSize': '1'})
+        activemq_client.connect()
+
+        activemq_client.subscribe_amq(consumer_name=self.consumer_name,
+                                      listener=Listener())
+        # activemq_client._connection.subscribe(destination='/queue/ReductionPending',
```
ElliotAOram, line 157: if the above is now working then please feel free to remove this commented code

utils/clients/queue_client.py
```diff
@@ -44,6 +44,9 @@ def connect(self, listener=None):
         if self._connection is None or not self._connection.is_connected():
             self.disconnect()
             return self._create_connection(listener)
+        # pylint:disable=fixme
+        # TODO: To avoid stomp being used directly in future, I think we should no longer
+        #  return _connection; i.e. connection and _create_connection should return nothing
```
ElliotAOram, line 49: I think we are okay to do this (although this is definitely a separate PR)

JEThompson97:
> Can you change these to be the correct way round and consistent with package names (e.g. `queue_processor` and `autoreduction_processor`

I've done this in the latest commit (8645133). But it's worth noting the following, which you may well know already:
- autoreduction_processor.py uses `subscribe_amq` (to ReductionPending only)
- queue_processor.py uses `subscribe_autoreduce` (all queues)

This seems a little backwards to me (although I don't fully under the code) but functionally it works fine, and it least now the names are consistent with the filenames

ElliotAOram:
This is a heritage issue on naming, really it would be better to only have a single subscribe method and then pass in the queues you want to subscribe to, but that is a job for another issue

PR #566, opened 2020-05-19
Add 'get filenames matching regex' method to SFTPClient

### LINKED ISSUE #565: Extend SFTPClient to take regex-filenames

### What?
- The SFTPClient is currently only able retrieve files from exact and complete file paths.
- The client should be able to take a directory path + filename regex and return a list of files which match the regex.

### Where?
`utils/clients/sftp_client.py`

### How?
- In implementing the plot handler, we discovered a need for this functionality extension.

- Ensure the client can return a list of files matching a given regex.

JEThompson97:
### Summary of work
- This PR adds a `get_filenames` method to `sftp_client` which returns a list of filenames within a given SFTP-server directory which match a given regex
- It additionally refactors server path checking into it's own method (since it is now used across `retrieve` and this new method)
- Lastly, it changes parameter names for paths to be more explicit about the type of path expected (file or directory)
- **Note:** regex' must match the [python re lib specification](https://docs.python.org/2/library/re.html)

### How to test your work
- Check that `get_filenames` returns a list of filenames which match the given regex

Fixes #565

### DIFF
--- a/utils/clients/sftp_client.py
+++ b/utils/clients/sftp_client.py
@@ -9,6 +9,8 @@ Client class for retrieving files via SFTP from servers (e.g. CEPH)
 """
 import os.path
 import logging
+import re
+
 import pysftp
 from utils.clients.abstract_client import AbstractClient
 from utils.clients.connection_exception import ConnectionException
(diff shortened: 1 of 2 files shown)


ElliotAOram:
This looks good to me, I think it fulfils all the criteria we discussed 👍

PR #557, opened 2020-05-15
Integrate Message to data_ready (and revert variable names)

JEThompson97:
### Summary of work
- This PR integrates the Message class into `queue_processor.data_ready`
- In order to allow systemtests to pass following this change, the code is updated where necessary to accommodate the Message class usage.
- In addition, this PR reverts variable message names to their original dictionary keys.
  * These will be refactored in a later issue

Part of #512 

**Before merging ensure the release notes have been updated**

PR #567, opened 2020-05-19
Refactor ppa 'do not overwrite' handling

### LINKED ISSUE #564: Refactor "do not overwrite" code in PPA
Issue raised by: [developer]

### What?
- The code within `post_process_admin.reduce` which handles the case where the data contains `{"overwrite":None}` does not work
  * It removes the 'autoreduced' component of the path for some reason.
  * This points to a directory we don't have permissions to,
- This 'no-overwrite handling' should be refactored into a separate method which works and does the following:
  
1. Searches the target directory for any datafile(s) associated with the given run.
_Note: Multiple versions of the same run should always have files appended with the version number. 
i.e. `/autoreduced/<run number>/<version number>`_
2. If any exist, return a new path with with 1+ the highest version number amongst the files
3. The code within reduce will then handle creation of the log directory (by appending `reduction_log/`
___
#### Looking ahead
- After the new ORM has been implemented, the database should be used to determine the new version number instead of existing files
- This is to ensure the version number on the database matches the file name. And to avoid user filename edits resulting in incorrect _next version numbers_

### Where?
`queue_processors/autoreduction_processor/post_process_admin.reduce`

### How?
- This was discovered during work on #557

- Check the 'no-overwrite handling' functionality works as expected

JEThompson97:
### Summary of work
- This PR refactors the code which handles cases where the overwrite flag is False/None in `post_process_admin.reduce`
- The old implementation was broken, but never used until changes to other areas of the code were changed (#557)
- _Note: to enable this code to run, additional changes were required._
  * Due to the Message class having been integrated into manual submission (#542), the data dictionary passed to PPA contained a full set of keys.
  * The previous code only checked whether a `message` key existed. This PR extended this by checking whether the value was `None`.
  * Additionally, since message cannot be null in the DB, this PR converts a None `message` to `=""`
    * This is an open question regarding where in the code this should occur.
___
- Additionally, this PR cherry-picks commits from #557 which revert the data_dict keys (/Message attributes) to their original names

### How to test your work
- Pull the code to the dev server
- Perform a manual submission
- Ensure the run completes without error

### Additional comments
 

Fixes #564

### DIFF
--- a/message/job.py
+++ b/message/job.py
@@ -31,7 +31,7 @@ class Message:
     instrument = attr.ib(default=None)
     rb_number = attr.ib(default=None)
     started_by = attr.ib(default=None)
-    file_path = attr.ib(default=None)
+    data = attr.ib(default=None)
     overwrite = attr.ib(default=None)
     run_version = attr.ib(default=None)
     job_id = attr.ib(default=None)
@@ -39,8 +39,13 @@ class Message:
     reduction_arguments = attr.ib(default=None)
     reduction_log = attr.ib(default=None)
     admin_log = attr.ib(default=None)
-    return_message = attr.ib(default=None)
+    message = attr.ib(default=None)
     retry_in = attr.ib(default=None)
+
+    reduction_data = attr.ib(default=None)
+    run_description = attr.ib(default=None)
+    error = attr.ib(default=None)
+
 
     def serialize(self, indent=None):
         """
--- a/message/job.py
+++ b/message/job.py
@@ -25,8 +25,6 @@ class Message:
     A class that represents an AMQ Message.
     Messages can be serialized and deserialized for sending messages to and from AMQ
     """
-    # Note: Attributes default to None, but when sent to the DB, they cannot be null.
-    #   Perhaps could add check in QueueClient before sending, or change the defaults here.
     description = attr.ib(default=None)
     facility = attr.ib(default=None)
     run_number = attr.ib(default=None)
--- a/message/job.py
+++ b/message/job.py
@@ -25,6 +25,8 @@ class Message:
     A class that represents an AMQ Message.
     Messages can be serialized and deserialized for sending messages to and from AMQ
     """
+    # Note: Attributes default to None, but when sent to the DB, they cannot be null.
+    #   Perhaps could add check in QueueClient before sending, or change the defaults here.
     description = attr.ib(default=None)
     facility = attr.ib(default=None)
     run_number = attr.ib(default=None)
--- a/queue_processors/autoreduction_processor/post_process_admin.py
+++ b/queue_processors/autoreduction_processor/post_process_admin.py
@@ -31,6 +31,7 @@ import importlib.util as imp
 from sentry_sdk import init
 
 # pylint:disable=no-name-in-module,import-error
+from paths.path_manipulation import append_path
 from queue_processors.autoreduction_processor.settings import MISC
 from queue_processors.autoreduction_processor.autoreduction_logging_setup import logger
 from queue_processors.autoreduction_processor.timeout import TimeOut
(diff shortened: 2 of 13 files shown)


queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -394,6 +373,25 @@ def reduce(self):
                         prettify(self.data))
             logger.info("Reduction job successfully complete")
 
+    def _new_reduction_data_path(self, path):
+        logger.info("_new_reduction_data_path argument: %s", path)
+        # if there is an 'overwrite' key/member with a None/False value
+        if 'overwrite' in self.data and not self.data["overwrite"]:
+            if os.path.isdir(path):           # if the given path already exists..
+                contents = os.listdir(path)
+                highest_vers = 0
```
ElliotAOram, line 382: can we set this to `-1`? The minimum version in autoreduction is 0 so in the code below if you can't find a higher number it will increment on 0 so lowest version will be 1. but if we start at -1 then default will be 0.

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -222,30 +223,8 @@ def reduce(self):
...
-                            else:
-                                maximum = 1
-                    if maximum == 0:
-                        new_path = new_path + 'autoreduced'
-                    else:
-                        new_path = new_path + 'autoreduced' + str(maximum) + '/'
-                    final_result_dir = new_path
-                    final_log_dir = new_path + 'reduction_log/'
+            final_result_dir = self._new_reduction_data_path(final_result_dir)
+            final_log_dir = append_path(final_result_dir, ['reduction_log'])
```
ElliotAOram, line 227: good use of append path to avoid OS conflicts :)

queue_processors/autoreduction_processor/post_process_admin.py
JEThompson97: Can we chat about this? Just to confirm I'm on the same page

JEThompson97:
@ElliotAOram I've tested the changes (3 times) on Dev - reduction job successfully completing with expected folder structure
image
I've dotted comments about the code (starting with _"Note:"_) regarding the fix I've made, different options, and other implications

message/job.py
```diff
@@ -25,23 +25,30 @@ class Message:
     A class that represents an AMQ Message.
     Messages can be serialized and deserialized for sending messages to and from AMQ
     """
+    # Note: Attributes default to None, but when sent to the DB, they cannot be null.
+    #   Perhaps could add check in QueueClient before sending, or change the defaults here.
```
ElliotAOram, line 29: Really we want message to be nullable in the database. can you change this in the model? (`WebApp/autoreduce_webapp/reduction_viewer/models.py :: L65`)

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -284,7 +264,9 @@ def reduce(self):
                 raise exp
 
             self.data['reduction_data'] = []
-            if "message" not in self.data:
+            # Note: Open question: where to set Message=""
+            #  (here? QueueClient.send? Message class? allow null in DB?)
+            if "message" not in self.data or self.data["message"] is None:
```
ElliotAOram, line 269: Changing this in the DB as suggested above should remove the need to do this

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -388,12 +372,37 @@ def reduce(self):
 
         else:
             # reduction has successfully completed
+            self.data["message"] = ""   # Note: this assignment is redundant but precautionary
```
ElliotAOram, line 375: Again can be removed with db change

queue_processors/autoreduction_processor/tests/test_post_process_admin.py
```diff
@@ -39,8 +40,8 @@ def test_windows_to_linux_autoreduce_path(self):
         self.assertEqual(actual, '/temp/data/some/more/path.nxs')
 
 
-class TestPostProcessAdmin(unittest.TestCase):
-
+class TestPostProcessAdmin(unittest.TestCase):  # pylint:disable=too-many-public-methods
```
ElliotAOram, line 43: Move this pylint comment to line above

ElliotAOram:
Checked on dev machines and all seems good :shipit:

PR #568, opened 2020-05-19
Create client for django orm use

### LINKED ISSUE #548: Remove SQLAlchemy ORM from Database client
Issue raised by: [developer]

### What?
* Database client currently uses the SQLAlchemy ORM
* Replace this with DjangoORM model
* Replace references to their use where required. 

### Where?
Database_client

### How?
part of #546 

* ensure all unit tests still pass
* Ensure data is correctly retrieved from the database (manual submission on dev should do this)
* Ensure the SQLAlchemy use is removed from the `database_client.py`

ElliotAOram:
### Summary of work
* In order to stagger the implementation, I have created a client for the django ORM
* We should be able to start using this in different areas of the code base (will implement in separate issues) and then remove the old ORM once we have full coverage with the new ORM. 

### How to test your work
* unit tests
* code review
* functional tests will be easier when we start  to implement

### Additional comments
* Will create new issues for integration with different sections of the code base

Fixes #548 

### DIFF
--- a/utils/clients/django_database_client.py
+++ b/utils/clients/django_database_client.py
@@ -1,0 +1,68 @@
+# ############################################################################### #
+# Autoreduction Repository : https://github.com/ISISScientificComputing/autoreduce
+#
+# Copyright &copy; 2020 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+# ############################################################################### #
+"""
+Creates a database session for the reduction database
+"""
+import logging
+
+from model.database import DjangoORM
+from utils.clients.connection_exception import ConnectionException
+from utils.project.static_content import LOG_FORMAT
+from utils.project.structure import get_log_file
+
+logging.basicConfig(filename=get_log_file('django_database_client.log'), level=logging.INFO,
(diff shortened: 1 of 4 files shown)


utils/clients/django_database_client.py
```diff
@@ -0,0 +1,66 @@
...
+                    format=LOG_FORMAT)
+
+
+class DatabaseClient:
+    """
+    Single access point for the mysql database
+    """
+
+    data_model = None
+    variable_model = None
```
JackEAllen, line 27: NIT: Would it be worth moving data_model and  variable_model inside an __init__ method with a private prefix and adding parameter descriptions?
```
class DatabaseClient:
    """
    Single access point for the mysql database
    """
    def __init__(self, data_model, variable_model):
        """
        :param data_model: 
        :param variable_model: 
        """
        self._data_model = data_model
        self._variable_model = variable_model
```

Or just adding comments to describe both variables?

ElliotAOram: I think just adding comments in this case is sufficient. We don't actually want them to be private. Whilst I'll add some common access functions, complex queries will just be using the models directly. Either way I shall ad some comments to them

Issue #480, opened 2020-02-26
IMPLEMENT: Plot Handler


ElliotAOram:
Issue raised by: [project requirement]

### What?
Implement the plotHandler class- which acts as the controller for the plot functionality. This class is responsible for:
* Getting the plotting file (via the SFTP client #470)
* Getting the plot type from the autoreduction database (via the database client)
* Getting the associated plotting meta data file (#474)
* Instructing the Plotting factory to build an IFrame based on the above

This class must be accompanied by a full set of test cases. Given the `How to test` section below, it may be better to develop this with TestDrivenDevelopment. 

This implementation should match the design:

![image](https://user-images.githubusercontent.com/13132128/75336811-b3e04180-5883-11ea-9471-d2d0459f7548.png)

### Where?
This will be a new class in the plotting module

### How to test the issue is resolved
* Try running the following scripts and ensure their outputs:
```
iframe = PlotHandler('MARI', rb_number=1968049).construct plot()
> NotImplementedError "Plotting of RB numbers are not currently supported, please specify a run number"

iframe = PlotHandler('MARI', run_number=27070).construct_plot()
# Should return a string representation of a iframe that will show the data for this run
# Discuss this with whoever implements the plotting factory + plotting meta language

iframe = PlotHandler('MARI', run_number=1).construct_plot()
> RuntimeError "Unable to find run number 1 in the autoreduction database"

iframe = PlotHandler('Fake instrument', run_number=27070).construct_plot()
> RuntimeError "Unable to find instrument 'Fake Instrument` in the autoreduction database"
```

keiranjprice101:
closed in #573

PR #573, opened 2020-05-20
504 plothandler imageonly

franzlang:
### Summary of work

1. added plotting/plot_handler.py which contains the PlotHandler class
The PlotHandler class can now check if a plot file exists already and retrieve it in that case. 
added unittests for PlotHandler class
inserted a call to PlatHandler in reduction_viewer/views.py in the location where it should be called

### How to test your work
run the tests in plotting/tests/test_plot_handler.py


plotting/plot_handler.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+
+    def __init__(self, instrument_name, rb_number, run_number, plot_type=None):
+        self.instrument_name = instrument_name
+        self.rb_number = rb_number
+        self.run_number = run_number
+        if plot_type is None:
+            # list of plot types the handler looks for if none is provided; it searches for the types in the order of
+            # the list and selects the first one that exists. In future: could add vector graphics (.pdf, .eps)
+            self.plot_type = ["png", "jpg", "bmp", "gif", "tiff"]
```
JEThompson97, line 39: Since `self.plot_type` is a list, I'd change it to `self.plot_types` (plural)
Or perhaps even `self.file_extensions` would be more accurate? Unless this can take file names as well as extensions

JEThompson97, line 39: Might it be more accurate to include dots ('.') at the beginning of each string? Just for the sake of a more refined regex

plotting/plot_handler.py
```diff
@@ -0,0 +1,107 @@
...
+    def _check_for_plot_files(self):
+        """
+        Searches the CEPH directory for existing plot files using the directory specified by instrument_name, rb_number,
+        run_number, and a regular expression for the file name based on the instrument_name, run_number and plot_type.
+        """
+        # start sftpclient
+        client = SFTPClient()
+        # initialise list to store names of existing files matching the search
+        _found_files = []
+        for plot_type_i in self.plot_type:
```
JEThompson97, line 75: I would change `plot_type_i` to something like `plot_type` or simply `type`.
Just because this value will be the item within the list, rather than an index.

plotting/plot_handler.py
```diff
@@ -0,0 +1,107 @@
...
+        if len(self._existing_plot_files) == 0:  # no existing file was found
+            return []  # return an empty list for now
+        else:  # if one or more existing files were found, use the first one (order of items in plot_type affects this)
+            _ceph_path = self._rbfolder + self._existing_plot_files[0]
+            # in case the local path to which the file gets copied to needs to be specified
+            _local_path = None
+            # create sftpclient object and try to retrieve the file
+            client = SFTPClient()
+            try:
+                client.retrieve(server_file_path=_ceph_path, local_file_path=_local_path, override=True)
```
JEThompson97, line 99: I'd be inclined to put a `LOGGER.info` something along the lines of
"_<local_file_path> saved._"

plotting/plot_handler.py
```diff
@@ -0,0 +1,107 @@
...
+            except RuntimeError:
+                logging.error("file does not exist")
```
JEThompson97, line 101: Because you've saved your logger as `LOGGER` (line 17), I believe **all** log statements need to begin with
`LOGGER.` rather than `logging.`

plotting/tests/test_plot_handler.py
```diff
@@ -0,0 +1,204 @@
...
+        # check that the sftpclient.get_filenames method was called with the right parameters
+        mock_get_filenames.assert_any_call(server_dir_path="/instrument/WISH/RBNumber/RB87654321/4321/autoreduced/",
+                                           regex="WISH4321*.png")
+        mock_get_filenames.assert_any_call(server_dir_path="/instrument/WISH/RBNumber/RB87654321/4321/autoreduced/",
+                                           regex="WISH4321*.jpg")
+        mock_get_filenames.assert_any_call(server_dir_path="/instrument/WISH/RBNumber/RB87654321/4321/autoreduced/",
+                                           regex="WISH4321*.bmp")
+        mock_get_filenames.assert_any_call(server_dir_path="/instrument/WISH/RBNumber/RB87654321/4321/autoreduced/",
+                                           regex="WISH4321*.gif")
+        mock_get_filenames.assert_any_call(server_dir_path="/instrument/WISH/RBNumber/RB87654321/4321/autoreduced/",
```
JEThompson97, line 118: Rather than having this string copied and pasted each time, I would create a variable within `setUp()` for this string, and use that variable name throughout

JEThompson97: I would do this for all constant values used across the test file.
See the `setUp()` method of `utils/clients/tests/test_sftp_client.py` for an example

plotting/plot_handler.py
```diff
@@ -0,0 +1,107 @@
...
+Getting the associated plotting meta data file
+Instructing the Plotting factory to build an IFrame based on the above
+"""
+import logging
+from utils.clients.sftp_client import SFTPClient
+
+LOGGER = logging.getLogger('app')
+
+
+# pylint:disable=line-too-long, no-else-return
```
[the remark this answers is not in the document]
ElliotAOram: Either is fine for line too long

plotting/plot_handler.py
```diff
@@ -0,0 +1,107 @@
...
+        self._existing_plot_files = []
+
+    def _regexp_for_file_name(self, plot_type):
+        """
+        Regular expression used for looking for plot files. It assumes that the file names follow the convention
+        <instrument_abbreviation><run_number>*<.png or other extension>
+        """
+        try:
+            _regexp = self._instrument_dict[self.instrument_name] + str(
+                self.run_number) + "*." + plot_type
```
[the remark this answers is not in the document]
ElliotAOram: Agreed

plotting/plot_handler.py
```diff
@@ -0,0 +1,107 @@
...
+            return _regexp
+        except KeyError:  # if the instrument name does not appear in the dictionary of known instruments
+            LOGGER.info("The instrument name is not recognised")
```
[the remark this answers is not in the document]
ElliotAOram: Yes just return here - we don't want to raise as it will be live and user exposed on the webapp

plotting/plot_handler.py
ElliotAOram: use plot_type rather than type. `type` is a reserved key word in python

plotting/plot_handler.py
```diff
@@ -0,0 +1,107 @@
...
+        Searches the CEPH directory for existing plot files using the directory specified by instrument_name, rb_number,
+        run_number, and a regular expression for the file name based on the instrument_name, run_number and plot_type.
+        """
+        # start sftpclient
+        client = SFTPClient()
+        # initialise list to store names of existing files matching the search
+        _found_files = []
+        for plot_type_i in self.plot_type:
+            # directory where existing plot file(s) would be located
+            ceph_dir = self._rbfolder
```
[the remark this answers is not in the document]
ElliotAOram: I would argue to just use self._rbfolder directly in the loop. can we rename to `self._rb_folder`

plotting/plot_handler.py
ElliotAOram: both actually work but we should use `LOGGER` to be consistent

ElliotAOram:
Agree with @JEThompson97 comments. Will have another look over them after these are fixed. Looks good on the whole though :)

plotting/plot_handler.py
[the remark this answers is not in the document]
franzlang: I kept all the lines within 120 characters according to PEP8. I found the 100 lines from pylint a bit too restrictive in terms of readability in a few places. Happy to refactor lines to conform to 100 max characters.

plotting/plot_handler.py
franzlang: Will make the suggested change. The naming was initially chosen because of issue 480 (https://github.com/ISISScientificComputing/autoreduce/issues/480), but we later realised that we don't really need a get_plot_type method.

franzlang: The regex already adds the dot. I chose this usage as it felt more naturally to me that someone might use the PlotHandler class thinking "I want to look for a png file" rather than "I want to look for a .png file". Either way is fine with me; the change would be easy (_regexp_for_file_name would have to be changed too)

plotting/plot_handler.py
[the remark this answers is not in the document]
franzlang: The pylint 'no-else-return' is actually about the construction:
`if A:
    return A
else:
   do something else`
pylint wants you to write
`if A:
    return A
do something else`
which I personally find less readable.
I have added a return statement to the exception so that a construction of the form 
`if PlotHandler.get_plot_file:`
can be used

plotting/plot_handler.py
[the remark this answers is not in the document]
franzlang: it should have been before the loop, but as Elliot said it did not need to be made that explicit. I have renamed self._rbfolder to self._rb_folder and use that within the loop now.

plotting/plot_handler.py
franzlang: added a 
`LOGGER.info`
line

plotting/plot_handler.py
franzlang: good catch. changed it.

plotting/plot_handler.py
JEThompson97: No problem, go ahead with the way that seems best to you 👍

plotting/plot_handler.py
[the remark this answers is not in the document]
JEThompson97: The code is definitely readable in my opinion so happy for you to stick with the pylint:disables

plotting/plot_handler.py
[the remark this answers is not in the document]
ElliotAOram: Yep this is fine

Issue #546, opened 2020-05-12
Integrate DjangoORM into Autoreduction


ElliotAOram:
Issue raised by: [developer/user/project requirement]

### What?
* Identify sections that should use Django ORM
* Create sub issues for these:
* Replace sections of code with Django ORM 

### Where?
* To be identified

### How to test the issue is resolved
* Each section will be tested individually. The issue will be closed by removing `queue_processor.orm` `queue_processor.base` and the ORM in `queue_client`

PR #575, opened 2020-05-21
Django ORM manual submission

ElliotAOram:
### Summary of work
* Remove old way of accessing database in manual_submission
* Add function for retrieving reduction_runs in django db client

### How to test your work
* Attempt to manually submit a run already in the database using the manual_submission script.
* Ensure that it reaches the webapp / database successfully  

Part of #546

### DIFF
--- a/scripts/manual_operations/manual_submission.py
+++ b/scripts/manual_operations/manual_submission.py
@@ -20,7 +20,7 @@ from message.job import Message
 from utils.clients.connection_exception import ConnectionException
 from utils.clients.icat_client import ICATClient
 from utils.clients.queue_client import QueueClient
-from utils.clients.database_client import DatabaseClient
+from utils.clients.django_database_client import DatabaseClient
 
 
 def submit_run(active_mq_client, rb_number, instrument, data_file_location, run_number):
@@ -45,10 +45,11 @@ def submit_run(active_mq_client, rb_number, instrument, data_file_location, run_
     print("Submitted run: \r\n" + message.serialize(indent=1))
 
 
-def get_location_and_rb_from_database(database_client, run_number):
+def get_location_and_rb_from_database(database_client, instrument, run_number):
     """
     Retrieves a run's data-file location and rb_number from the auto-reduction database
     :param database_client: Client to access auto-reduction database
+    :param instrument: (str) the name of the instrument associated with the run
     :param run_number: The run number of the data to be retrieved
     :return: The data file location and rb_number, or None if this information is not
     in the database
@@ -57,29 +58,16 @@ def get_location_and_rb_from_database(database_client, run_number):
         print("Database not connected")
         return None
 
-    db_connection = database_client.connect()
-    location_query = f"""
-                    SELECT file_path
-                    FROM reduction_viewer_reductionlocation
-                    WHERE reduction_run_id = {run_number}
-                    """
-    location_result = db_connection.execute(location_query).fetchall()
-    if len(location_result) == 0:
+    reduction_run_records = database_client.get_reduction_run(instrument, run_number)
+
+    if not reduction_run_records:
         return None
 
-    rb_query = f"""
-                SELECT reduction_viewer_experiment.reference_number
-                FROM reduction_viewer_reductionrun
-                    JOIN reduction_viewer_experiment
-                    ON reduction_viewer_reductionrun.experiment_id = reduction_viewer_experiment.id
-                WHERE reduction_viewer_reductionrun.id = {run_number}
-                """
+    reduction_run_record = reduction_run_records.order_by('run_version').first()
+    data_location = reduction_run_record.data_location.first().file_path
+    experiment_number = reduction_run_record.experiment.reference_number
 
-    rb_result = db_connection.execute(rb_query).fetchall()
-    if len(rb_result) == 0:
-        return None
-
-    return location_result[0].file_path, rb_result[0].reference_number
+    return data_location, experiment_number
 
 
 def get_location_and_rb_from_icat(icat_client, instrument, run_number, file_ext):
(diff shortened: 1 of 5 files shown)


scripts/manual_operations/manual_submission.py
```diff
@@ -57,29 +54,16 @@ def get_location_and_rb_from_database(database_client, run_number):
...
 
-    rb_result = db_connection.execute(rb_query).fetchall()
-    if len(rb_result) == 0:
+    if not reduction_run_records:
         return None
 
-    return location_result[0].file_path, rb_result[0].reference_number
+    reduction_run_record = reduction_run_records.order_by('run_version').first()
+    data_location = reduction_run_record.data_location.first().file_path
+    experiment_number = reduction_run_record.experiment.reference_number
```
JEThompson97, line 64: Your first usage of `reduction_run_record` (ln 63) calls `first()`
Your second usage (ln 64) doesn't. Is this correct?

JEThompson97: Also, it is a bit difficult at first glance to distinguish `reduction_run_records` and `reduction_run_record`.
Do you think it's worth naming one a bit differently? e.g. just `run_record` for the singular?

scripts/manual_operations/manual_submission.py
```diff
@@ -143,8 +127,56 @@ def get_location_and_rb(database_client, icat_client, instrument, run_number, fi
     return get_location_and_rb_from_icat(icat_client, instrument, run_number, file_ext)
 
 
-def main():
-    """ File usage description, validation and running mechanism """
+def login_icat():
```
JEThompson97, line 130: Nice refactoring. This was getting a bit cumbersome in `main`

scripts/manual_operations/manual_submission.py
```diff
@@ -165,36 +197,29 @@ def main():
...
+        raise RuntimeError("Unable to proceed. Unable to connect to ICAT or Database. "
+                           "At least one connection is required to perform manual submission.")
+
+    activemq_client = login_queue()
 
     for run in run_numbers:
         location, rb_num = get_location_and_rb(database_client, icat_client, instrument, run, "nxs")
-        submit_run(activemq_client, rb_num, instrument, location, run)
+        if location and rb_num:
+            submit_run(activemq_client, rb_num, instrument, location, run)
```
JEThompson97, line 221: Perhaps an `else: print()` letting the user know run with location _x_ and rb_num _y_ wasn't found?

scripts/manual_operations/tests/test_manual_submission.py
```diff
@@ -156,3 +148,130 @@ def test_submit_run(self):
         self.sub_run_args[0].send.assert_called_with('/queue/DataReady',
                                                      message,
                                                      priority=1)
+
+    @patch('icat.Client')
```
JEThompson97, line 152: Shouldn't these be `icat.Client.__init__`? (i.e. mocking the init method rather than class)

scripts/manual_operations/tests/test_manual_submission.py
```diff
@@ -156,3 +148,130 @@ def test_submit_run(self):
...
+        Note: We mock the connect so it does not actual perform the connect (default pass)
+        """
+        actual = ms.login_database()
+        self.assertIsInstance(actual, DatabaseClient)
+
+    @patch('utils.clients.DatabaseClient.connect')
+    def test_database_login_invalid(self, mock_connect):
+        """
+        Test: None is returned
+        When: We are unable to log in via the client
```
JEThompson97, line 191: "via the **database** client"?

scripts/manual_operations/tests/test_manual_submission.py
```diff
@@ -156,3 +148,130 @@ def test_submit_run(self):
...
+        actual = ms.login_icat()
+        self.assertIsInstance(actual, ICATClient)
+        mock_connect.assert_called_once()
+
+    @patch('icat.Client')
+    @patch('utils.clients.ICATClient.connect')
+    def test_icat_login_invalid(self, mock_connect, _):
+        """
+        Test: None is returned
+        When: We are unable to log in via the client
```
JEThompson97, line 170: "via the **icat** client"?

scripts/manual_operations/tests/test_manual_submission.py
```diff
@@ -156,3 +148,130 @@ def test_submit_run(self):
...
+        Note: We mock the connect so it does not actual perform the connect (default pass)
+        """
+        actual = ms.login_queue()
+        self.assertIsInstance(actual, QueueClient)
+
+    @patch('utils.clients.QueueClient.connect')
+    def test_queue_login_invalid(self, mock_connect):
+        """
+        Test: An exception is raised
+        When: We are unable to log in via the client
```
JEThompson97, line 212: "via the **AMQ** client"?

scripts/manual_operations/tests/test_manual_submission.py
```diff
@@ -156,3 +148,130 @@ def test_submit_run(self):
...
+        """
+        con_exp = ConnectionException('Queue')
+        mock_connect.side_effect = con_exp
+        self.assertRaises(RuntimeError, ms.login_queue)
+
+    def test_submit_run_no_amq(self):
+        """
+        Test: That there is an early return
+        When: calling submit_run with active_mq as None
```
JEThompson97, line 221: "When: **Calling** submit_run"

utils/clients/tests/test_django_database_client.py
```diff
@@ -95,3 +95,26 @@ def test_get_instrument_invalid(self):
...
+    def test_get_reduction_run_valid(self):
+        """
+        Test: A ReductionRun record is returned
+        When: get_reduction_run is called with values that match a database record
+        """
+        mock_data_model = Mock()
+        client = DatabaseClient()
+        client.connect()
+        client.data_model = mock_data_model
+        client.get_reduction_run('GEM', 1)
```
JEThompson97, line 109: Should we test that the return value of this is something?

scripts/manual_operations/tests/test_manual_submission.py
ElliotAOram: Doesn't matter in this case as we only care about the raised exception from connect

utils/clients/tests/test_django_database_client.py
ElliotAOram: It's a mocked value, so it does not actually return anything. We don't want to use the real database as it will fail locally without the test database

PR #580, opened 2020-05-26
Integrate Message into queue_processor

### LINKED ISSUE #512: Integrate AMQ Message into autoreduction

JEThompson97:
### Summary of work
- This PR integrates the Message class into the entire `queue_processor`

### How to test your work
- Ensure tests pass
- Test on dev

### Additional comments
- This is part of a series of PRs to integrate the Message class throughout the database.
- Previous, un-merged PRs attempted this is various areas of the code base, but were unsuccessful due to changes that had to be made to other areas of the code (see #567)

Fixes #512

### DIFF
--- a/queue_processors/queue_processor/queue_processor.py
+++ b/queue_processors/queue_processor/queue_processor.py
@@ -219,7 +219,7 @@ class Listener:
                               priority=self._priority)
             logger.info("Run %s ready for reduction", self.message.run_number)
 
-    # note: question: Why does this take arguments and not just take from _data_dict/message
+    # note: Why does this take arguments and not just take from the message attribs
     def _construct_and_send_skipped(self, rb_number, reason):
         """
         Construct a message and send to the skipped reduction queue
@@ -433,12 +433,9 @@ class Listener:
                     experiment.id,
                     int(self.message.run_number),
                     int(self.message.run_version))
-        reduction_run = session.query(ReductionRun).filter_by(experiment=experiment,
-                                                              run_number=int(
-                                                                  self.message.run_number),
-                                                              run_version=int(
-                                                                  self.message.run_version))\
-            .first()
+        reduction_run = session.query(ReductionRun)\
+            .filter_by(experiment=experiment, run_number=int(self.message.run_number),
+                       run_version=int(self.message.run_version)).first()
         return reduction_run
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


queue_processors/queue_processor/queue_processor.py
```diff
@@ -72,7 +73,7 @@ def on_message(self, headers, message):
         logger.info("Destination: %s Priority: %s", destination, self._priority)
         # Load the JSON message and header into dictionaries
         try:
-            self._data_dict = json.loads(message)
+            self.message.populate(json.loads(message))
```
ElliotAOram, line 76: do we need the json load here? I think ideally we want to remove the JSON dependency from the queue_processors and encapsulate this into the message itself (e.g. only the message needs to know how it can be packed up and sent between queues)

JEThompson97: I'm finding that doing this causes the `test_end_to_end` tests to fail (e8afd5d).
I tried quickly integrating Message into them (86279bf), but this still caused errors. I think that will require a more thoughtful approach.

It seems removing this json tolerance in queue_processor requires json to be removed from other areas of the code base. I wonder if the best approach would be to keep this for now by reverting to (2c92250) and adding a note to return to it?

queue_processors/queue_processor/queue_processor.py
```diff
@@ -71,12 +72,10 @@ def on_message(self, headers, message):
...
-        try:
-            self._data_dict = json.loads(message)
-        except ValueError:
-            logger.error("Could not decode message from %s", destination)
-            logger.error(sys.exc_info()[1])
-            return
+        if not isinstance(message, Message):
+            raise ValueError("message received is not of type Message")
+        self.message = message
+
```
ElliotAOram, line 78: Here we still want to initalise the message using the Message class, but just pass in the JSON object and allow it not use the json.load message here. The above won't work as AMQ can't pass Message class objects, it has to be JSON

queue_processors/queue_processor/queue_processor.py
```diff
@@ -72,7 +72,10 @@ def on_message(self, headers, message):
         logger.info("Destination: %s Priority: %s", destination, self._priority)
         # Load the JSON message and header into dictionaries
         try:
-            self._data_dict = json.loads(message)
+            if isinstance(message, Message):
```
ElliotAOram, line 75: This is fine to have, although I'm not sure we would ever hit this statement. Still might save us a run failure at some point

ElliotAOram:
Looks good to me, just going to run it through some manual tests and check it is all okay

PR #581, opened 2020-05-26
Integrate Message into PPA

JEThompson97:
### Summary of work
- This PR integrates the Message class into  `post_process_admin`
- As part of the changes, the `prettify` method (which reduces the length of the reduction_script in a json dump) was replaced by a new argument of `Message.serialize`: `limit_reduction_script`.

### How to test your work
- Ensure tests pass
- Test on dev

### Additional comments
- This is part of a series of PRs to integrate the Message class throughout the database.
- Previous, un-merged PRs attempted this is various areas of the code base, but were unsuccessful due to changes that had to be made to other areas of the code (see #567)
- During this PR, I began to refactor the PPA attributes, but then reverted the change as I found it to be a larger task than anticipated

Fixes #512

**Before merging ensure the release notes have been updated**

PR #584, opened 2020-05-27
Integrate Message into autoreduction_processor and messaging_utils

### LINKED ISSUE #512: Integrate AMQ Message into autoreduction

JEThompson97:
### Summary of work
- This PR integrates the Message class into `autoreduction_processor` and `messaging_utils`.
- In doing so, a 'cancel' attribute has been added to the Message class.

### How to test your work
- Ensure tests pass
- Test on dev

### Additional comments
- This is part of a series of PRs to integrate the Message class throughout the database.
- **I rebased following #581 being merged. Many of the 'files changed' are actually changes from that branch**

Fixes #512

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,10 +37,6 @@ script:
   - echo "Running System Tests"
   - py.test systemtests --cov-append
 
-  - cat logs/queue_processor.log
-  - cat logs/queue_client.log
-  - cat logs/autoreduction_processor.log
-
   # ================ Static Analysis ================= #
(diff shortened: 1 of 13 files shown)


JEThompson97:
Travis has stopped attaching reports, but [can be found here](https://travis-ci.org/github/ISISScientificComputing/autoreduce/pull_requests) (look for the latest #584)

JEThompson97:
Tested on dev:
image

PR #550, opened 2020-05-12
Develop Tour implementation

JEThompson97:
### Summary of work
This PR develops our initial implementation of Bootstrap Tour based on comments from the team. Specially, it adds the following:
- Added divs on either side of the "ISIS Auto-reduction" title to place Tour button within **[1]**
  * This affords a consistent location to have the Tour button on across the WebApp
- Refactored code to dynamically insert Tour button if, and only if, `create_tour.js` script loaded
- Added padding around tour elements **[2]**
- Naming convention of files aligned with the rest of the code base (using underscores '_' rather than hyphens '-')

image

### How to test your work
- Confirm whether the changes are appropriate (especially those to code files e.g. `header.html`)

Part of #531

**Before merging ensure the release notes have been updated**

PR #590, opened 2020-05-28
Move block scripts sections to it's previous state

### LINKED ISSUE #589: jQuery and other scripts not loading on WebApp
Issue raised by: [developer]

### What?
- PR #550 introduced a bug which stopped scripts from being run properly.
- The bug lies in the change in the screenshot below (although I'm not sure why this causes the bug)
image

### Where?
- WebApp/autoreduce_webapp/templates/base.html

### How?
- Found by tested the webapp and noticing jQuery wasn't working for the 'Show reduction logs' link

### Reproducible?
Yes

- Scripts load on pages which extend `base.html` (presumably all pages?)

JEThompson97:
### Summary of work
- Reverts a change made in #550 which caused a bug
- For an unknown reason, this caused jQuery (and potentially other scripts) to not load on run_summary.js (and presumably other pages since the change was in `base.html`).

### How to test your work
- Ensure scripts load throughout webapp as expected.

Fixes #589 

### DIFF
--- a/WebApp/autoreduce_webapp/templates/base.html
+++ b/WebApp/autoreduce_webapp/templates/base.html
@@ -28,13 +28,12 @@
                 {% block footer %}{% endblock %}
                 {% include "footer.html" %}
             </div>
-            {% block scripts %}
             <script src="{% static "javascript/vendor/jquery.min.js" %}"></script>
             <script src="{% static "javascript/vendor/bootstrap.min.js" %}"></script>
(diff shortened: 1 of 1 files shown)


ElliotAOram:
I think we can remove the `block scripts` syntax entirely. I believe it is only required for overriding in inherited templates see: https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#block

Issue #513, opened 2020-04-20
AMQ message validation check


ElliotAOram:
Issue raised by: [developer]

### What?
After the AMQ Message has been integrated into the code base, we should look to validate the Message at various stages of the pipeline. Some of the rules for validation have been noted in the image below.

we should take advantage of the valid functionality offered by `attr`: When we add validation to this we can also use attrs to help us: http://www.attrs.org/en/stable/examples.html#validators

![image](https://user-images.githubusercontent.com/13132128/79734805-38cd6180-82ef-11ea-92f5-8e66d3b8f285.png)

### Where?
`message.job.Message`

### How to test the issue is resolved
* New test cases that ensure both valid and invalid input are handled in the expected way. 
* Ensure that Autoreduction still behaves as expected (system tests/unit test)
* Add this change to the development environment and check it continues to work.

PR #594, opened 2020-06-02
Add validation for data ready

ElliotAOram:
### Summary of work
* Add implementation and tests for the data_ready validation 

### How to test your work
* Ensure unit tests pass
* Code review
* Needs 100% coverage of new lines

### Additional comments
* The next PR will be integrating the validation to pre-data_ready 

Part of  #513

### DIFF
--- a/message/validation/process.py
+++ b/message/validation/process.py
@@ -1,0 +1,41 @@
+# ############################################################################### #
+# Autoreduction Repository : https://github.com/ISISScientificComputing/autoreduce
+#
+# Copyright &copy; 2020 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+# ############################################################################### #
+"""
+Functions that handle the processing of the validation dictionary
+These can be considered helper functions to message.validation.stages
+"""
+import logging
+
+
+def check_validity_dict(validity_dict):
+    """
+    Loop through a dictionary and assert if all are true
+    If not log / print those that are false
+    :param validity_dict: (dict) of key (name of validity check), values (result of check)
+    :return: True if all values in dict are true, else false
+    """
+    is_valid = True
+    for status in validity_dict.values():
+        if not status:
+            is_valid = False
+
+    if is_valid:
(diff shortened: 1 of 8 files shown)


message/validation/process.py
```diff
@@ -0,0 +1,41 @@
...
+    return False
+
+
+def dict_to_string(validity_dict):
+    """
+    create and return a string representation of the validity check
+    :param validity_dict: (dict) containing the validity checks
+    :return: (str) A string to print
+    """
+    result_str = ""
```
JackEAllen, line 38: NIT: would it be cleaner to place this inside the method input?
`def dict_to_string(validity_dict, result_str=str()):`

message/validation/stages.py
```diff
@@ -0,0 +1,28 @@
...
+    :param message: A message object to be validated
+    :return: True if valid
+    """
+    validity_dict = {
+        'run_number_valid': validate_run_number(message.run_number),
+        'instrument_valid': validate_instrument(message.instrument),
+        'rb_number_valid': isinstance(message.rb_number, int),
+        'started_by_valid': isinstance(message.started_by, int),
+        'file_path_valid': isinstance(message.data, str),
+        'facility': isinstance(message.facility, str)
```
JackEAllen, line 26: NIT: For naming_convention consistency in the dictionary, could we change the 'facility' key to be 'facility_valid'?

message/validation/validators.py
```diff
@@ -0,0 +1,26 @@
...
+"""
+Validators to be used in the Message class
+"""
+from utils.settings import VALID_INSTRUMENTS
+
+
+def validate_run_number(run_number):
+    """
+    Assert a run number is valid
+    :param run_number: The run numebr to validate
```
JackEAllen, line 16: Minor typo: numebr -> number

message/validation/process.py
ElliotAOram: Not sure there would ever be a case for passing this in as param. Think it might just be more confusing as this is not done explicitly anywhere else in autoreduction.

PR #605, opened 2020-06-16
Add and integrate Software table

### LINKED ISSUE #528: Add and integrate software table into autoreduction

ElliotAOram:
### Summary of work
* Added the Software table to the model
* Added access function for get/create
* Added version find function for mantid
* Propagate version through autoreduction processor to be saved in queue processor

### How to test your work
* Push this code to the dev nodes (will need to perform a database migration) to get the new table
* perform manual submission and ensure that the software records are created and associated with each run

Fixes #528 


ElliotAOram:
Need to test this manually before it's ready for review

model/database/tests/test_access.py
```diff
@@ -79,6 +79,40 @@ def test_get_experiment(self):
...
 
+    @patch('model.database.access.save_record')
+    def test_get_experiment_create(self, mock_save):
+        """
+        Test: An Experiment record is created
+        When: get_experiment is called with create option True
+        """
+        database = access.start_database()
+        actual = access.get_experiment(rb_number=9999999, create=True)
+        self.assertIsNotNone(actual)
```
JackEAllen, line 90: Is this test necessary to check if not None? 
Would this not be highlighted in the assert equality checks.

WebApp/autoreduce_webapp/reduction_viewer/models.py
```diff
@@ -46,6 +46,17 @@ def __unicode__(self):
         return u'%s' % self.value
 
 
+class Software(models.Model):
+    """
+    Represents the software used to perform the reduction
+    """
+    name = models.CharField(max_length=100, blank=False)
+    version = models.CharField(max_length=20, blank=False)
```
JackEAllen, line 54: Based on my comment in `model/database/tests/test_access.py` about not None validation, we could alternatively just add not null validation to both columns in the model itself unless there is a need to have just the name or just the version column populated for a given row.

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -359,6 +360,20 @@ def reduce(self):
                         self.message.serialize(limit_reduction_script=True))
             logger.info("Reduction job successfully complete")
 
+    @staticmethod
+    def _get_mantid_version():
```
JackEAllen, line 364: Should we add the return to the docstring for explicite?

ElliotAOram:
@JackEAllen coverall fails sue to lines in untested files so we can ignore this time

ElliotAOram:
@JackEAllen If you restart the queue you should be able to submit again. That bug won't be fixed until the QP is re-written

JackEAllen:
This is now good to merge!  🎉

PR #629, opened 2020-06-26
Handle multiple images for a single job

### LINKED ISSUE #617: Handle displaying multiple **images** from drop down in webapp

### What?
What is the issue

### Where?

### How?
How did the issue come about

### Reproducible?
[Yes/No]

ElliotAOram:
### Summary of work
* Allow the webapp to show multiple images for a single reduction job 

### How to test your work
* Re-run MAR27036
* After it completes find the run in the dev webapp (afetr switching to this branch)
* Observe multiple plots in the webapp

### Additional comments
* Note: this does not handle the drop down menu for plot selection but instead shows all. We will handle the display of this in a separate issue

Fixes #617

### DIFF
--- a/plotting/plot_handler.py
+++ b/plotting/plot_handler.py
@@ -84,20 +84,23 @@ class PlotHandler:
         :return: (str) local path to downloaded files OR None if no files found
         """
         _existing_plot_files = self._check_for_plot_files()
+        local_plot_paths = []
         if _existing_plot_files:
-            # Generate paths to data on server and destination on local machine
-            _server_path = self.server_dir + _existing_plot_files[0]
-            _local_path = os.path.join(self.static_graph_dir, _existing_plot_files[0])
-
             client = SFTPClient()
-            try:
-                client.retrieve(server_file_path=_server_path,
-                                local_file_path=_local_path,
-                                override=True)
-                LOGGER.info('File %s found and saved to %s', _server_path, _local_path)
-            except RuntimeError:
-                LOGGER.error("file does not exist")
-                return None
-            return f'/static/graphs/{_existing_plot_files[0]}'  # shortcut to static dir
+            for plot_file in _existing_plot_files:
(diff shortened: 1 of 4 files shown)


plotting/plot_handler.py
```diff
@@ -84,20 +84,23 @@ def get_plot_file(self):
...
+                _server_path = self.server_dir + plot_file
+                _local_path = os.path.join(self.static_graph_dir, plot_file)
+
+                try:
+                    client.retrieve(server_file_path=_server_path,
+                                    local_file_path=_local_path,
+                                    override=True)
+                    LOGGER.info('File %s found and saved to %s', _server_path, _local_path)
+                except RuntimeError:
+                    LOGGER.error("file does not exist")
```
JackEAllen, line 101: Should we add the file and location that can't be found in this log for more context?

plotting/plot_handler.py
```diff
@@ -84,20 +84,23 @@ def get_plot_file(self):
...
-                                override=True)
-                LOGGER.info('File %s found and saved to %s', _server_path, _local_path)
-            except RuntimeError:
-                LOGGER.error("file does not exist")
-                return None
-            return f'/static/graphs/{_existing_plot_files[0]}'  # shortcut to static dir
+            for plot_file in _existing_plot_files:
+                # Generate paths to data on server and destination on local machine
+                _server_path = self.server_dir + plot_file
+                _local_path = os.path.join(self.static_graph_dir, plot_file)
```
JackEAllen, line 93: NITPICK would _host_path be better name than _local_path? (only personal preference so please ignore if you don't agree)

ElliotAOram: I think `host` is more ambiguous as it doesn't explicitly refer to which machine we are referring to. Local is immediately obvious that it is the local machine the code is running on.

JackEAllen:
Just performed manual testing and all looks good!

PR #645, opened 2020-06-30
Refactor queue processors

DavidFair:
### Summary of work
- Splits Queue Processor into two separate classes `StompClient` and `MessageHandler`
- Refactor to use exceptions rather than catching the base exception (which would include say OOM exceptions)
- Fix a bug where we use run number without RB number to select a reduction record
- Consolidate some duplicated code
- More tests now things throw so are observable
- (TODO) Rename packages based on feedback

### How to test your work
- Code Review
- If possible, run a queue processor in a test environment and check everything still works for the moment

### Additional comments

Next up:
- adding a way to disconnect a Stomp Client gracefully
- Timeout period for shutting down gracefully so that cron can startup a new instance
- File Lock to ensure we only have one instance

**Release notes will be updated when work finishes**


queue_processors/queue_processor/stomp_client.py
```diff
@@ -0,0 +1,98 @@
...
+import traceback
+from model.message.job import Message
+from queue_processors.queue_processor.handle_message import HandleMessage
+from queue_processors.queue_processor.handling_exceptions import \
+    InvalidStateException
+from queue_processors.queue_processor.settings import LOGGING
+from utils.clients.queue_client import QueueClient
+
+
+class StompClient:
```
ElliotAOram, line 24: Not sure about this name, only because to me this suggests that it is one of the utils clients like database_client or queue_client. To me this class represents the routing of messages and the middle layer between queue_client and message processing. Open to counter arguments here, this is just the way I interpreted it

DavidFair: So `queue_client`? 
I originally picked this name as the class layout (e.g. `on_message`) is specific to Stomp, yet there isn't really much to indicate that this abides to an API

DavidFair: I've gone for `QueueListener` since `QueueClient` is already used by something else

Feel free to let me know if you have any other ideas for what to call it

ElliotAOram: `QueueListener` sounds good

queue_processors/queue_processor/handle_message.py
```diff
@@ -0,0 +1,388 @@
...
+from model.database import access as db_access
+from model.message.job import Message
+from queue_processors.queue_processor._utils_classes import _UtilsClasses
+from queue_processors.queue_processor.handling_exceptions import \
+    MissingReductionRunRecord, InvalidStateException, MissingExperimentRecord
+from queue_processors.queue_processor.settings import LOGGING
+from utils.settings import ACTIVEMQ_SETTINGS
+
+
+def validate_rb_num(rb_number):
```
ElliotAOram, line 28: FYI: This can be refactored out into the Message class - but we'll do that in a different PR

queue_processors/queue_processor/handle_message.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+class HandleMessage:
+    """
+    Handles messages from the queue client and forwards through various
+    stages depending on the message contents.
+    """
+    # We cannot type hint stomp client without introducing a circular dep.
+    def __init__(self, stomp_client):
+        self._client = stomp_client
```
ElliotAOram, line 52: update this to queue_listener

queue_processors/queue_processor/handle_message.py
```diff
@@ -0,0 +1,388 @@
...
+        if instrument.is_paused:
+            self._logger.info("Run %s has been skipped",
+                              message.run_number)
+        else:
+            self._client.send_message('/queue/ReductionPending', message)
+            self._logger.info("Run %s ready for reduction",
+                              message.run_number)
+
+    # pylint: disable=too-many-arguments
+    def _create_reduction_run_record(self, experiment, instrument, message,
```
ElliotAOram, line 153: Maybe this could be refactored into the `model.orm.access`?

queue_processors/queue_processor/handle_message.py
```diff
@@ -0,0 +1,388 @@
...
+            created=datetime.datetime.utcnow(),
+            last_updated=datetime.datetime.utcnow(),
+            experiment_id=experiment.id,
+            instrument_id=instrument.id,
+            status_id=status.id,
+            script=script_text,
+            started_by=message.started_by)
+        return reduction_run
+
+    def _get_last_run_version(self, run_no, experiment):
```
ElliotAOram, line 172: Refactor this to model.database.access (it should replace `find_highest_run_version` as this implementation is better)

queue_processors/queue_processor/handle_message.py
```diff
@@ -0,0 +1,388 @@
...
+        last_run = self._data_model.ReductionRun.objects \
+            .filter(run_number=run_no) \
+            .filter(experiment=experiment) \
+            .order_by('-run_version') \
+            .first()
+
+        # By returning -1 callers can blindly increment the version
+        return last_run.run_version if last_run else -1
+
+    def _get_db_inst(self, instrument_name):
```
ElliotAOram, line 187: This name is a little misleading as the function also activates the instrument if required. Perhaps its a tad elongated but can we go for `_get_and_activate_instrument(...)` or similar

ElliotAOram: Also needs a docstring

queue_processors/queue_processor/handle_message.py
```diff
@@ -0,0 +1,388 @@
...
+        self._utils = _UtilsClasses()
+
+        logging.config.dictConfig(LOGGING)
+        self._logger = logging.getLogger("handle_queue_message")
+
+        self._cached_db = None
+        self._cached_data_model = None
+
+    @property
+    def _database(self):
```
ElliotAOram, line 62: docstring

queue_processors/queue_processor/handle_message.py
```diff
@@ -0,0 +1,388 @@
...
+        if not self._cached_db:
+            self._cached_db = db_access.start_database()
+        return self._cached_db
+
+    @property
+    def _data_model(self):
```
ElliotAOram, line 68: docstring

queue_processors/queue_processor/handle_message.py
ElliotAOram: docstring

queue_processors/queue_processor/handling_exceptions.py
```diff
@@ -0,0 +1,46 @@
+# ############################################################################
+# Autoreduction Repository :
+# https://github.com/ISISScientificComputing/autoreduce
+#
+# Copyright &copy; 2020 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+# ############################################################################
+
+"""
+This module contains various exceptions that the message handling could
```
ElliotAOram, line 10: 👍

queue_processors/queue_processor/handle_message.py
DavidFair: Opened #652

queue_processors/queue_processor/handle_message.py
DavidFair: Definitely, I moved it into its own method as I think it shouldn't be here, but wanted to avoid making it difficult to track stuff moving around. I've opened #653 for these

I'm surprised Pylint didn't pick these up, I can't see a `disable=docstring` :thinking:

PR #646, opened 2020-07-01
Place log and updating of amq message with reduction started inside s…

### LINKED ISSUE #621: Place Reduction Started & Logs in Sub-method
### What?
Move logging message and actions to send reduction started to AMQ into a sub-method inside of reduce()
The method could be called:  `reduction_started`

### Where?
* Lines to place in method: 185 - 189

### How?

## Additional Notes

* Code review
* Ensure System tests pass

JackEAllen:
### Summary of work
Move logging message and actions to send reduction started to AMQ into a sub-method inside of reduce()
The method could be called: reduction_started

### How to test your work
This will be hard to test until methods are eventually moved outside of `reduce()`.
* Code review
* Ensure System tests pass

### Additional comments
This will not be a permanent change within the `reduce()` method. 
It is one of many small incremental changes to compartmentalise different processes taking place within `reduce()` before moving outside of the reduce method one by one to add unit tests

Fixes #621

### DIFF
--- a/queue_processors/autoreduction_processor/tests/test_post_process_admin.py
+++ b/queue_processors/autoreduction_processor/tests/test_post_process_admin.py
@@ -121,9 +121,7 @@ class TestPostProcessAdmin(unittest.TestCase):
         ppa = PostProcessAdmin(self.message, amq_client_mock)
         ppa.reduction_started()
 
-        mock_log.assert_called_with("Calling: %s\n%s",
-                                    ACTIVEMQ_SETTINGS.reduction_started,
-                                    self.message.serialize(limit_reduction_script=True))
+        mock_log.assert_called()
         amq_client_mock.send.assert_called_with(ACTIVEMQ_SETTINGS.reduction_started,
                                                      ppa.message)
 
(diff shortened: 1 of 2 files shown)


JackEAllen:

queue_processors/autoreduction_processor/tests/test_post_process_admin.py
```diff
@@ -24,6 +24,7 @@
 from queue_processors.autoreduction_processor.post_process_admin import (windows_to_linux_path,
                                                                          PostProcessAdmin,
                                                                          main)
+# from queue_processors.autoreduction_processor.post_process_admin
```
DavidFair, line 27: Drop this unused import if were no longer using it

queue_processors/autoreduction_processor/tests/test_post_process_admin.py
```diff
@@ -111,6 +112,22 @@ def test_load_reduction_script(self):
...
+    def test_reduction_started(self, mock_log):
+        """
+        Test: reduction started message has been sent and logged
+        When: called within reduce method
+        """
+        amq_client_mock = Mock()
+        ppa = PostProcessAdmin(self.message, amq_client_mock)
+        ppa.reduction_started()
+
+        mock_log.assert_called_with("Calling: %s\n%s",
```
DavidFair, line 125: This is a nit pick, but I try to avoid `assert_called_with` in favour of `assert_called` for logging;

This means someone can update the logging without changing the tests so that we don't become tightly coupled. Since this only really needs to be human readable.

However the `assert_called_once_with` uses a set API so should use the `*_with` variant

PR #647, opened 2020-07-01
Place the specifying of directory locations which Autoreduction shoul…

### LINKED ISSUE #622: Move Specifying of Instrument Directories into Sub-method
Issue raised by: developer

### What?
Move processes related to the specifying of instrument directories in relation to starting a reduction job into a sub-method. 

This will include the removal of the run number for excitations at the end and the location of output directories. 

### Where?
* Lines to place in method: 191 - 201

### How?

## Additional Notes

* Code review
* Ensure System tests pass

JackEAllen:
### Summary of work
Move processes related to the specifying of instrument directories in relation to starting a reduction job into a sub-method.

This will include the removal of the run number for excitations at the end and the location of output directories.

### How to test your work
* Code review
* Ensure System tests pass

## Additional comments
This will not be a permanent change within the reduce() method. 
It is one of many small incremental changes to compartmentalise different processes before moving outside of the reduce method one by one to add unit tests

Fixes #622 

### DIFF
--- a/queue_processors/autoreduction_processor/post_process_admin.py
+++ b/queue_processors/autoreduction_processor/post_process_admin.py
@@ -182,14 +182,19 @@ class PostProcessAdmin:
                                        instrument_output_directory,
                                        excitations,
                                        temporary_directory):
-        """Specifies instrument directories, including removal of run_number folder
+        """
+        Specifies instrument directories, including removal of run_number folder
         if excitations instrument
-        :return (str) Directories where Autoreduction should output"""
+        :param instrument_output_directory: (str) Ceph directory using instrument, proposal, run no
+        :param excitations: (list) List of excitations instruments
+        :param temporary_directory: (str) Temp directory location (root)
+        :return (str) Directories where Autoreduction should output
+        """
 
         if self.instrument in excitations:
             # Excitations would like to remove the run number folder at the end
-            instrument_output_directory = instrument_output_directory[
-                                          :instrument_output_directory.rfind('/') + 1]
+            no_run_number_directory = instrument_output_directory.rfind('/') + 1
+            instrument_output_directory = instrument_output_directory[:no_run_number_directory]
 
         # Specify directories where autoreduction output will go
         return temporary_directory + instrument_output_directory
--- a/queue_processors/autoreduction_processor/post_process_admin.py
+++ b/queue_processors/autoreduction_processor/post_process_admin.py
@@ -177,24 +177,31 @@ class PostProcessAdmin:
         """ Returns the path of the reduction script for an instrument. """
         return os.path.join(self._reduction_script_location(instrument_name), 'reduce.py')
 
-
     def specify_instrument_directories(self,
                                        instrument_output_directory,
-                                       excitations,
+                                       no_run_number_directory,
                                        temporary_directory):
         """
         Specifies instrument directories, including removal of run_number folder
         if excitations instrument
         :param instrument_output_directory: (str) Ceph directory using instrument, proposal, run no
-        :param excitations: (list) List of excitations instruments
+        :param no_run_number_directory: (bool) Determines whether or not to remove run_number from dir
         :param temporary_directory: (str) Temp directory location (root)
         :return (str) Directories where Autoreduction should output
         """
 
-        if self.instrument in excitations:
-            # Excitations would like to remove the run number folder at the end
-            no_run_number_directory = instrument_output_directory.rfind('/') + 1
-            instrument_output_directory = instrument_output_directory[:no_run_number_directory]
+        directory_list = [i for i in instrument_output_directory.split('/') if i]
+
+        if directory_list[-1] != f"{self.run_number}":
+            return ValueError("directory does not follow expected format "
+                              "(instrument/RB_no/run_number) \n"
+                              "format: \n"
+                              "%s", instrument_output_directory)
+
+        if no_run_number_directory is True:
+            # Remove the run number folder at the end
+            remove_run_number_directory = instrument_output_directory.rfind('/') + 1
+            instrument_output_directory = instrument_output_directory[:remove_run_number_directory]
 
         # Specify directories where autoreduction output will go
         return temporary_directory + instrument_output_directory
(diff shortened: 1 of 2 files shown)


JackEAllen:

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -188,17 +188,24 @@ def reduce(self):
...
-                                                              self.proposal,
-                                                              self.run_number)
-
-            if self.instrument in MISC["excitation_instruments"]:
-                # Excitations would like to remove the run number folder at the end
-                instrument_output_dir = instrument_output_dir[:instrument_output_dir.rfind('/') + 1]
-
-            # Specify directories where autoreduction output will go
-            reduce_result_dir = MISC["temp_root_directory"] + instrument_output_dir
+            def specify_instrument_directories():
```
DavidFair, line 191: Can we pass in the variables we use in the function, rather than using globals for the same reasons as in the other PR (#649 ) I reviewed

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -188,17 +188,24 @@ def reduce(self):
...
+                if excitations instrument
+                :return (list) Directories where Autoreduction should output"""
+                # Specify instrument directory
+                instrument_output_dir = MISC["ceph_directory"] % (self.instrument,
+                                                                  self.proposal,
+                                                                  self.run_number)
+
+                if self.instrument in MISC["excitation_instruments"]:
+                    # Excitations would like to remove the run number folder at the end
+                    instrument_output_dir = instrument_output_dir[
```
DavidFair, line 202: Can we assert that the run number folder exists before blindly removing it. As if someone does this elsewhere this will end up putting us into one parent higher than we expect

JackEAllen:
@DavidFair  Thanks for the reviews, I should have probably left a comment under all of these PR's to say that any that have the "Don't Merge" tag are going to change quite significantly after discussions in this morning catchup.

JackEAllen:
This PR is currently not ready for review

ElliotAOram:
@JackEAllen if not ready for review, please use the `in progress` tag

JackEAllen:
This is now ready for review 🙂

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -177,6 +177,28 @@ def _load_reduction_script(self, instrument_name):
...
+
+    def specify_instrument_directories(self,
+                                       instrument_output_directory,
+                                       excitations,
+                                       temporary_directory):
+        """
+        Specifies instrument directories, including removal of run_number folder
+        if excitations instrument
+        :param instrument_output_directory: (str) Ceph directory using instrument, proposal, run no
+        :param excitations: (list) List of excitations instruments
```
DavidFair, line 189: Since the list of excitation instruments isn't likely to change could we either:

- (Preferred) Change the param to `remove_run_no_dir` then make the caller do the check, this also means if other instruments want the same we don't have to touch this code later
- Alternatively, store the list of excitation instruments in the class init, since we already store the current instrument too so it doesn't get passed in

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -177,6 +177,28 @@ def _load_reduction_script(self, instrument_name):
...
+        :param temporary_directory: (str) Temp directory location (root)
+        :return (str) Directories where Autoreduction should output
+        """
+
+        if self.instrument in excitations:
+            # Excitations would like to remove the run number folder at the end
+            no_run_number_directory = instrument_output_directory.rfind('/') + 1
```
DavidFair, line 196: Were still not asserting or checking that `instrument_output_directory` has the right format

If someone changed the call from `ALF/RB123/50005` to just `ALF/RB123` (such as a bug) we would just dump data into `ALF/`causing someone a painful cleanup down the road rather than throwing an assert error

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -194,17 +216,13 @@ def reduce(self):
...
-
-            if self.instrument in MISC["excitation_instruments"]:
-                # Excitations would like to remove the run number folder at the end
-                instrument_output_dir = instrument_output_dir[:instrument_output_dir.rfind('/') + 1]
-
-            # Specify directories where autoreduction output will go
-            reduce_result_dir = MISC["temp_root_directory"] + instrument_output_dir
+            # Specify instrument directories
+            reduce_result_dir = self.specify_instrument_directories(
+                instrument_output_directory=MISC["ceph_directory"] % (self.instrument,
```
DavidFair, line 221: Nit: For readability:
```python
instrument_output_dir = MISC["ceph_directory"] % (....)
reduce_result_dir = self.specify_instrument_directories(
           instrument_output_directory=inst_output_dir,
           excitations=MISC[...],
           temporary_directory=MISC[....])

File: queue_processors/autoreduction_processor/tests/test_post_process_admin.py [hunk 6]
```diff
@@ -111,6 +111,30 @@ def test_load_reduction_script(self):
...
+            temporary_directory=MISC["temp_root_directory"])
+
+        directory_list = [i for i in actual.split('/') if i]
+
+        expected_end = ['reduced-data', 'WISH', 'RB1234', 'autoreduced', '27282']
+        expected_start = 'autoreducetmp'
+
+        self.assertEqual(directory_list[-5:], expected_end)
+        self.assertEqual(directory_list[0], expected_start)
+        self.assertIsInstance(actual, str)
```
DavidFair, line 136: We don't need this last assert since `assertEqual(2782, "2782")` is not true, so the fact our expected is already a string means this is guaranteed to be a string too

DavidFair: Also were hard coding `reduced-data` and `autoreduced` which makes for a brittle test, since these can change elsewhere causing an unrelated failure here for someone to come in and fix later. 
Is there a way to just get the values from `MISC` here rather than hard coding them?

JackEAllen:
Ready for review

PR #649, opened 2020-07-01
Create two sub-methods to handle final result directory path

### LINKED ISSUE #624: Path Striping of Temporary Directories for Final Archives Directory
Issue raised by: developer

### What?
Move processes related to the initialisation of variables related to striping of temp paths in front of temp directories for final archives directory into sub-method. 
This should also include the logging of final result and log directories. 

### Where?
* Lines to place in method:  211 - 220

### How?

## Additional Notes

* Code review
* Ensure System tests pass

JackEAllen:
### Summary of work
Move processes related to the initialisation of variables related to striping of temp paths in front of temp directories for final archives directory into sub-method.
This should also include the logging of final result and log directories.

### How to test your work
* Code review
* Ensure System tests pass

### Additional comments
This will not be a permanent change within the reduce() method.
It is one of many small incremental changes to compartmentalise different processes before moving outside of the reduce method one by one to add unit tests

Fixes #624


JackEAllen:

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -208,16 +208,29 @@ def reduce(self):
...
-
-            logger.info('Final Result Directory = %s', final_result_dir)
-            logger.info('Final Log Directory = %s', final_log_dir)
+            def result_directory():
+                """
+                create and return final result directory, stripping temporary path off the front
+                of temporary directories
+                :return (str) final result directory path
+                """
+                final_result_directory = reduce_result_dir[len(MISC["temp_root_directory"]):]
```
DavidFair, line 217: Shame were not up to Python 3.9 else we could have done `reduce_result_dir.removeprefix(MISC["temp_root_directory"])`

Could we do either an assert (if we know this always should be true) or an if for that:
`reduce_result_dir.starts_with(MISC["temp_root_directory"])` before slicing

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -208,16 +208,29 @@ def reduce(self):
...
+                final_result_directory = self._new_reduction_data_path(final_result_directory)
+                logger.info(f"Final Result Directory = {final_result_dir}")
```
DavidFair, line 219: Pylint (rightly) points out we should not use f strings for logging, since we force the interpreter to eval this now. Instead you should do

`logger.info("Final Result Directory = %s", final_result_dir)`

Then the logger can avoid the operation if were not using info and below. This keeps it consistent if we also add debug too later on

queue_processors/autoreduction_processor/post_process_admin.py
DavidFair: Also could the caller pass in these fields at the point of call? I know they're global, but this slowly fixes the global usage and makes it trivial to unit test as a free function

Something like:
`def result_directory(reduce_result_dir, root_directory_dict)` ?

JackEAllen:
This PR is currently not ready for review

JackEAllen:
This is now ready for review - Slightly confused about travis failing the build. Seems to believe trailing whitespace and unused variable are present.

DavidFair:
Code-wise I'm happy

@JackEAllen Pylint is also failing on a unused var which needs resolving first

JackEAllen:
Build fails as Travis still seems to think trailing whitespace is present.

ElliotAOram:
Still can't work out the issue here, but we should try and work this out before merging else it's going to break all the rest of the builds

ElliotAOram:
perhaps we can add a disable

JackEAllen:
I have no idea how the change i made fixed the persistent trailing whitespace error that should't even exist, but it does.

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -184,10 +184,35 @@ def reduction_started(self):
                      self.message.serialize(limit_reduction_script=True))
         self.client.send(ACTIVEMQ_SETTINGS.reduction_started, self.message)
 
+    def result_and_log_directory(self, temporary_root_directory, reduce_dir):
```
keiranjprice101, line 187: Could this be named to something more descriptive?

I prefer something like `create_result_and_log_directories` but I am not fussed. 

I think a string of nouns is always confusing for a func/method name, and I think the singular directory is misleading, when a tuple is returned.

PR #650, opened 2020-07-03
Shutdown QP client on timer gracefully

DavidFair:
### Summary of work
- Update `QueueClient` to call stomp with the magic combo of params to actually wait for the listener thread to finish
- Prep work for QP -> Queue Listener including the shell scripts
- Add stop mechanism to QP Daemon
- Add timer to shutdown after 23 hours 45 minutes, the cron job can run every 24 hours then with a 15 minute shutdown margin
- Log error if after 1 minute we have not shutdown yet. We don't try and be clever and do `sys.exit` as at this point attaching `gdb` is more important than blowing the stack and making it tomorrows problem
- Unit tests :tada: 

### How to test your work
- Wait for #645 to be merge (as we will develop merge conflicts)
- Code review whilst we wait
- Test that Daemon still starts up correctly

### Additional comments
- I can unit test this locally, but I haven't integration tested it (with ACTIVEMQ and trying to Daemonise it) as I would need to setup a messaging server then my own settings files for the little time I have. Therefore to the reviewer (sorry) you'll need to briefly integration test it.

Fixes: #651 

##Release Notes

None yet but were close :wink:


queue_processors/queue_processor/tests/test_queue_listener_daemon.py
```diff
@@ -0,0 +1,116 @@
...
+"""
+Tests the Queue Listener Daemon
+"""
+
+import unittest
+from unittest import mock
+
+from queue_processors.queue_processor.queue_listener_daemon import \
+    QueueListenerDaemon
+import queue_processors.queue_processor.queue_listener_daemon
```
JackEAllen, line 16: Do we need both of these imports or could we just use one?

queue_processors/queue_processor/tests/test_queue_listener_daemon.py
```diff
@@ -0,0 +1,116 @@
...
+    @staticmethod
+    @mock.patch("queue_processors.queue_processor."
+                "queue_listener_daemon.QueueListenerDaemon")
+    @mock.patch("queue_processors.queue_processor."
+                "queue_listener_daemon.control_daemon_from_cli")
+    @mock.patch("logging.error")
+    def test_main(patched_logging, patched_cli, patched_daemon):
+        """
+        Tests the main method correctly starts and handles the nominal case
+        """
```
JackEAllen, line 35: Could we adopt the new format we use for unit tests? e.g.
```
   """
   Test: main method correctly handles the nominal case
   When: called, called with etc...
   """
```

queue_processors/queue_processor/tests/test_queue_listener_daemon.py
```diff
@@ -0,0 +1,116 @@
...
+        patched_logging.assert_not_called()
+
+    @staticmethod
+    @mock.patch("queue_processors.queue_processor."
+                "queue_listener_daemon.QueueListenerDaemon")
+    @mock.patch("logging.error")
+    def test_main_timeout_logs(patched_logging, patched_daemon):
+        """
+        Tests main will correctly log a timeout as an error
+        """
```
JackEAllen, line 52: Could we adopt the new format we use for unit tests here?

queue_processors/queue_processor/tests/test_queue_listener_daemon.py
```diff
@@ -0,0 +1,116 @@
...
+    def setUp(self):
+        self.instance = QueueListenerDaemon(pidfile="/should/never/be/created")
+
+    @mock.patch("queue_processors.queue_processor."
+                "queue_listener_daemon.queue_processor")
+    @mock.patch("threading.Timer")
+    def test_run(self, patched_timer, patched_processor):
+        """
+        Tests that run method sets up the stop timer and the main worker thread
+        """
```
JackEAllen, line 70: Could we adopt the new format we use for unit tests here?

queue_processors/queue_processor/tests/test_queue_listener_daemon.py
```diff
@@ -0,0 +1,116 @@
...
+        patched_timer.assert_called_once_with(mock.ANY, self.instance.stop)
+        self.assertEqual(self.instance._client_handle,
+                         patched_processor.main.return_value)
+
+    @mock.patch("queue_processors.daemon.Daemon.stop")
+    def test_stop(self, super_class_stop):
+        """
+        Tests the stop method correctly shuts down the listener and then
+        marks the thread as safely stopped
+        """
```
JackEAllen, line 83: Could we adopt the new format we use for unit tests here?

queue_processors/queue_processor/tests/test_queue_listener_daemon.py
```diff
@@ -0,0 +1,116 @@
...
+
+        mocked_client.disconnect.assert_called_once()
+        super_class_stop.assert_called_once()
+        mocked_event.set.assert_called_once()
+
+    def test_timer_fires_stop(self):
+        """
+        Tests the timer correctly calls stop (without testing stop)
+        when a given time elapses
+        """
```
JackEAllen, line 99: Could we adopt the new format we use for unit tests here?

queue_processors/queue_processor/tests/test_queue_listener_daemon.py
```diff
@@ -0,0 +1,116 @@
...
+                "queue_listener_daemon.queue_processor")
+    @mock.patch("threading.Timer")
+    def test_run(self, patched_timer, patched_processor):
+        """
+        Tests that run method sets up the stop timer and the main worker thread
+        """
+        self.instance.run()
+
+        patched_processor.main.assert_called_once()
+        patched_timer.assert_called_once_with(mock.ANY, self.instance.stop)
```
JackEAllen, line 74: NITPICK: Can we always be certain that the type will be correct given mock.ANY doesn't perform type checking?

utils/clients/queue_client.py
```diff
@@ -57,7 +58,12 @@ def disconnect(self):
         """
         logging.info("Disconnecting from activemq")
         if self._connection is not None and self._connection.is_connected():
-            self._connection.disconnect()
+            # By passing a receipt Stomp will call stop on the transport layer
+            # which causes it to wait on the listener thread (if it's still
+            # running). Without this step we just drop out, so the behaviour
+            # is not guaranteed. UUID is used by Stomp if we don't pass in
+            # a receipt so this matches the behaviour under the hood
```
JackEAllen, line 65: NITPICK: Because of the size of this comment, could we not just turn it into a docstring?

JackEAllen:
From manual testing on develop node, when using restart.sh after stopping processes using stop.sh, only `queue_listener_daemon.py` seems to start. Is this the intended behaviour for this issue?

JackEAllen:
@DavidFair From manual testing, when stopping both processes using `stop.sh` on attempt to then restart both processes, using restart.sh, only `queue_listener_daemon.py` starts.

queue_processors/queue_processor/tests/test_queue_listener_daemon.py
DavidFair: This is a subtlety with patching that I really dislike and run into all the time:

If our module is `foo.bar`

`from x import y` we now need to patch `foo.bar.y`
`import x` we now patch `x.y`

As we are patching an import happening in module `x` rather that `foo` it makes it much easier to just do a bare import rather than a from import then try and decipher the new namespace.

For this reason I've left it as a `from` and `import`, any other time agree completely

queue_processors/queue_processor/tests/test_queue_listener_daemon.py
DavidFair: True, but equally we have to start pulling apart call args and it violates the Python constraint EAFP

My thought process is:
If we pass in the wrong arg for the method invoked it's not immediately obvious and you need to wait 24 hours to find out
If you pass in the wrong arg (which could be say a `datetime` instead of an int) for the first param the Daemon will die before starting since the Timer class checks the types beforehand

utils/clients/queue_client.py
DavidFair: In this case, no, since it's an implementation detail of stomp and shouldn't really be documented in the API doc string IMO

JackEAllen:
> @DavidFair From manual testing, when stopping both processes using `stop.sh` on attempt to then restart both processes, using restart.sh, only `queue_listener_daemon.py` starts.
> so commands in this order:
> 
> * `./restart.sh` - Works
> * `./stop.sh` - Works
> * `./restart.sh` - Now Fails starting only `queue_listener_daemon.py`

From manual testing since last changes were made, this now works! 🎉 
@DavidFair Is this PR ready for another code review?

DavidFair:
Yes, should be ready for another round

This will end up having to go through a rebase once #645 is merged, so feel free to be as picky as you'd like since we have at least another code review after this too

DavidFair:
Blocked by #645

PR #657, opened 2020-07-10
Unpin test deps

DavidFair:
### Summary of work
Drops the test package requirements from pinned to the current stable.

I noticed this whilst adding `pytest-django` to `requirements.txt` in another branch, the majority of the pinning stems from Python 2 support being dropped in our deps

### How to test your work
- CI Tests should pass


ElliotAOram:
Looks good

PR #659, opened 2020-07-13
656 interactive tour instrument summary

JackEAllen:
### Summary of work
Extended Virtual interactive tour to instrument summary page

### How to test your work
* Code review
* Ensure all descriptions are grammatically correct.
* Ensure that tour descriptions are ordered in logical manner i.e descriptions are linear to how a user may navigate through webpage.

Fixes 656

### DIFF
--- a/WebApp/autoreduce_webapp/static/javascript/tours/instrument_summary_tour.js
+++ b/WebApp/autoreduce_webapp/static/javascript/tours/instrument_summary_tour.js
@@ -1,0 +1,58 @@
+steps = [
+    {
+        element: "#runs-table",
+        title: "Runs Table",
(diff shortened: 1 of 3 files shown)


WebApp/autoreduce_webapp/static/javascript/tours/instrument_summary_tour.js
```diff
@@ -0,0 +1,64 @@
+steps = [
+    {
+        element: "#not-active",
```
ElliotAOram, line 3: This works only when the instrument is not active. We should change this step to say instrument is not active - I am actually going to remove this from the db soon as we don't use it so no need to elaborate.

PR #637, opened 2020-06-29
Fire for CLI

ElliotAOram:
### Summary of work
* Change the custom CLI with argparse to use fire

### How to test your work
* Try out manual submission and ensure it still work as expected

### Additional comments
* We should look to do this with any other CLIs in autoreduction (manual_remove, database_operations)

*No associated issue*

:fire::fire::fire:

PR #662, opened 2020-07-13
643 fire cli for manual remove

### LINKED ISSUE #643: Fire CLI for manual remove script
Issue raised by: [developer]

### What?
After the successful changes to manual_submission to make it use fire CLI #637 we want to do the same for the manual_remove script. Similar to manual submission we want to replace the argparse information and handling with fire CLI. See the linked pr #637 for details

### Where?
* unit tests: `scripts/manual_operations/tests/test_manual_remove.py`
* source code: `scripts/manual_operations/tests/test_manual_submission.py`

### How?
Continuation of using new CLI tool

* Use the manual_remove script (following the README.md instructions `scripts/manual_operations/README.md`)
* Ensure these functions still pass
* Ensure all travis unit tests pass and are updated where required

keiranjprice101:
### Summary of work

Used #637 as a guide. Not sure if I should be adding any labels, so please let me know :)

I added mocking for `user_input_check` as without it the test would hang until y/n was entered. (I don't understand why this wasn't happening previously)

I didn't add any type casting as from the docs https://google.github.io/python-fire/guide/#argument-parsing this should be automatic.

### How to test your work
 
The only test failing is `test_find_run` though this was failing already due to my sub optimal setup.
### Additional comments
 

closes #643 

### DIFF
--- a/scripts/manual_operations/manual_remove.py
+++ b/scripts/manual_operations/manual_remove.py
@@ -216,6 +216,9 @@ def user_input_check(instrument, run_numbers):
 def main(instrument, first_run, last_run=None):
     """
     Parse user input and run the script to remove runs for a given instrument
+    :param instrument: Instrument to run on
+    :param first_run: First run to be removed
+    :param last_run: Optional last run to be removed
     """
     run_numbers = [first_run]
 
--- a/scripts/manual_operations/manual_remove.py
+++ b/scripts/manual_operations/manual_remove.py
@@ -213,7 +213,7 @@ def user_input_check(instrument, run_numbers):
     return user_input
 
 
-def main(instrument, first_run, last_run=None):
+def main(instrument: str, first_run: int, last_run: int = None):
     """
     Parse user input and run the script to remove runs for a given instrument
     :param instrument: Instrument to run on
--- a/scripts/manual_operations/manual_remove.py
+++ b/scripts/manual_operations/manual_remove.py
@@ -225,8 +225,7 @@ def main(instrument, first_run, last_run=None):
     if last_run:
         if last_run < first_run:
             print(f"last run {last_run} must be greater than first run {first_run}")
-            print("e.g ./manual_remove.py GEM 83880 83882")
-            sys.exit(1)
+            raise ValueError(f"last run: {last_run} was smaller than first run: {first_run}")
         run_numbers = range(first_run, last_run + 1)
 
     instrument = instrument.upper()
--- a/scripts/manual_operations/manual_remove.py
+++ b/scripts/manual_operations/manual_remove.py
@@ -216,9 +216,9 @@ def user_input_check(instrument, run_numbers):
 def main(instrument: str, first_run: int, last_run: int = None):
     """
     Parse user input and run the script to remove runs for a given instrument
-    :param instrument: Instrument to run on
-    :param first_run: First run to be removed
-    :param last_run: Optional last run to be removed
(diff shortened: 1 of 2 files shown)


scripts/manual_operations/tests/test_manual_remove.py
```diff
@@ -208,16 +205,14 @@ def test_main(self, mock_delete, mock_process, mock_find):
     @patch('scripts.manual_operations.manual_remove.ManualRemove.find_runs_in_database')
     @patch('scripts.manual_operations.manual_remove.ManualRemove.process_results')
     @patch('scripts.manual_operations.manual_remove.ManualRemove.delete_records')
-    @patch('scripts.manual_operations.manual_remove.handle_input')
-    def test_main_range(self, mock_handle_input, mock_delete, mock_process, mock_find):
+    @patch('scripts.manual_operations.manual_remove.user_input_check')
+    def test_main_range(self, mock_delete, mock_process, mock_find, mock_uic):
```
JackEAllen, line 209: It's a bit of a subtlety of python here where the patched method parameters are read in as arguments from bottom to top so mock_uic should be placed just after `self` such as `def test_main_range(self, mock_uic, mock_delete, mock_process, mock_find):`
Right now, mock_uic is representing `@patch('scripts.manual_operations.manual_remove.ManualRemove.find_runs_in_database')`

JackEAllen, line 209: Could we have 2 more tests here? 
* Test the removal of an individual run and also a range of runs less than 10.
* Test invalid input - so maybe something to asset that an error is raised if the first run > end run

scripts/manual_operations/tests/test_manual_remove.py
```diff
@@ -231,20 +231,6 @@ def test_main_range_less_than_ten(self, mock_delete, mock_process, mock_find):
         mock_process.assert_called()
         mock_delete.assert_called()
 
-    # pylint:disable=no-self-use
-    @patch('scripts.manual_operations.manual_remove.ManualRemove.find_runs_in_database')
-    @patch('scripts.manual_operations.manual_remove.ManualRemove.process_results')
-    @patch('scripts.manual_operations.manual_remove.ManualRemove.delete_records')
-    def test_main_single_run(self, mock_delete, mock_process, mock_find):
```
keiranjprice101: This was just the exact same as test_main 🤦

scripts/manual_operations/manual_remove.py
```diff
@@ -242,11 +213,22 @@ def user_input_check(instrument, run_numbers):
     return user_input
 
 
-def main():
+def main(instrument: str, first_run: int, last_run: int = None):
```
JackEAllen, line 216: This is probably my bad providing a poor description XD What i mean by adding the types is that we add the types in the docstring themselves so something like this:
`  :param instrument: (str) Instrument to run on
    :param first_run: (str) First run to be removed
    :param last_run: (str) Optional last run to be removed`

JackEAllen: As soon as this is changed, you can merge in 👍

keiranjprice101: My mistake 👍

keiranjprice101:
@JackEAllen I notice you changed the base from develop to master. Are we merging straight into master now?

JackEAllen:
Yes, as of yesterday 🙂

PR #666, opened 2020-07-14
665 upgrade twisted

### LINKED ISSUE #665: Upgrade Twisted version
Issue raised by: [developer]

### What?
We should update the twisted version to avoid security concern with current version. 

### Where?
setup.py

### How?
Raised by git security bot

* Travis build passes

keiranjprice101:
### Summary of work

Unpin the version of Twisted in `setup.py`
### Additional comments
 

Fixes #665

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ from build.commands.migrate_settings import MigrateTestSettings
 from build.commands.start import Start
 
 
-setup_requires = ['attrs',
+setup_requires = ['attrs==19.2.0',
                   'dash',
                   'dash_html_components',
                   'dash_core_components',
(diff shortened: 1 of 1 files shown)


ElliotAOram:
Cool so it looks like there is something different in the new Twisted package that our code base doesn't like very much (as I suspected hence why we pinned the version) we'll have to have a look into this

ElliotAOram:
This looks like it could be helpful here:
https://github.com/HypothesisWorks/hypothesis/issues/2113

ElliotAOram:
This is now resolved. I've had to pin `attrs` to be a higher version than the default for python 3.7 (current version running on travis). When Mantid catch up to python 3.8, we can update our own python version and remove the pin for attrs.

PR #667, opened 2020-07-14
Add unit tests to webapp

DavidFair:
### Summary of work
- Fixes up Pytest to work gracefully with Django.
- Adds an example test
- Does some fancy rebase and merge to get the commits from #657 into `master` to avoid a merge conflict when develop goes in too   (note this this PR was created develop branch has been deleted)

Now developers like myself don't have an excuse for not writing unit tests for the web app :wink: 

### How to test your work
- Install the new Python packages requires from `requirements.txt`
- Locally, try running `pytest WebApp` (or alternative `pytest` in your root dir if testing is setup for other frameworks). This should "just work"

### Additional comments
- TODO Re-run all our migrations each time we test to ensure they actually work, rather than hoping we can ever get back the same schema

Does this apply to non-user facing changes? ^


DavidFair:
@ElliotAOram I won't have a chance to complete this for a while, but to help you weigh up effort of completing it:

Pros:
- Barrier to entry is removed. I genuinely think this is why our coverage is 30%
- Database is created in memory. This means nobody should need to go in and update our `.sql` file as tests change reducing the barrier to entry
- As the DB is fresh each time we can't get into a state where someone's dev machine doesn't match the test server
- Fast setup for new devs, don't need to faff with MySQL instances
- Can test Django rather than Django side effects from viewing the DB

Cons:
- We need to swap out all our tests where we manually connect using the MySQL connector to use `django.db.connection` instead. This is also best practice as Django should handle logging in / setting up instance for us anyway
- Existing tests which rely on data from `.sql` file need to be rewritten to use mocks (preferred) or insert the data into the DB before the test runs 
- Mismatch between MySQL and SQLite DB, this should only bite us if we start doing things behind the ORM's back


What are your thoughts?

ElliotAOram:
@DavidFair this is definitely the right direction for the project. I'll try and pick this up myself when I get time.

pytest.ini
```diff
@@ -1,2 +1,8 @@
 [pytest]
-norecursedirs = .* dist CVS _darcs {arch} *.egg
\ No newline at end of file
+DJANGO_SETTINGS_MODULE = autoreduce_webapp.test_settings
```
Anders-Markvardsen, line 2: Is it strictly necessary that DJANGO_SETTINGS_MODULE is defined here? Could it be defined just in test_admin.py?

pytest.ini
```diff
@@ -1,2 +1,8 @@
...
+norecursedirs = .* dist CVS _darcs {arch} *.egg
+# We have to pretend that the root of the project also exists in
+# WebApp/autoreduce_webapp since trying to import stuff through it
+# will break Django. (Really we should split out everything in
+# WebApp/autoreduce_webapp into its own repo in the future)
+python_paths = WebApp/autoreduce_webapp
```
Anders-Markvardsen, line 8: Could defining python_paths in top level pytest.ini have a side effect somewhere else?

WebApp/autoreduce_webapp/tests/reduction_variables/test_admin.py
```diff
@@ -0,0 +1,27 @@
...
+# ################################################################################
+
+import unittest
+from unittest import mock
+
+
+class TestAdmin(unittest.TestCase):
+
+    @mock.patch("django.contrib.admin")
+    def test_something(self, admin_mock):
```
Anders-Markvardsen, line 15: Would it be possible to have a more descriptive name of this test?

Anders-Markvardsen:
Also agree the general direction of this seems right and with test_admin.py. Initially a bit concerned with using SQLite DB over MySQL since the latter is what we use in production, however as long as the latter is used in systemtests

requirements.txt
```diff
@@ -9,8 +11,10 @@ lxml
 mock
 pandas
 plotly
-pylint==2.4.4
```
Anders-Markvardsen, line 12: how confident are we that removing specific version is not causing issues somewhere else... a sanity question

Anders-Markvardsen:
@dtasev could you have a look at this one. This PR is starting to age. See if a version of it can be merged and open follow up issues as needed for later

pytest.ini
dtasev: I don't think anything reads these config files apart from the tools they're intended for - this should be OK

requirements.txt
dtasev: I've found that keeping linters/formatters to the latest version rarely (but likely not never) causes issues, and if it does it will cause some builds to fail because of new rules (or bug fixes that didn't apply previous rules), but no runtime errors will pop up randomly because of them

Overall this seems OK to me

dtasev:
I am not sure why the tests are failing - perhaps because `test_settings.py` no longer uses the MySQL credentials and some tests are expecting that?

I am not quite sure why this PR changes the DB backend to SQLite3, but maybe that is related to the test failures?

requirements.txt
Anders-Markvardsen: thanks for 2nd opinion and ok

Anders-Markvardsen:
> I am not sure why the tests are failing - perhaps because `test_settings.py` no longer uses the MySQL credentials and some tests are expecting that?
> 
> I am not quite sure why this PR changes the DB backend to SQLite3, but maybe that is related to the test failures?

From what I understand above using SQLite3 for unit tests provide a speedup over MySQL. However, if you think this speed up is not needed please use stay with MySQL, and where needed this can be changed to SQLite3 at a later stage

dtasev:
With a few tests failing on this PR and not on master I wonder if it's worth reverting the DB changes? We could also look into fixing the MySQL dependent bits, but that might take longer

dtasev:
Sorry didn't realise Ctrl+Shift+Enter will also close the PR

Anders-Markvardsen:
@dtasev is it ready for testing (i.e. provide a partially improvement) or should it have the In Progress label assigned to it?

PR #670, opened 2020-07-15
Pull out some validation from manual_operations into util.py

### LINKED ISSUE #607: Manual Operations Utility Centralisation
Issue raised by: developer

### What?
Centralise methods performing the same or similar processes in `manual_remove.py` and `manual_submission.py` in a `utils.py` script to reduce redundancy/duplication.

Move methods listed below into separate script `utils.py`:
* `handle_input()`
*` user_input_check()`

### Where?
Scripts to remove redundancy from:
* `scripts/manual_operations/manual_remove.py`
* `scripts/manual_operations/manual_submission.py`

Create utils script here:
* `scripts/manual_operations/utils.py`
* `scripts/manual_operations/tests/test_utils.py`

### How?
A need to develop constantly DRY (don't repeat yourself) code to follow project standards

* Ensure all unit tests for `manual_remove`, `manual_submission` and `utils.py` pass
* No drop in coverage
* Code review
* Manually test `manual_submission.py` and `manual_remove.py`

keiranjprice101:
### Summary of work

Pulled out the validation on run numbers. This also removed the discrepancy where `manual_submission` would not include the `last_run` and `manual_removal` would. 

@JackEAllen I agree with you, that these 2 scripts could easily be combined and be cleaner overall.

### Additional comments
 

Fixes #607

### DIFF
--- a/scripts/manual_operations/tests/test_manual_remove.py
+++ b/scripts/manual_operations/tests/test_manual_remove.py
@@ -203,8 +203,7 @@ class TestManualRemove(unittest.TestCase):
         mock_process.assert_called_once()
         mock_delete.assert_called_once()
 
-    # pylint:disable=no-self-use
-    # pylint:disable=too-many-arguments
+    # pylint:disable=no-self-use, too-many-arguments
     @patch('scripts.manual_operations.manual_remove.ManualRemove.find_runs_in_database')
     @patch('scripts.manual_operations.manual_remove.ManualRemove.process_results')
     @patch('scripts.manual_operations.manual_remove.ManualRemove.delete_records')
--- a/scripts/manual_operations/tests/test_manual_remove.py
+++ b/scripts/manual_operations/tests/test_manual_remove.py
@@ -191,27 +191,33 @@ class TestManualRemove(unittest.TestCase):
     @patch('scripts.manual_operations.manual_remove.ManualRemove.find_runs_in_database')
     @patch('scripts.manual_operations.manual_remove.ManualRemove.process_results')
     @patch('scripts.manual_operations.manual_remove.ManualRemove.delete_records')
-    def test_main(self, mock_delete, mock_process, mock_find):
+    @patch('scripts.manual_operations.manual_remove.get_run_range', return_value=range(1, 2))
+    def test_main(self, mock_get_run_range, mock_delete, mock_process, mock_find):
         """
         Test: The correct control functions are called
         When: The main() function is called
         """
         main(instrument='GEM', first_run=1)
+        mock_get_run_range.assert_called_once_with(1, last_run=None)
         mock_find.assert_called_once_with(1)
         mock_process.assert_called_once()
         mock_delete.assert_called_once()
 
     # pylint:disable=no-self-use
+    # pylint:disable=too-many-arguments
     @patch('scripts.manual_operations.manual_remove.ManualRemove.find_runs_in_database')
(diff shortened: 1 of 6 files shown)


scripts/manual_operations/tests/test_manual_remove.py
```diff
@@ -191,27 +191,33 @@ def test_validate_csv_invalid(self):
...
         When: The main() function is called
         """
         main(instrument='GEM', first_run=1)
+        mock_get_run_range.assert_called_once_with(1, last_run=None)
         mock_find.assert_called_once_with(1)
         mock_process.assert_called_once()
         mock_delete.assert_called_once()
 
     # pylint:disable=no-self-use
+    # pylint:disable=too-many-arguments
```
JackEAllen, line 207: Could we move both pylint disables onto the same line?

scripts/manual_operations/tests/test_util.py
```diff
@@ -0,0 +1,42 @@
+# ############################################################################### #
+# Autoreduction Repository : https://github.com/ISISScientificComputing/autoreduce
+#
+# Copyright &copy; 2020 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+# ############################################################################### #
+
+"""
+Test cases for util functions
```
JackEAllen, line 9: NITPICK: To be explicit about the purpose of the utils, could we reword this slightly to something along the lines of "Unit tests for manual operations utilities"

scripts/manual_operations/tests/test_util.py
```diff
@@ -0,0 +1,42 @@
...
+        When: second_run is None
+        """
+        self.assertEqual(range(1, 2), get_run_range(1))
+
+    def test_get_run_numbers_valid_input_returns_valid_range(self):
+        """
+        Test: Correct range is returned
+        When: second_run is provided
+        """
+        self.assertEqual(range(1, 5), get_run_range(1, last_run=4))
```
JackEAllen, line 34: To maintain a consistent format when calling a method (`get_run_range()`), we should either call just using raw parameters or with explicit argument assignment, but not interchange between the two e.g

Good practice:
* `my_method(1, 2)`
* `my_method(x=1, y=2)`

Bad practice:
* `my_method(1, y=2)`

scripts/manual_operations/tests/test_util.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_get_run_numbers_invalid_input_raises_value_error(self):
+        """
+        Test: ValueError is raised
+        When: first_run is greater than last_run
+        """
+        with self.assertRaises(ValueError):
+            get_run_range(5, last_run=2)
```
JackEAllen, line 42: Same here as above comment

scripts/manual_operations/tests/test_util.py
keiranjprice101: Interesting, I thought the way I implemented it was idiomatic python, since it shows which param is optional. (Even some of  the examples in pep8 do it) Though I like the consistency. I'll change this and below.

ElliotAOram: @keiranjprice101 @JackEAllen 
The way keiran has called this is actually fine as `last_run` is optional. So the intent is clear. having said that for our own sanctity of mind when testing it's easier to read if it's verbose

PR #671, opened 2020-07-17
625 ppar result paths test access

### LINKED ISSUE #625: Encase Test Access to Result Paths in sub-method
Issue raised by: developer

### What?
Encase all processes relating to test access to result paths in sub-method. I suggest calling this method `path_access_validate` for the moment.

Note -  This will be a temporary measure and should ideally be converted into a class later on after breaking up the whole reduce method into sub-methods.

### Where?
* Lines to place in method:  222 -  253

### How?

## Additional Notes

* Code review
* Ensure System tests pass

JackEAllen:
### Summary of work
Refactor processes relating to test access of result and log paths, placing in sub-method outside of `reduce()`.

### How to test your work
* Code review
* No drop in coverage
* Unit tests Pass

### Additional comments
This pull request is part of many small intermediate changes to be made to post_process admin and will likely change again in the next stage of refactoring once the `reduce()` method has been further broken up and unit tested. 

Fixes #625


queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -177,6 +178,13 @@ def _load_reduction_script(self, instrument_name):
         """ Returns the path of the reduction script for an instrument. """
         return os.path.join(self._reduction_script_location(instrument_name), 'reduce.py')
 
+    def reduction_started(self):
```
keiranjprice101, line 181: I guess the name comes from the reduction status. But I think something like `send_reduction_started_status()` is more descriptive and less confusing to a new developer of the project.

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -206,12 +214,62 @@ def specify_instrument_directories(self,
...
         return temporary_directory + instrument_output_directory
 
-    def reduction_started(self):
-        """Log and update AMQ message to reduction started"""
-        logger.debug("Calling: %s\n%s",
-                     ACTIVEMQ_SETTINGS.reduction_started,
-                     self.message.serialize(limit_reduction_script=True))
-        self.client.send(ACTIVEMQ_SETTINGS.reduction_started, self.message)
+    @staticmethod
+    def write_and_readability_checks(directory_list, read_write):
```
keiranjprice101, line 218: I think this would be cleaner as 2 methods.  e.g. `is_directory_writable(directory)` and `is_directory_readable(directory)` This also would not increase run time as this is already called twice on separate loops.

As a side point, this method is only called within `path_access_validate` and if it ever returns false an exception is raised. It might be nicer to simply raise an exception here and let it propagate up the stack. 

e.g.
```python
def verify_directory_readable(directory):
    if not os.access(directory, os.R_OK):
        raise some error
```
Then in the other method:
```python
for directory in readable_directories:
    verify_directory_readable(directory)
for directory in writable_directories:
    verify_directory_writable(directory)

```

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -206,12 +214,62 @@ def specify_instrument_directories(self,
...
+        read_write_map = {"R": "read", "W": "write"}
+        try:
+            read_write_map[read_write]
+        except KeyError:
+            raise KeyError("Invalid read or write input: %s read_write argument must be either"
+                           " 'R' or 'W'" % read_write)
+
+        for location in directory_list:
+            if not os.access(location, getattr(sys.modules[os.__name__], f"{read_write}_OK")):
+                if not os.access(location, os.F_OK):
```
keiranjprice101, line 235: Do we need to test if the directory exists after it fails the read/write permission check fails? 
I believe the only reason the directory wouldn't exist is if it was deleted in the time between it was created and here, or it failed to be created. In the second scenario an `OSError` would have already been raised, so we would be aware of this situation already. Alternatively if it had been deleted, it wouldnt be readable/writable this would fail, the reduction would be aborted and it would be recreated on the next run anyway.

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -206,12 +214,62 @@ def specify_instrument_directories(self,
...
+                    problem = "does not exist"
+                else:
+                    problem = "no %s access", read_write_map[read_write]
+                raise OSError("Couldn't %s %s  -  %s" % (read_write_map[read_write],
+                                                         location,
+                                                         problem))
+            return True
+
+    def path_access_validate(self, should_be_writable, should_be_readable):
```
keiranjprice101, line 244: I feel this name is misleading, I would expect only validation and not the creation of directories.

queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -206,12 +214,62 @@ def specify_instrument_directories(self,
...
+        """
+        Test for access to result paths and raise exception and attempt re-run later if problem
+        :param should_be_writable: (list)
+        :param should_be_readable: (list)
+        """
+        try:
+            # Try to make directories which should exist
+            for path in filter(lambda p: not os.path.isdir(p), should_be_writable):
```
keiranjprice101, line 252: I think this would be easier to read as a list comp i.e.
```python
for path in [p for p in should_be_writable if not os.path.isdir(p)]:
    os.makedirs(path)
```
Though this is completely subjective.

queue_processors/autoreduction_processor/post_process_admin.py
JackEAllen: Though this is't code i have changed as part of this PR (it's just a result of a merge conflict resolution), I agree, will change this to be more verbose

queue_processors/autoreduction_processor/post_process_admin.py
JackEAllen: This is not code i have changed. It is the result of merging Master into 625. I will consider changing this later.

queue_processors/autoreduction_processor/post_process_admin.py
JackEAllen: This is not code i have changed. It is the result of merging Master into 625. I will consider changing this later. The purpose of the PR's focused on the refacter of `reduce()` given this is the core of autoreduction is not to yet change the processes taking place, but simply to break apart reduce into sub-methods to test first as this method is largely untested.

queue_processors/autoreduction_processor/post_process_admin.py
JackEAllen: The purpose of this PR is to not inherently change the processes already taking place, but to simply move processes outside of reduce to first add unit tests before considering the changes which should take place to clean up the script as a whole.
I do however still believe we should test this so we can verbose in understanding the error. If the file exists, that doesn’t necessarily mean that we will have read or write access to the file. 
If you added a command to travis.yml to cat the logs produced by logs/autoreduction_processor.log, you will see that this method is used regularly to create paths that don’t exist when running end-to-end tests. 
If the directory does not exist, we should  try to make it instead of raising an OSERROR and trying the whole reduction again in 6 hours time.

PR #672, opened 2020-07-21
602 autoreduction icat instrument mappings

### LINKED ISSUE #602: Create Mapping for Autoreduction Instruments and ISISICAT Instrument Prefixes
Issue raised by: developer

### What?
Inability to find Instruments and associated runs when manually submitting run numbers if the prefix chosen by ICAT for the instrument doesn't match the full name.

As a temporarily, but potentially long lasting fix, we should create an explicit, hard coded mapping to map all ICAT instrument prefixes to their full instrument name. 

In the future, it would be nice to simply be able to query ISISICAT to get an explicit mapping of full instrument name and associated prefix. 

### Where?
* scripts/manual_operations/manual_submission.py
* utils/clients/tools/isisicat_prefix_mapping.py
* utils/clients/tools/tests/test_isisicat_prefix_mapping.py

### How?
When running manual submission for a given run or range of runs for an instrument, the database is first checked to see if the runs already exist inside Autoreduction. 
If they do not, they are then searched for in ICAT using the instrument name and run number. ICAT doesn't consistently use the full instrument name. 

ICAT picks a prefix (currently unsure is there is any consistency to this picking process). 
This means that when we are searching for an instrument using the full name, the instrument and associated runs may not be found if the prefix differs from the full instrument name. 

For example, the instrument MARI is called MAR in ICAT, but ENGINX is still ENGINX. 

* Ensure there is no drop in coverage
* Ensure all unit tests pass
* Ensure that manual submission works for instruments that use a shortened prefix in ISISICAT such as MARI.
* Code review

rickybassom:
### Summary of work
Replaced the static ICAT_PREFIX_MAP in scripts/manual_operations/manual_submission.py with a dynamic map found by querying ICAT for instrument prefixes.

### How to test your work
- Code review
- No drop in coverage
- Unit tests pass
- Ensure correct ICAT & Autoreduction terminology used
- Ensure that manual submission works for instruments that use a shortened prefix in ISISICAT such as MARI.

Fixes #602


utils/clients/tools/tests/test_isisicat_prefix_mapping.py
```diff
@@ -0,0 +1,41 @@
...
+"""
+Test ICAT prefix mapping
+"""
+import unittest
+
+from utils.clients.tools.isisicat_prefix_mapping import ICAT_PREFIX_MAP
+from utils.settings import VALID_INSTRUMENTS as AUTOREDUCTION_INSTRUMENT_NAMES
+
+
+class TestICATPrefixMappings(unittest.TestCase):
```
ElliotAOram, line 16: We'll need to mock this as this wont work on travis

utils/clients/tools/isisicat_prefix_mapping.py
```diff
@@ -0,0 +1,44 @@
...
+# ############################################################################### #
+"""
+fetch_instrument_fullname_mapping() can be used to map Autoreduction to ICAT instrument prefixes
+"""
+from utils.clients.icat_client import ICATClient
+from utils.settings import VALID_INSTRUMENTS as AUTOREDUCTION_INSTRUMENT_NAMES
+from utils.clients.tools.isisicat_prefix_mapping_logging_setup import logger
+
+
+def fetch_instrument_fullname_mapping():
```
JackEAllen, line 15: NITPICK: Optionally, this method could take an optional argument such as a list of instruments, to return prefixes only for those instruments instead of returning all instruments when called. 
As there are only around 30 instruments in total (and only 10 using AR right now - could grow to 30 eventually) or so for ISIS, whether or not you make this change is entirely up to you.
My thinking here is that you could call the method with not arguments to get all mappings or call it with a list of N instruments to return mappings for N instruments only. This would mean  you don't have to filter through the returned values as you should get exactly what you are expecting.

utils/clients/tools/tests/test_isisicat_prefix_mapping.py
```diff
@@ -0,0 +1,87 @@
...
+
+# pylint:disable=no-self-use,too-few-public-methods,too-many-public-methods
+class MockInstrumentQueryResult:
+    """
+    Mocks result of isisicat_prefix_mapping.client.execute_query for an instrument
+    """
+
+    def __init__(self, name, full_name):
+        self.name = name
+        self.fullName = full_name  # pylint:disable=invalid-name
```
JackEAllen, line 24: If you change this to self.fullname, you should be able to remove the # pylint:disable=invalid-name

utils/clients/tools/tests/test_isisicat_prefix_mapping.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class TestICATPrefixMapping(unittest.TestCase):
+    """
+    Test ICAT prefix mapping
+    """
+
+    @patch('icat.Client.__init__', return_value=None)
+    @patch('utils.clients.tools.isisicat_prefix_mapping.AUTOREDUCTION_INSTRUMENT_NAMES',
```
JackEAllen, line 33: To make some of the patches look a bit neater/fit on one line, may be worth placing regularly referenced paths in a variable such as:
`DIR = "utils.clients.tools.isisicat_prefix_mapping"`
and then reference when needed such as:
`@patch(f"{DIR}.AUTOREDUCTION_INSTRUMENT_NAMES", return_value = ['ENGINGX']`)

utils/clients/tools/tests/test_isisicat_prefix_mapping.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        fetch_instrument_fullname_mapping()
+        mock_execute_query.assert_called_once()
+
+    @patch('icat.Client.__init__', return_value=None)
+    @patch('utils.clients.tools.isisicat_prefix_mapping.AUTOREDUCTION_INSTRUMENT_NAMES',
+           ['INVALIDINSTUMENT'])
+    @patch('logging.Logger.warning')
+    @patch('utils.clients.icat_client.ICATClient.execute_query',
+           returns_value=[MockInstrumentQueryResult("ABCDE", "ABC")])
```
JackEAllen, line 49: We don't need this return value here, we can just do a plain patch 
```suggestion
    @patch('utils.clients.icat_client.ICATClient.execute_query')
```

Also, I think there's a typo here. i don't think returns_value exists and should have been return_value (for this branch anyway, i think this may have actually been introduced by me 😅)

utils/clients/tools/tests/test_isisicat_prefix_mapping.py
```diff
@@ -0,0 +1,87 @@
...
+    @patch('utils.clients.icat_client.ICATClient.execute_query',
+           return_value=[MockInstrumentQueryResult("ENG", "ENGINX"),
+                         MockInstrumentQueryResult("GEM", "GEM")])
+    def test_icat_prefix_mapping_length(self, _mock_instruments, _mock_execute_query):
+        """
+        Test: fetch_instrument_fullname_mapping produces the same number of results as stored in
+              utils.settings.VALID_INSTRUMENTS
+        When: Called when testing to see if the correct number of instruments in prefix mapping
+        """
+        prefix_map = fetch_instrument_fullname_mapping()
```
JackEAllen, line 71: NITPICK: could we re-word this to "actual"

utils/clients/tools/tests/test_isisicat_prefix_mapping.py
```diff
@@ -0,0 +1,87 @@
...
+    def test_icat_prefix_mapping(self, mock_execute_query, _):
+        """
+        Test: If fetch_instrument_fullname_mapping properly maps instrument names map to ICAT
+              instrument prefixes using utils.settings.VALID_INSTRUMENTS
+        When: Called when testing correct mapping
+        """
+        icat_test_instrument = ("ENG", "ENGINX")
+        mock_execute_query.return_value = [MockInstrumentQueryResult(*icat_test_instrument)]
+
+        prefix_map = fetch_instrument_fullname_mapping()
```
JackEAllen, line 86: NITPICK: could we re-word this to "actual"

utils/clients/tools/isisicat_prefix_mapping.py
```diff
@@ -0,0 +1,44 @@
...
+        icat_instruments = client.execute_query("SELECT i FROM Instrument i")
+    except Exception: # pylint:disable=broad-except
+        warning_message = "ICAT instrument query failed"
+        print(warning_message)
+        logger.warning(warning_message)
+        return None
+
+    for instrument_fullname in AUTOREDUCTION_INSTRUMENT_NAMES:
+        icat_instrument = next((x for x in icat_instruments if x.fullName == instrument_fullname),
+                               None)
```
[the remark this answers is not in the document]
rickybassom: execute_query is only called once and returns all the instruments. Then I loop through all the instruments returned from the query.

utils/clients/tools/tests/test_isisicat_prefix_mapping.py
rickybassom: The reason for using camel case is because I'm mocking the result of execute query which returns a list of objects where fullName is an attribute.

utils/clients/tools/tests/test_isisicat_prefix_mapping.py
rickybassom: Which line?

utils/clients/tools/tests/test_isisicat_prefix_mapping.py
rickybassom: Which line?

JackEAllen: Change prefix_map to `actual = fetch_instrument_fullname_mapping()`

utils/clients/tools/tests/test_isisicat_prefix_mapping.py
JackEAllen: Change prefix_map to `actual = fetch_instrument_fullname_mapping()`

PR #674, opened 2020-07-22
Check local storage for graphs

### LINKED ISSUE #618: Check webapp cache for image before attempting to find
Issue raised by: [developer]

### What?
To increase efficiency wee want to check the local webapp cache to see if we have already downloaded the static image that we want before attempting to re-download it. To do this we should:
* Add a check in the plot handler to attempt to find an existing file in the local cache before attempting to re-download a file.
  * Add another function in plot handler that will be called on the first line of get_plot_file. This function will return the local address of the file if one exists. 

### Where?
* test code: `plotting/tests/plot_handler.py`
* source code: `plotting/plot_handler.py`

* Run a reduction job that produces a graph. (e.g. MARI 27282)
* Check the graph is found in the webapp successfully
* Check the webapp node and look in the `WebApp/autoreduce_webapp/static/graphs` dir.
  * In here (using example of MARI 27282) you should see `MARI/27282/MAR27282_Ei50.05meV.png`
* Change the `utils/credentials.ini` file so that the SFTP section is wrong
* Re-load the webapp and ensure that the image is still displayed. 
* REMEMBER to change credential file back afterwards!

keiranjprice101:
### Summary of work

Probably easiest to review commit by commit in order, instead of all at once.

Added a method to check the local storage for a graph before searching server.

I also added a method to generate the file extension part of the filename regex pattern. This moves 1 existing for loop and meant another wouldn't have to be introduced.

Test names maybe aren't great

### How to test your work
 
I tested this by loading a reduction that produced a graph (MARI 27282) Then I commented out the code to search remotely for the image and restarted the server and made sure the graph was then reloaded when I reopened that reduction.

### Additional comments
 

Fixes #618

### DIFF
--- a/plotting/plot_handler.py
+++ b/plotting/plot_handler.py
@@ -64,10 +64,7 @@ class PlotHandler:
         .png, .gif and .jpg: The returned value would be (png|gif|jpg)
         :return: (str) expression pattern matching the file extensions of the plot handler
         """
-        _extension_regex = '('
-        for extension in self.file_extensions:
-            _extension_regex += f'{extension}|'
-        return _extension_regex[:-1] + ')'
+        return f"({','.join(self.file_extensions).replace(',','|')})"
 
     def _get_plot_files_locally(self):
         """
--- a/plotting/plot_handler.py
+++ b/plotting/plot_handler.py
@@ -56,6 +56,17 @@ class PlotHandler:
             LOGGER.info("The instrument name is not recognised")
             return None
 
+    def _generate_file_extension_regex(self):
+        """
+        Generates the file extension part of the file regex. For example if the file extensions were
+        .png, .gif and .jpg: The returned value would be (png|gif|jpg)
+        :return: (str) expression pattern matching the file extensions of the plot handler
+        """
+        _extension_regex = '('
+        for extension in self.file_extensions:
+            _extension_regex += f'{extension}|'
+        return _extension_regex[:-1] + ')'
+
     def _check_for_plot_files(self):
         """
         Searches the server directory for existing plot files using the directory specified.
--- a/plotting/plot_handler.py
+++ b/plotting/plot_handler.py
@@ -104,6 +104,10 @@ class PlotHandler:
         but will only copy over one.
         :return: (str) local path to downloaded files OR None if no files found
         """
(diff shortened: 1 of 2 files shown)


plotting/plot_handler.py
```diff
@@ -42,20 +44,40 @@ def __init__(self, instrument_name, run_number, server_dir, rb_number=None):
...
+    def _generate_file_extension_regex(self):
+        """
+        Generates the file extension part of the file regex. For example if the file extensions were
+        .png, .gif and .jpg: The returned value would be (png|gif|jpg)
+        :return: (str) expression pattern matching the file extensions of the plot handler
+        """
+        _extension_regex = '('
+        for extension in self.file_extensions:
+            _extension_regex += f'{extension}|'
+        return _extension_regex[:-1] + ')'
```
JackEAllen, line 70: Could this not just be replace with something like?  
```suggestion
 return f"({','.join(self.file_extensions).replace(',', '|')})"
```
or:
```suggestion
 file_extensions_as_str = ','.join(self.file_extensions)
 file_extensions_for_regex = file_extensions_as_str.replace(',', '|')
 return f"({file_extensions_for_regex})"
```

plotting/tests/test_plot_handler.py
```diff
@@ -59,21 +61,82 @@ def test_generate_file_name_regex(self):
...
         When: Valid instrument names are supplied to _generate_file_name_regex
         """
-        expected_mari = "MAR(I)?1234.*.png"
-        actual = self.test_plot_handler._generate_file_name_regex('png')
-        self.assertEqual(expected_mari, actual)
-        expected_wish = "WISH1234.*.tiff"
+        actual = self.test_plot_handler._generate_file_name_regex()
+        self.assertEqual(self.expected_mari_file_regex, actual)
+        expected_wish = f"WISH1234.*.{self.expected_file_extension_regex}"
         self.test_plot_handler.instrument_name = "WISH"
```
JackEAllen, line 67: NIT: Could we change these to use the setup variables

plotting/plot_handler.py
keiranjprice101: Much cleaner, after 8 months of writing java I forget python actually has nice string tools 👍

PR #680, opened 2020-07-27
630 display reduction software in run summary

### LINKED ISSUE #630: Display software used in reduction on webapp
Issue raised by: [developer]

### What?
Now that the software table is being used and is populated #528 . We need to display the software name and version that was used on the reduction job summary page. `run_summary.html` This should just be added as another field in the meta information box below the title / plot 

### Where?
run_summary.html & `reduction_viewer.views.run_summary()`

* Load the webapp
* Navigate to any run
* See that the Software used for reduction is displayed

rickybassom:
### Summary of work
Added software name - version to run_summary.html below Retrying.

### How to test your work
* Code review
* Load the webapp and navigate to a run where there's an entry for that run in the software table.
* Then see that the software name - version is being displayed.
* Also navigate to a run without a software entry and see that "No software data found" is displayed.

Fixes #630 

### DIFF
--- a/WebApp/autoreduce_webapp/templates/run_summary.html
+++ b/WebApp/autoreduce_webapp/templates/run_summary.html
@@ -103,9 +103,11 @@
                                     {% endif %}
                                 </div>
                                 <div>
+                                    <strong>Software used:</strong>
                                     {% if run.software is not null %}
-                                        <strong>Software used:</strong>
                                         {{ run.software.name }} - {{ run.software.version }}
+                                    {% else %}
+                                        <em>No software data found</em>
                                     {% endif %}
                                 </div>
                             </div>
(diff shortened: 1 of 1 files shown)


WebApp/autoreduce_webapp/templates/run_summary.html
```diff
@@ -102,6 +102,14 @@ <h2>Reduction Job #{{ run.title }}</h2>
                                         <em>Not yet finished</em>
                                     {% endif %}
                                 </div>
+                                <div>
+                                    <strong>Software used:</strong>
+                                    {% if run.software is not null %}
```
JackEAllen, line 107: NITPICK/Suggestion: This will still work if we alter this slightly to  `{% if run.software is not null %}` which looks a little cleaner. This is only a personal preference so no need to change if you don't agree :)

rickybassom: I believe I have written it as `{% if run.software is not null %}`. This follows the style of other conditions in the file. Do you mean something else like `{% if run.software %}`?

JackEAllen: sorry, yes i do XD i just blatantly copied if without actually changing it - my bad

PR #691, opened 2020-08-03
Constrain status model

### LINKED ISSUE #608: Constrain Status model
Issue raised by: [developer]

### What?
Our Model (database) is defined using the Django ORM (Object relational mapping - allows for database records to be treated as python objects). The Status table is used to denote what the status of a particular job is, e.g. `Queued`, `Processing`, `Skipped`, `Complete`, `Error`

The 5 values above are currently the only valid options for Status. As such, we should look to constrain the database to ensure that it is not possible to add any other values by mistake. 

Take a look at this [SO post](https://stackoverflow.com/questions/48040008/django-restrict-data-that-can-be-given-to-model-field) that in the top answer discusses `choices`. We should aim to use this for our Status table.

In addition to developing the constraint, we need to test that it works. Currently the model is not tested so we will add a single test class for this in this issue. 

Consider attempting to make this change with Test Driven Development (TDD) e.g. :
1. Add a new test file to `WebApp/autoreduce_webapp/reduction_viewer/tests/test_models.py`
2. Write a test that creates a `Status` Model and adds a non-valid option e.g. `"test"` as the `status_value`.
3. To ensure the acceptance case as well, and a test that checks a valid status_value in the Status record passes e.g. `Complete`.
4. Running these tests BEFORE making code changes should show that both pass.
5. Running these tests AFTER code changes should show the test from (2) fails and (3) passes

### Where?
source file: `WebApp/autoreduce_webapp/reduction_viewer/models.py`
test file: `WebApp/autoreduce_webapp/reduction_viewer/tests/test_models.py` (needs to be created)

* Unit tests are acceptable: code review 
* Unit tests pass on travis

rickybassom:
### Summary of work
Constrained the Status model to 5 choices

### How to test your work
- Unit tests are acceptable: code review
- Unit tests pass on travis

Fixes #608


queue_processors/queue_processor/handle_message.py
```diff
@@ -181,7 +181,7 @@ def reduction_started(self, message: Message):
                 rb_number=message.rb_number, run_number=message.run_number,
                 run_version=message.run_version)
 
-        if reduction_run.status.value not in ("Error", "Queued"):
+        if reduction_run.status.value not in ['e', 'q']:
```
ElliotAOram, line 184: Would prefer not to use short hand in code base it's less obvious what is going on.

rickybassom: I tried this with `.verbose_value()` but it didn't work because it's being mocked I think. I could add comments explaining the verbose values in these instances?

PR #695, opened 2020-08-04
Refactor of send reduction message to AMQ and message determination o…

### LINKED ISSUE #639: Encase Updating of Reduction Status Messages and Log Information in Sub-method
Issue raised by: developer

### What?
Encase processes which update message and log info to state if an error has occurred somewhere during reduction or if the reduction completed successfully into a sub-method. 

I recommend calling this method  `reduction_return_status`

### Where?
* Lines to place in method: 341-360

### How?

## Additional Notes

* Code review
* Ensure System tests pass

JackEAllen:
### Summary of work
Encase processes which update message and log info to state if an error has occurred somewhere during reduction or if the reduction completed successfully into a sub-method.

Refactor method which updates AMQ message

### How to test your work
• Code review
* Ensure all unit tests for post_process_admin pass
* No drop in coverage

Fixes #639

### DIFF
--- a/queue_processors/autoreduction_processor/post_process_admin.py
+++ b/queue_processors/autoreduction_processor/post_process_admin.py
@@ -189,11 +189,10 @@ class PostProcessAdmin:
                          amq_message,
                          self.message.serialize(limit_reduction_script=True))
             self.client.send(amq_message, self.message)
-            logger.info("Reduction: %s" % message)
+            logger.info("Reduction: %s", message)
 
-        except AttributeError as exp:
+        except AttributeError:
             logger.debug("Failed to find send reduction message: %s", amq_message)
-            return exp
 
     def determine_reduction_status(self):
         """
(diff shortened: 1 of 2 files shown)


queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -179,12 +179,42 @@ def _load_reduction_script(self, instrument_name):
...
+            logger.info("Reduction: %s", message)
+
+        except AttributeError:
+            logger.debug("Failed to find send reduction message: %s", amq_message)
+
+    def determine_reduction_status(self):
+        """
+        Determine which message type to log and send to AMQ, triggering exception if job failed
+        """
+        if self.message.message is not None:
```
franzlang, line 201: could we rename `self.message.message` to something more meaningful/obvious?

JackEAllen: self.message is the Message class.
self.message.message refers to the message attribute within the message class for an AMQ message located here `model/message/message.py`.  Changing this to be something more meaningful/obvious should be its own issue to be handled separately. This is a good point to raise as the name is't completely obvious.

PR #702, opened 2020-08-10
Comment out search bar in help page

### LINKED ISSUE #701: Disable webapp help page search
Issue raised by: [developer]

### What?
As a temporary solution to the webapp search filter being poor, we should remove the search functionality. This will be fixed as part of #699 

### Where?
Webapp help page

* Ensure that the search bar on the webapp help page is not accessible on page load

ElliotAOram:
### Summary of work
* Comment out search functionality in help pages

### How to test your work
* Open webapp local and ensure the help page does NOT have a search bar

Fixes #701

### DIFF
--- a/WebApp/autoreduce_webapp/templates/help.html
+++ b/WebApp/autoreduce_webapp/templates/help.html
@@ -9,7 +9,7 @@
             <p>Get in touch with the <a href="mailto:isisreduce@stfc.ac.uk">Support Team</a></p>
         </div>
     </div>
-    <div class="row">
+    <!--<div class="row">
         <div class="col-md-8">
             <div class="form-group">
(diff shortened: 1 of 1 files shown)


WebApp/autoreduce_webapp/templates/help.html
```diff
@@ -35,7 +35,7 @@ <h2>Help Pages</h2>
                 </div>
             </div>
         </div>
-    </div>
+    </div>-->
```
JackEAllen, line 38: Are we removing the filter buttons Ricky just put in too?  
They help filter sections by category quite effectively.

ElliotAOram: I don't think the filter buttons work without search bar due to what the `JS` is running on. I can double check this before merging

PR #708, opened 2020-08-10
`python setup.py start` does not inform user if not started properly

### LINKED ISSUE #707: python setup.py start does not inform user if not started properly
Issue raised by: [developer]

### What?
E..g. run

```
python setup.py start`
```

on ubuntu with user that do not have full permission access to /opt/autoreduce_deps, and the output is

```
running start
```

even though in this case activemq have failed to start up properly.

Provide better message to developer installing and using above setup.py start

### Reproducible?
Yes, and see above

Test the above scenario where

* do not have permission to /opt/autoreduce_dev  , and you should see error message 
* do have permission and you should see `running start`

Anders-Markvardsen:
### Summary of work
See #707 

### How to test your work
Run 

```
python setup.py start
```

from a user that does not have access to /opt/autoreudce_devs, which result in error message. (unfortunately the pid will still hang around, which you have to kill manually; in script I tried popen.kill() and popen.terminate() but neither had an effect, but at least now the message gives a helping hand).

Then do the same but using sudo. On my OS this is:

```
sudo /home/ajm64/autoreduce/venv3/bin/python setup.py start
```

Fixes #707


build/utils/process_runner.py
```diff
@@ -23,11 +23,17 @@ def run_process_and_log(list_of_args):
...
+    # not have the right permission with activemq
+    if exit_status != 0 or 'Permission denied' in process_error.decode("utf-8"):
+        # log both process_output and process_error, since process_output
+        # may also info that help with diagnose the problem
         BUILD_LOGGER.logger.info(process_output)
-    if exit_status != 0:
         BUILD_LOGGER.logger.error(process_error)
         return False
+    if process_output:
+        BUILD_LOGGER.logger.info(process_output)
```
JackEAllen, line 36: Could we move this into a seperate method so we can make the methods more concise and easier to debug if problems ever arise given this setup is untested. 
e.g.

```python
def log_process(process_output, process_error, exit_status):
    """ 
    Typically, an exit status of 0 indicates that the process ran
    successfully. However, this is for example found not to be true when
    not have the right permission with activemq
    :param process_output: (sub-process) <Param Description>
    :param process_error: (sub-process) <Param Description>
    :param exit_status: (sub-process) <Param Description>
    :return: (bool) True or None
    """

    if exit_status != 0 or 'Permission denied' in process_error.decode("utf-8"):
        # log both process_output and process_error, since process_output
        # may also info that help with diagnose the problem
        BUILD_LOGGER.logger.info(process_output)
        BUILD_LOGGER.logger.error(process_error)
        return False
    if process_output:
        BUILD_LOGGER.logger.info(process_output)
    return True


def run_process_and_log(list_of_args):
    """
    Call a process using Popen and logs output to file
    :param list_of_args: list of arguments for Popen
    :return: (bool) True or None
    """
    process = subprocess.Popen(list_of_args,
                               stdout=subprocess.PIPE,
                               stderr=subprocess.PIPE)
    process_output, process_error = process.communicate()
    exit_status = process.returncode

    return log_process(process_output, process_error, exit_status)


def run_process_with_shell(list_of_args):
    """
    Run a process using the shell true command and do not log the output
    :param list_of_args: list of args for Popen
    :return:
    """
    subprocess.call(list_of_args, shell=True)
```

JackEAllen (approved):
Looks good! 👍  

I only have a minor suggestion to break one of the methods up a little. This is more a personal  preference on my behalf as i believe it looks cleaner. It could be easier to debug should problems arise here as we can easily distinguish if the problem is related to the creation of the process or the logic determining how to log the process. 

As this is only a minor suggestion i'm approving this PR so you can choose whether or not to make my suggested change.

build/utils/process_runner.py
Anders-Markvardsen: Thanks @JackEAllen for suggestion, I have added time hint to return value, but because let as one method for now and your suggestion for splitting can be added when we get testing of deployment setup

Anders-Markvardsen:
@rickybassom could you double check this PR and if OK merge, also related to setting up local a to z environment

rickybassom:
Here's what I've found:
- When doing a fresh clone and going through the setup tutorial on Fedora 31, I discovered that the [Install external programs](https://github.com/ISISScientificComputing/autoreduce/wiki/Setup#install-external-programs) section only works on Debain/Ubuntu based Linux distributions since `apt` is not available on Fedora/Centos etc... It should be mentioned in the wiki that the `python setup.py externals` command can only be executed on Debain/Ubuntu based Linux distributions. As a work around I installed activemq manually.

- I don't think there's anywhere in the wiki that mentions adding your mysql credentials to credentials.ini for development.

- A small section that tells the developer how to run the queue processor and WebApp straight after the Setup section could be added.

- Running `sudo python setup.py start` outputted "running start" and `./restart.sh` worked perfectly.

Anders-Markvardsen:
@rickybassom agree with your observation in the first 3 bullet points above and can I suggest you create a issue for this including with a-to-z local install wiki doc updates and with yum install also. With your observation in the last bullet point which also verify this pr, and hence please could you go ahead and merge this pr

PR #712, opened 2020-08-11
Fix qp issues

### LINKED ISSUE #614: Fix running for string RB numbers
Issue raised by: [developer]

### What?
When getting the RB number from ICAT we can sometimes get a string instead something like: "CAL_GEM_12/03/2020 10:15:08". In this case we need NOT to perform the reduction (because it is a calibration run) and add the run to the skipped queue to ensure it is still added to the database. 

There are 2 things that we could do here:
1. Check if rb is string and then send to skipped (fix in queue processor)
2. See if it possible to get the ACTUAL RB number from ICAT (fix in manual submission)

Preference here is to go for option 2 as it is more consistent with the way the code base currently works. 

### Where?
Either queue_processor or manual_submission depending on implementation option 

### How?
* Un-noticed side affect of Message class implementation

Attempt to submit GEM90375 on the development environment via the manual submission script. At present this run is not in the autoreduction database so will be taken from ICAT instead.

### LINKED ISSUE #673: Consumers disconnect
Issue raised by: [developer]

### What?
At present the consumers for the queue processor disconnect when `MARI27252 ` is submitted. There is no logging as to why this happens and it is not obvious what in particular about this run is causing the issue. The issue needs to be tracked down and resolved as a matter of priority. 

### Where?
Queue Processsors

### How?
seen on dev and production nodes

### Reproducible?
Yes - manually resubmit `MARI27252` and check hawtio to see QP consumers disconnect

Manually resubmitting this run should see it complete and the consumers remain active.

ElliotAOram:
### Summary of work
* This fixes this issue with invalid runs breaking the consumers and forcing them to disconnect
* Enables str RB's to be properly processed

### How to test your work
* Manually submit a str RB e.g. `MARI 27252` - This should be skipped and assigned to RB0
* Manually submit a normal RB e.g. `MARI 27037` - This should complete with no issue
* Manually submit a RB 0 e.g. `MARI 26920` - This should be skipped and assigned to RB0

Fixes #673 & Fixes #614 

### DIFF
--- a/model/message/validation/process.py
+++ b/model/message/validation/process.py
@@ -35,7 +35,7 @@ def dict_to_string(validity_dict):
     :param validity_dict: (dict) containing the validity checks
     :return: (str) A string to print
     """
-    result_str = ""
+    result_str = "\n"
     for key, value in validity_dict.items():
-        result_str += f"{key} - {value}\n"
+        result_str += f"\t{key} - {value}\n"
     return result_str
--- a/model/message/validation/tests/test_process.py
+++ b/model/message/validation/tests/test_process.py
@@ -56,5 +56,5 @@ class TestProcess(unittest.TestCase):
         Test: An expected string is returned
         When: dict_to_string is called with a valid dict
         """
-        expected = "check_1 - True\ncheck_2 - True\n"
+        expected = "\n\tcheck_1 - True\n\tcheck_2 - True\n"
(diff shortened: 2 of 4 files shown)


ElliotAOram:
This is not done do not merge / review

ElliotAOram:
This is now ready for review

PR #715, opened 2020-08-13
Fixes to allow successful execution of tests on Ubuntu 18.04

dtasev:
### Summary of work
- Uses the DB configuration from the generated `credentials.ini`
- Moves the credential generation to happen before the DB config tries to read the generated file
- Removes database-generating commands that require ROOT privileges 
  - These seemed to just be duplicating the `mysql < build/database/reset_autoreduction_db.sql` command.
  - Locally, they were failing as MySQL does not allow using the `root` user without `sudo`. And I could not find a way to force MySQL to allow that

### How to test your work
- Tests should pass
- Check that the following commands execute OK locally 
  - python3 setup.py test_settings
  - python3 setup.py database 
- Run all tests

### Additional comments
I also had to manually run `activemq` on Ubuntu, not sure how that is passing on Travis as I don't see it being manually started from the `.travil.yml`

What worries me is that this PR pretty much undoes some of the work added in these commits:
- 082598b3a9e8cf8ef7a184b14ad6f0f36453e071 (just a rename/re-order)
- 38146a03b4a887ae58fcf293099c342e7d47c0d7 adds set-up user

but I can't see a reason _now_ that we would want to reset the whole database twice before even adding the django tables

No related issue

### DIFF
--- a/build/commands/database.py
+++ b/build/commands/database.py
@@ -11,8 +11,7 @@ import os
 # pylint:disable=no-name-in-module,import-error
 from distutils.core import Command
 
-from build.database.generate_database import (get_sql_from_file, get_test_user_sql,
-                                              run_sql, generate_schema)
+from build.database.generate_database import (get_sql_from_file, run_sql, generate_schema)
 from build.utils.common import BUILD_LOGGER, ROOT_DIR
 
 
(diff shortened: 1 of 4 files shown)


Anders-Markvardsen:
Tested workflow on my WSL v2 ubuntu setup

```
python3 setup.py test_settings
python3 setup.py database
```

which works

Anders-Markvardsen:
Concerning you activemq question then to start run 

```
python setup.py start
```

and see line 30 in .travis.yml

perhaps could named clearer. Elliot told me that at some point used to start both activemq and queue_processor

PR #721, opened 2020-08-18
Cache plots

### LINKED ISSUE #616: Ensure cached graph images are removed after time

### What?
What is the issue

### Where?

### How?
How did the issue come about

### Reproducible?
[Yes/No]

keiranjprice101:
### Summary of work

Creates a custom django cache backend that is used for storing plots.
The cache extends the filebased cache and stores value with key, value pairs being plot regex string: local plot path
When `is_expired` is called the cache will also delete the plot file locally, the same way the normal cache would delete the key.

I have changed `PlotHandler` to generate the regex pattern inside the `__init__` since it is used in most places and a new `PlotHandler` is created on every access to the `run_summary` view anyway.

Currently the cache expires once per day (Elliots suggestion) but this can be increased or decreased dependant on what we need.

### How to test your work
 
Run a reduction that gets a graph. Reload the page and verify that the plot_cache has been accessed and not ceph.
Make sure plot is deleted on expiry

### Additional comments
 

Fixes #616

### DIFF
--- a/plotting/tests/test_plot_cache.py
+++ b/plotting/tests/test_plot_cache.py
@@ -4,13 +4,12 @@
 #  Copyright &copy; 2020 ISIS Rutherford Appleton Laboratory UKRI
 #  SPDX - License - Identifier: GPL-3.0-or-later
 # ############################################################################### #
-
 """
 Tests for the PlotCache class
 """
 from unittest import TestCase
 
-from mock import patch, MagicMock, call
+from unittest.mock import patch, MagicMock, call
 
 from plotting.plot_cache import PlotCache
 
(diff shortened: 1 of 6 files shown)


dtasev:
Rebased on master

plotting/tests/test_plot_cache.py
```diff
@@ -0,0 +1,106 @@
...
+    """Tests the functionality of the PlotCache"""
+    PLOT_PATH = "path"
+
+    def setUp(self):
+        """
+        Setup the plot cache and the opened cache file mock
+        """
+        dir_ = MagicMock()
+        params = MagicMock()
+        self.cache = PlotCache(dir_, params)
```
dtasev, line 28: This line now seems to be throwing `TypeError: expected str, bytes or os.PathLike object, not MagicMock` but I'm not sure what I changed in the rebase that broke it!

dtasev:
This PR seems to be pretty much ready and nicely self contained. I'll have a go testing it

PR #733, opened 2020-08-26
Contrast fix

### LINKED ISSUE #663: Text on auto reduction home does not meet WCAG guidelines
Issue raised by: [developer]

### What?
The footer text does not meet [WCAG](https://www.w3.org/TR/WCAG21/#contrast-enhanced) contrast guidelines. Currently the ratio is 1.6:1

image

This may be true for other text/logos in the webapp but I did not check.

### Where?
Being unfamiliar with the webapp, I don't know where to look to change this. @ElliotAOram I guess you would be able to point in the right direction.

keiranjprice101:
### Summary of work
Changes font colour of header welcome and footer navbar text to be in line with WCAG 

### How to test your work
 
Load home page, ctrl+f5. Check colours are within WCAG contrast

### Additional comments
 

Fixes #663 

### DIFF
--- a/WebApp/autoreduce_webapp/static/css/base.css
+++ b/WebApp/autoreduce_webapp/static/css/base.css
@@ -60,6 +60,9 @@ hr {
     padding: 10px;
 }
 
+.navbar-default .navbar-text {
+    color: #444444;
+}
 /****************
  3. Header
 ****************/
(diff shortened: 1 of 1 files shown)


JackEAllen (approved):
I'm happy with these changes. 👍  
Once we move everything to scss #735 changes like this will be even easier to make and manage.

PR #776, opened 2020-09-17
NULL autoreduction script/step for copy raw data for Muon instruments

### LINKED ISSUE #725: NULL autoreduction script/step for copy raw data for Muon instruments
Issue raised by: [user, Steve Cottrell]

### What?
Have NULL autoreduction script/step for EMU, MuSR, HIFI, ARGUS and CHRONUS which simply just copy raw file to reduction folder (RB number).

Temporary fix to make such files more readily available to CEPH Filebrowser tool.

### Where?
I don't think this actually need any change to the Autoreduction code base, but will need changes:

* Create dummy reduction scripts which simply just save raw file, and add these to relevant instrument user areas
Needs to enable
* On the production environment, simply enable monitoring component to these instruments

Anders-Markvardsen:
### Summary of work
See #725 . NULL or "Copy raw data" scripts added for Muon instruments, see e.g. (needs to be within ISIS firewall) \\isis\inst$\NDXMUSR\user\scripts\autoreduction . In addition Max Skoda also requested this, and this script added for POLREF, SURF and INTER instruments also.

Replaced out-of-date example_reduce.py scripts, with a copy of this null script.

### How to test your work
* Test that copy raw data script works, e.g. by submitting a HIFI, MUSR or POLREF job to the dev environment
* Check the comments in the code are they easy to understand, and the example code
 
Fixes #725

### DIFF
--- a/WebApp/example_reduce.py
+++ b/WebApp/example_reduce.py
@@ -10,18 +10,21 @@ import shutil
 import sys
 import reduce_vars as web_var
 
-def validate(input_file, output_dir):
+def validate(file, dir):
     """
-    Function to ensure that the files we want to use in reduction exist.
-    Please add any files/directories to the required_files/dirs lists.
+    Function that validates if a file and/or directory exist. If not a
+    RunTimeError is raised which is picked up by Autoreduction.
+
+    :param file: full path of data file. Provide empty string to ignore
+    :type file: str
+    :param dir: full path of a directory. Provide empty string to ignore
+    :type dir: str
     """
     print("Running validation")
-    required_files = [input_file]
-    required_dirs = [output_dir]
-    for file_path in required_files:
-        if not os.path.isfile(file_path):
-            raise RuntimeError("Unable to find file: {}".format(file_path))
-    for dir in required_dirs:
+    if file:
+        if not os.path.isfile(file):
+            raise RuntimeError("Unable to find file: {}".format(file))
+    if dir:
         if not os.path.isdir(dir):
             raise RuntimeError("Unable to find directory: {}".format(dir))
     print("Validation successful")
--- a/WebApp/example_reduce.py
+++ b/WebApp/example_reduce.py
@@ -5,32 +5,52 @@
 # SPDX - License - Identifier: GPL-3.0-or-later
 # ############################################################################### #
 # pylint: skip-file
-import sys
+import os
 import shutil
+import sys
 import reduce_vars as web_var
 
+def validate(input_file, output_dir):
+    """
+    Function to ensure that the files we want to use in reduction exist.
+    Please add any files/directories to the required_files/dirs lists.
+    """
+    print("Running validation")
+    required_files = [input_file]
+    required_dirs = [output_dir]
+    for file_path in required_files:
+        if not os.path.isfile(file_path):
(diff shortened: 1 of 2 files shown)


WebApp/example_reduce.py
```diff
@@ -5,32 +5,52 @@
...
 import reduce_vars as web_var
 
+def validate(input_file, output_dir):
+    """
+    Function to ensure that the files we want to use in reduction exist.
+    Please add any files/directories to the required_files/dirs lists.
+    """
+    print("Running validation")
+    required_files = [input_file]
+    required_dirs = [output_dir]
```
JEThompson97, line 20: I notice the names of this functions arguments imply a single paths (i.e. 1 input file and 1 output dir).
Here you put them in an array. Is the idea here to allow the function to take an array of input files/output dirs?

If so, would this code would if validate was called with an array arguments? From looking at it, it seems like these lines would make it a 2D array, example:

```python
validate([file1, file2], dir)
...
required_files = [[file1, file2]]
```
If I have misunderstood the intention, please correct me

JEThompson97 (approved):
Looks good overall 👍
Just curious about the conversion to arrays

WebApp/example_reduce.py
Anders-Markvardsen: This is a brilliant observation, I didn't actually look at the code of validate and just the signature of main() within reduce.py examples we have, which to me from names input_file and output_dir implied 1 input file and 1 output dir. Looking at the code in `queue_processors/autoreduction_processor/post_process_admin.py` it is not crystal clear what the behaviour should be, for example the place where reduce.py is called is (in a method with named `check_for_skipped_runs` which name no longer match its content entirely)

```
                out_directories = reduce_script.main(input_file=str(self.data_file),
                                                     output_dir=str(reduce_result_dir))
```

which to me also indicate 1 input file and 1 output dir. However, following this upwards in code the picture is more muddled as to what the intention is. 

@JEThompson97 are you aware of a use case we currently have (or did @ElliotAOram mention) where we currently call reduce.py script with more than one input file and output directory? Or if there was plans for this, e.g. to accommodate reflectometry instruments or other?

JEThompson97: Ah, I didn't realise this was pre-existing code! That's interesting..
I'm not aware of any case where we call the script with more than 1 file/directory. From looking at the code, I don't think this would work as the method doesn't appear to actually _handle_ `input_file` being an array (nor `output_dir`), rather it just converts them to arrays within the method (see the code snippet example in my comment above)...

I'm not aware of any plans to allow this to handle multiple files; however the docstring comment at the top of `verify` suggests this was intentional to some extent (Line 15-16).

None the less, the code should still work here. If nothing else, we should keep the way this is implemented in mind. But I wouldn't have said straightening it out was a top priority.

JEThompson97 (approved):

Anders-Markvardsen:
Thanks @JEThompson97 and please don't hesitate to merge. Note FI tested updated script and that it works with HRPD on dev environment, and will add later a issue for Oct sprint and as reminder to between cycle update reduce.py scripts accordingly on instruments which are in production

Anders-Markvardsen:
Merge based on approved

PR #781, opened 2020-09-23
764 fix output structure

### LINKED ISSUE #764: Remove unacceptable user facing folder/file structure accidentally introduced

### What?
Observed during smoke testing before ISIS cycle 20_2.

This user facing change likely accidentally occurred during implementation of webapp plotting.

See picture 

unnamed

The new user offending part which has accidentally been introduced is that is a new directory is created for every run submitted to a RB number with the unhelpful user facing numbering 0, 1, 2, ... In this example the user will see in /instrument/MUSR/RBNumber/RB1920237/autoreduced:

```
0/
1/
```

whereas the correct and previously behaviour for this instrument (which is 'labelled' as an `excitation` instrument, name 'excitation' here separate issue to change this name in queue processors setting file):

```
files for reduction of run 80968
files for reduction of run 80969
reduction_log/  (which contains the logs from the reduction)
```

Note if instrument not labelled as 'excitaiton` instrument the output would have been (for the above example):

```
80968/
80969/
```
i.e. where extra sub-folders are run numbers. This is useful for those instruments producing masses of files for each run.

To reproduce:

1. Run any manual submission job, e.g. `python manual_submission.py MUSR 80968` and  `python manual_submission.py MUSR 80969` (note in picture above reduce.py is setup with dummy reduction script that simply copies the raw data file)

Purpose of this issue is the revert back to the previous way of outputting reduction files and logs, while accommodating the new webapp plotting.

### Where?
Probably needs both changes to queue processors code and webapp code

### Reproducible?
Yes

### LINKED ISSUE #765: test_create_log_path will always fail on windows
Issue raised by: [developer]

### What?
`test_create_log_path` will always fail on a windows machine

### Where?
`queue_processors/autoreduction_processor/tests/test_post_process_admin.py`

### How?
`PosixPath` cannot be instantiated on windows machines.

### Reproducible?
[Yes]

Tests should pass on all systems.

### LINKED ISSUE #767: Rename excitation_instruments in test_settings.py file
The name of the `excitation_instruments` in

https://github.com/ISISScientificComputing/autoreduce/blob/c62c1e3f26939fb36aa36455bcd415ca570d57dc/queue_processors/autoreduction_processor/test_settings.py#L27

historical was added to cater for a specific Excitation requirement early in the process and as a temporary solution (there is a saying 'there is nothing as permanent as a temporary solution:-)). Anyway in this issue change `excitation_instruments` to `flat_output_file_structure` or something similar` and as a minimum document it in the test_settings.py file. 

What it is does is explained in #764 , see 'excitation' in that issue. 

Short documentation of it could be "All instruments added to this list will have the reduced data outputted into a flat folder structure, meaning without 'run number' sub-folders"

keiranjprice101:
### Summary of work
 - Renamed the directory variables in `post_process_admin` so it is more clear that one is temporary and the other final
 - Removed a duplicated log path creation in `post_process_admin`
 - Flat outputs no longer have the number folders
 - Non flat outputs instead have `/run-version-0` and increment on each new version
 - Renamed excitation_instruments to flat_output_instruments
 - Fixes a test that fails only on windows

I have opened this related issue #801

****Note:**** There will be a discrepancy between the real version and the version folder created until the database is cleared and the output folders are cleared.

Previous logging:
image

New logging:
image

### How to test your work
Submit a run for a flat output instrument. Check that it is outputted without the numbered subfolder
Submit a run for a non flat output instrument. Check it is outputted to `run-version-x`
Submit another run and make sure it is incremented

### Additional comments
 

closes #767 closes #764 closes #765


queue_processors/autoreduction_processor/post_process_admin.py
```diff
@@ -487,29 +491,20 @@ def _get_mantid_version():
...
+    def _append_run_version(self, path):
+        """
+        Append the run version to the output directory. If its the first run run-version-0 will be
+        added, if overwrite is true run-version-0 will always be overwritten
+        :param path: The reduction output path
+        :return: The reduction output path with the run version appended
+        """
+        if self.message.overwrite:
+            return append_path(path, ["run-version-0"])
+        run_versions = [int(i.split("-")[-1]) for i in glob.glob(f"{path}/run-version-[0-9]*")]
```
JEThompson97, line 503: Nice refactor here 👍 glob definitely seems the way to go

queue_processors/autoreduction_processor/test_settings.py
```diff
@@ -24,5 +24,5 @@
     "post_process_directory": append_path(os.path.dirname(os.path.realpath(__file__)), ["post_process_admin.py"]),
     "ceph_directory": append_path(get_project_root(), ['reduced-data', '%s', 'RB%s', 'autoreduced', '%s']),
     "temp_root_directory": "/autoreducetmp",
-    "excitation_instruments": ["LET", "MARI", "MAPS", "MERLIN", "WISH", "GEM"]
+    "flat_output_instruments": ["LET", "MARI", "MAPS", "MERLIN", "WISH", "GEM"]
```
JEThompson97, line 27: I like this renaming, makes it clearer what the relevance of this instruments are for the code side of things

JEThompson97 (approved):
Nice work 👍

Anders-Markvardsen:
@keiranjprice101 perhaps a stupid question. Do you known why we two travis-ci builds for this PR? And where continuous-integration/travis-ci/push comes from?

keiranjprice101:
@Anders-Markvardsen I was testing the manual travis submission feature on a build I knew was already passing. This luckily identified the sentry issue.

Anders-Markvardsen (approved):
tested on dev with flat and not flat dir instruments, WISH and GEM respectively, and output as described and code clear

PR #791, opened 2020-09-30
Fix broken help navigation

### LINKED ISSUE #788: Help Navigation broken
Issue raised by: [developer]

### What?
The help navigation is broken See gif
help-vansih
This happens for me on chrome and with any of the side links

### Where?
`WebApp/autoreduce_webapp/templates/help.html` 

### How?
How did the issue come about

### Reproducible?
[Yes]

The navigation links should move the user to the correct section.

rickybassom:
### Summary of work
* Commented out a bit of JavaScript in help.js which was causing #788 
* Added comments pointing to relevant issue https://github.com/ISISScientificComputing/autoreduce/issues/701 where the search bar code has been disabled

### How to test your work
* Code review
* All tests pass
* Go to the help page and test if the navigation sidebar works

Fixes #788

### DIFF
--- a/WebApp/autoreduce_webapp/static/javascript/help.js
+++ b/WebApp/autoreduce_webapp/static/javascript/help.js
@@ -93,12 +93,13 @@
         });
 
         // Reset search filters on clicking sidebar link
-        $('#sidenav').click(function () {
-            $('[data-category="all"]').click();
-            $('#help-search').val("");
-            searchFilter(stringToSearchTerms($(this).val()),
-                $('#category-filter > .btn.active').data("category"));
-        });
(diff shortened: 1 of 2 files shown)


keiranjprice101:
Working for me. Is there a reason to comment out the JS instead of just deleting it, since we never lose anything because of git anyway?

keiranjprice101 (approved):

rickybassom:
> Working for me. Is there a reason to comment out the JS instead of just deleting it, since we never lose anything because of git anyway?

Good point. Surrounding bits of code reference a search feature and so it might be confusing if that bit of code was removed completely.

keiranjprice101:
> Surrounding bits of code reference a search feature

Fair enough. Looks good for me and working

PR #793, opened 2020-10-05
Fixes runs not displaying due to any error in plot retrieving code

### LINKED ISSUE #760: Run not found when run exists
Issue raised by: [developer]

Found in #750

### What?
For some instruments,(e.g. GEM or HRPD), clicking on a run that exists results in `Reduction run not found` instead of seeing the result from the run!

To reproduce:

* If it does not already exist on the WebApp submit `python manual_submission.py HRPD 78938` where HRPD at this point is setup to run the simple copy-raw-file script and therefore shouldn't be failing on any circumstance
* Then in WebApp click on HRPD and then 78938

Which result in the WebApp 'Reduction run not found'.

### Where?
Either a bug in the WebApp code or that the database get populated in such a why that mismatch happens. To view content of database use e.g. MySQL Workbench

### Reproducible?
Yes

The above steps can be repeated successfully and with result from run displayed. Furthermore repeat the above steps but purposes, during cycle 20_2, purposely insert error in HRPD reduce.py script (HRPD is not enabled in production for cycle 20_2). Test that for this case also result of run are displayed. Selenium tested of webapp is still not enabled, but meanwhile it may possible to test aspect of this via e.g. systemtests/test_end_to_end.py - not sure

dtasev:
### Summary of work
The issue was caused by the rendering code falling over when trying to retrieve plot images for:
- Instruments that we don't support plotting at all
- Instruments that have no cached plots

Now renders the page regardless of presence of plot images. If plots are supported for the instrument (seems like only MARI and WISH are according to the code) but not found due to some issue (e.g. locally it was `Authentication failed` because I don't have the correct mount), a warning is displayed:
image

### How to test your work
- Run web-app locally or pull onto dev env. 
- Go into any run on any instrument.

Need to test whether plots still render properly! I couldn't find a run with plots, nor a way to generate them.

### Additional comments
The first commit is PyCharm enforcing it's formatting on the changed files. I'm not sure we have a strict formatter (e.g. YAPF) so can't avoid this for now.

Fixes #760 

~~**Before merging ensure the release notes have been updated**~~
-- Not sure this needs release notes

### DIFF
--- a/WebApp/autoreduce_webapp/reduction_viewer/views.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/views.py
@@ -28,13 +28,11 @@ from autoreduce_webapp.settings import UOWS_LOGIN_URL, USER_ACCESS_CHECKS, DEVEL
 from autoreduce_webapp.uows_client import UOWSClient
 from autoreduce_webapp.view_utils import (login_and_uows_valid, render_with,
                                           require_admin, check_permissions)
+from plotting.plot_handler import PlotHandler
 from reduction_variables.utils import MessagingUtils
 from reduction_viewer.models import Experiment, ReductionRun, Instrument, Status
 from reduction_viewer.utils import StatusUtils, ReductionRunUtils
 from reduction_viewer.view_utils import deactivate_invalid_instruments
-
-from plotting.plot_handler import PlotHandler
-
 from utilities.pagination import CustomPaginator
 
 LOGGER = logging.getLogger('app')
--- a/WebApp/autoreduce_webapp/reduction_viewer/views.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/views.py
@@ -275,7 +275,15 @@ def run_summary(_, instrument_name=None, run_number=None, run_version=0):
                               'reduction_location': reduction_location,
                               'started_by': started_by}
 
-        if reduction_location:
+    except PermissionDenied:
+        raise
+    except Exception as exception:
+        # Error that we cannot recover from - something wrong with instrument, run, or experiment
+        LOGGER.error(exception)
+        return {"message": str(exception)}
+
+    if reduction_location:
+        try:
             plot_handler = PlotHandler(instrument_name=run.instrument.name,
                                        rb_number=rb_number,
                                        run_number=run.run_number,
@@ -283,12 +291,12 @@ def run_summary(_, instrument_name=None, run_number=None, run_version=0):
             plot_locations = plot_handler.get_plot_file()
             if plot_locations:
                 context_dictionary['plot_locations'] = plot_locations
-
-    except PermissionDenied:
-        raise
-    except Exception as exception:
-        LOGGER.error(exception)
-        context_dictionary = {}
+        except Exception as exception:
+            # Lack of plot images is recoverable - we shouldn't stop the whole page rendering
+            # if something is wrong with the plot images - but display an error message
+            err_msg = "Encountered error while retrieving plots for this run"
+            LOGGER.error(f"{err_msg}. Run {run}\n. Error: {exception}")
+            context_dictionary["plot_error_message"] = f"{err_msg}. Error: {str(exception)}"
 
     return context_dictionary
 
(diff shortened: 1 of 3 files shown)


WebApp/autoreduce_webapp/reduction_viewer/views.py
```diff
@@ -277,20 +275,28 @@ def run_summary(_, instrument_name=None, run_number=None, run_version=0):
...
-        raise
-    except Exception as exception:
-        LOGGER.error(exception)
-        context_dictionary = {}
+        except Exception as exception:
+            # Lack of plot images is recoverable - we shouldn't stop the whole page rendering
+            # if something is wrong with the plot images - but display an error message
+            err_msg = "Encountered error while retrieving plots for this run"
+            LOGGER.error(f"{err_msg}. Run {run}\n. Error: {exception}")
+            context_dictionary["plot_error_message"] = f"{err_msg}. Error: {str(exception)}"
```
rickybassom, line 299: Just putting this here for discussion. Should we be displaying technical error messages to users, or is it enough to display a user style error message but also log the technical details separately? In other words would a error relating to the code be useful for a user to see?

rickybassom (approved):
Functions well and the code looks good!

WebApp/autoreduce_webapp/reduction_viewer/views.py
keiranjprice101: Good point. I think in this case its fine but is worth looking at in future. I don't know if Django (my lack of Django knowledge) has something similar to custom error pages in .net that can be toggled on or off (on for prod)

dtasev: I agree that printing the exception itself might be a bad idea - if something goes very wrong it could just fill up the whole page. I'll remove it from the user-facing error message

JEThompson97: I agree too. On top of the UX reason you've mentioned there, I've heard it's bad practice security wise to let a user know they encountered an authentication issue because in a way it's advertises _"restricted data here - don't look!"_ 😅

WebApp/autoreduce_webapp/templates/run_summary.html
```diff
@@ -13,29 +13,40 @@
             <div class="col-md-12 text-center">
                 <h2>Reduction Job #{{ run.title }}</h2>
                 {% if history.count > 1 %}
-                    (This run has been re-reduced, <a href="#" class="js-reduction-run-history" id="run_history">see history</a>)
+                    <p>
+                        (This run has been re-reduced,
+                        <a href="#" class="js-reduction-run-history" id="run_history">see history</a>)
```
JEThompson97, line 18: Does this still produce a space between `re-reduced,` and `see history`? Or will it appear as:
_(This run has been re-reduced,see history)_

JEThompson97 (approved):
Just a minor string formatting query from me
(+ the already mentioned not showing of error messages to users)

Anders-Markvardsen:
@dtasev concerning your question above "Need to test whether plots still render properly! I couldn't find a run with plots, nor a way to generate them.". Yes, as you know at the moment this is not as straightforward to test than it should be currently, and, in particular since I had to manually disable plotting on MARI for this cycle on production as this new feature was not production ready.

To test this I suggestion you modify the reduce.py script currently on HRPD. This instrument is not currently run on production, and it is simply just a copy of the example script: https://github.com/ISISScientificComputing/autoreduce/blob/master/WebApp/example_reduce.py . Could you look to modify this script with the minimum amount of code to demonstrate and to get some plot to show up. To help with are you @JEThompson97 or @keiranjprice101 aware of documentation we which explain how to enable plotting on an instrument?

WebApp/autoreduce_webapp/templates/run_summary.html
dtasev: Yup, still there 👍

dtasev:
I have made a run with a plot (HRPD 78938 - 30) and the page does load, but the plot isn't shown (nor an error message). This seems to be due to [this dictionary](https://github.com/ISISScientificComputing/autoreduce/blob/master/plotting/plot_handler.py#L23) used to construct a regex string to find the image. Only MARI and WISH are present. I named the plot file following `HRPD<RUN_NUMBER>.png` so just adding a `"HRPD":"HRPD"` fixes the plot loading (I tested this on the dev env)

The plot issue isn't part of this PR so we may look at it into #816

keiranjprice101 (approved):

Anders-Markvardsen (approved):
works as described and also thanks for created follow up issue on plotting with non WISH and MARI

Anders-Markvardsen:
@dtasev concerning your observation regarding "Before merging ensure the release notes have been updated" in template issue descriptions:-) Yes, that is currently an aspiration, but we could start phasing this  when we talk about dev beta testing monday

PR #796, opened 2020-10-06
Adds config for the `root` logger

### LINKED ISSUE #795: Logging doesn't always Log when expected
Issue raised by: developer

I apologise for the terrible title but struggled to write a better one

### What?
The [queue_processors/queue_processor/settings.py::LOGGING](https://github.com/ISISScientificComputing/autoreduce/blob/master/queue_processors/queue_processor/test_settings.py#L49-L60) dictionary is used in multiple places to configure a logger via calling `logging.config.dictConfig(LOGGING)`.

The issue is that configuring the logging that way allows only two loggers to work properly, namely the explicit ones in the LOGGING dict:

```
'loggers': {
    'queue_processor': {
        'handlers': ['file'],
        'propagate': True,
        'level': LOG_LEVEL,
    },
    'app': {
        'handlers': ['file'],
        'propagate': True,
        'level': 'DEBUG',
    },
}
```

This means that anywhere that does `logging.getLogger(...)` and is not either `queue_processor` or `app` gets a logger that has no handler - its output is completely discarded, e.g. `logging.getLogger("queue_listener")`.

### To fix
The dict needs to configure the root logger to also use the file handler:
```
'root': {
    'level':LOG_LEVEL,
    'handlers': ['file'],
    'propagate': True
},
```

Check that output from `queue_listener`, and `handle_queue_message` is present in logs.

dtasev:
### Summary of work
Adds root logger that should be inherited by all `getLogger` calls.

### How to test your work
This needs to be deployed on the dev env so that we can see if queue_listener starts logging now.

Fixes #795

### DIFF
--- a/queue_processors/queue_processor/test_settings.py
+++ b/queue_processors/queue_processor/test_settings.py
@@ -46,6 +46,11 @@ LOGGING = {
             'formatter': 'verbose'
         },
     },
+    'root': {
+        'level':LOG_LEVEL,
+        'handlers': ['file'],
+        'propagate': True
+    },
     'loggers': {
         'queue_processor': {
             'handlers': ['file'],


keiranjprice101 (approved):

JEThompson97 (approved):

Anders-Markvardsen:
merged and deployed on dev environment

PR #802, opened 2020-10-08
Limit reduction script output

### LINKED ISSUE #769: Massive log statements in autoreduction processor
Issue raised by: [developer]

### What?
The output of the log statements:
```python
        logger.info("Calling: %s %s %s %s",
                    python_path, MISC['post_process_directory'], destination,
                    message.serialize()) 
```
```python
        logger.info("Entered add_process. proc=%s message=%s", proc, message)
```

Both within `queue_processors/autoreduction_processor/autoreduction_processor.py` 

These 2 log statements produce massive logging statements that are unreadable. We should look to reformat, or trim these down.

### Where?

### How?
How did the issue come about
 `queue_processors/autoreduction_processor/autoreduction_processor.py` 
### Reproducible?
[Yes]

Run a manual submission, check the logs and see if it is legible.

keiranjprice101:
### Summary of work
Enable the limit reduction script flag when logging the message

The outputs in the `autoreduction_processor.log` are shorter but I still think they are pretty long 

Closes #769

### DIFF
--- a/queue_processors/autoreduction_processor/autoreduction_processor.py
+++ b/queue_processors/autoreduction_processor/autoreduction_processor.py
@@ -79,7 +79,7 @@ class Listener(stomp.ConnectionListener):
         python_path = sys.executable
         logger.info("Calling: %s %s %s %s",
                     python_path, MISC['post_process_directory'], destination,
-                    message.serialize(limit_reduction_script=True))    # TODO: limit reduction script  #pylint:disable=fixme
+                    message.serialize(limit_reduction_script=True))
         proc = subprocess.Popen([python_path,
                                  MISC['post_process_directory'],
                                  destination,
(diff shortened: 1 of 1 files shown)


queue_processors/autoreduction_processor/autoreduction_processor.py
```diff
@@ -79,7 +79,7 @@ def hold_message(self, destination, data, headers):
         python_path = sys.executable
         logger.info("Calling: %s %s %s %s",
                     python_path, MISC['post_process_directory'], destination,
-                    message.serialize())    # TODO: limit reduction script  #pylint:disable=fixme
```
JEThompson97, line 82: 😆 Seems like writing the todo was the same amount of work than to just do it?
I'm assuming the todo was written before the parameter was added or something? Just looks funny as a PR here, haha

JEThompson97 (approved):

queue_processors/autoreduction_processor/autoreduction_processor.py
keiranjprice101: Exactly what I thought!

PR #803, opened 2020-10-08
Moves the test MySQL DB used on Travis to be initialised in Docker

dtasev:
### Summary of work
Swaps out MySQL service in Travis, and the steps for manual configuration of the DB state to a Docker container that is configured with that.

### How to test your work
See that Travis build passes. 

Test out locally using instructions here:

For convenience run this in a new terminal because it will block the terminal you are using:

`docker run --network=host -it dtasev/mysql_autoreduction:latest`

This avoids having to do `docker logs <container id>` to see the output. 

If it is not configuring properly make sure you don't have a running local MySQL service (`sudo service mysql stop`) and try again.

--------------

<details><summary>Expand this for instructions on how to run the DB as a daemon</summary>
If you want it to run in the background you need to add the `--daemon` flag (`-d` for short) like this:

`docker run --network=host --name mysql_autoreduction -dit dtasev/mysql_autoreduction:latest`

A `--name` is also added to easily get logs and restart the container!

Note that it takes a bit while the MySQL DB gets configured, you can check the state by doing `docker logs mysql_autoreduction`

Once you see this, then it is ready for the tests to run:
```
running database
==================== Building Database ======================
Migrating databases from django model
Populating database with test data
Test database successfully initialised
```

You can restart the container with `docker restart mysql_autoreduction`. Alternatively to delete/rerun you need to do `docker kill mysql_autoreduction && docker rm mysql_autoreduction && docker run ...`

</details>

--------------

Then in another terminal do these steps (this assumes you have the autoreduction repo and a suitable environment):
```
# Warning: this will overwrite any manual changes to credentials.ini and settings.py - make sure to backup those!

python3 setup.py test_settings
python3 setup.py externals -s activemq

# start ActiveMQ
python3 setup.py start
```
Then just run `pytest`

### Additional comments

- Instructions above not guaranteed to be 100% complete - I haven't manually ran those in bulk yet, just writing from memory and bash `history`!

- FIXME: The DB configuration and the test execution is basically a race condition - if the DB is not configured and Travis installs the environment and reaches `pytest` then the tests will fail. Need to add a check that last line of logs equals `Test database successfully initialised` before continuing. Preferably with a minute or so timeout
  - Get last line with `docker logs mysql_autoreduction | tail -n2 | sed '$ d'`

- Using `--network=host` is a way around having to find the IP of the Docker container. Last I heard this isn't very recommended and we will need to review this (although it is not very different than just running the DB locally). I am also not sure if it works with `docker-compose` at all, as that will just want to expose port `3306` - but this should be fine as long as `docker-compose` configures the expected IP correctly.

### To do
- Move the image hosting on autoreduction's docker account
- Set up automated dockerfile build on hub.docker.com

Fixes ??? (todo: find issue)


Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN apt-get update && apt-get install -y sudo && apt-get install -y software-properties-common
+# RUN useradd -m docker && echo "docker:docker" | chpasswd && adduser docker sudo
+
+ADD requirements.txt .
+
+RUN python -m pip install -r requirements.txt
+
+ADD ./build ./build
+ADD ./setup.py .
+# ADD test_credentials.log ./utils/test_credentials.ini
```
keiranjprice101, line 16: I guess all these commented out hacks could be removed :D

dtasev: Yeah I need to clean-up the files and add a proper readme, as well as build scripts to automate the builds. 

I think ultimately there will be 3 Dockerfiles for DB/Queues/WebApp, we can probably base all 3 on the WebApp one though, and then figure out how to compose it so that you can just do `docker-compose run autoreduction.yml` (or something like this 😄) and get the whole thing running

Anders-Markvardsen:
@dtasev is this ready for PR testing? Does it relate to an existing issue?

dtasev:
This is too out of date

Issue #809, opened 2020-10-13
Look at database reset script


keiranjprice101:
Issue raised by: [developer]

### What?
The database reset script presented me with some problems when it came to the backup part.
It wont work on linux at all.

It would be good if we could optionally pass a backup directory to the script. We also need to verify that the backup is working correctly.

### Where?
`scripts/database_management/reset_database_post_cycle.py`

### How?
Identified in #766 
### Reproducible?
[Yes]

### How to test the issue is resolved
Be able to run the script on a linux/mac machine
Be able to optionally pass the backup dir
Backup and wipe functionality works hassle free.

dtasev:
Ideally we want to look into a way of doing this with Django, because it'll be more convenient

Issue #808, opened 2020-10-13
Look to pinning dependencies


keiranjprice101:
Issue raised by: [developer]

### What?
We should look to pin the versions of our dependencies. This will also be a good time to consider using pip-tools to generate requirements.txt (Nice tool, used before) and look at dependabot

### Where?
`setup.py` `requirements.txt`

### How?
Today the sentry dependency updated, which has broken the database setup script. This is the second time in a couple of months a dependency has updated and broken our build

PR #810, opened 2020-10-13
Control which instruments appear in WebApp

### LINKED ISSUE #766: Be able to control which instruments are displayed in the webapp
Issue raised by: [developer]

### What?
We have a setting in `/utils/settings.py`

```
VALID_INSTRUMENTS = ['GEM', 'MAPS', 'MARI', 'MAPS', 'OSIRIS', 'MUSR',
                     'EMU', 'HIFI', 'ARGUS', 'CHRONUS',
                     'POLREF', 'SURF', 'INTER', 'WISH']
```

Currently this setting has no effect on which instruments are displayed in the webapp after the database has been wiped, which is perhaps the right behaviour; regardless a comment needs added to this setting to explain.

An alternative bet for clearing the instrument names in the webapp is to wipe the database using the script `scripts/database_management/reset_database_post_cycle.py`. However, perhaps because of recent database model changes, this does not have any effect on which instruments are displayed in the webapp either.

On closer inspection the reason seems to relate to database table autoduction.reduction_viewer_instrument not be cleared, but this needs to be checked. 

### Where?
Database tables and associated script

### Reproducible?
Yes

Use documentated, added with this issue, action for resetting the instrument names. Following this use the manual_submission script to either dynamically add new instruments to the webapp to display and/or demonstrate that runs and be added to those in the VALID_INSTRUMENTS setting. Look into adding regression test where this is feasible within time constraint

keiranjprice101:
Note: Travis fails because of `sentry_sdk` issue mentioned on teams and referenced here #808

### Summary of work
The instrument buttons in the overview page are now populated from `VALID_INSTRUMENTS` from `utils.settings`.
The reset script will also now delete the instrument table.
I had quite a few issues running the script which were out of the scope of this issue, so I have opened #809

### How to test your work
Add an instrument to `VALID_INSTRUMENTS` and verify it appears on the overview page
Remove an instrument and verify it is removed from the overview page
Run the database reset script and verify that all instruments are also removed
Run `setup.py database` and verify all models successfully migrated.

### Additional comments
 

closes #766

### DIFF
--- a/WebApp/autoreduce_webapp/reduction_viewer/views.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/views.py
@@ -37,6 +37,8 @@ from plotting.plot_handler import PlotHandler
 
 from utilities.pagination import CustomPaginator
 
+from utils.settings import VALID_INSTRUMENTS
+
 LOGGER = logging.getLogger('app')
 
 
@@ -103,7 +105,7 @@ def overview(_):
     Render the overview landing page (redirect from /index)
     Note: _ is replacing the passed in request parameter
     """
-    instruments = Instrument.objects.order_by('name')
+    instruments = VALID_INSTRUMENTS
     context_dictionary = {}
     if instruments:
         context_dictionary = {'instrument_list': instruments}
--- a/WebApp/autoreduce_webapp/templates/overview.html
+++ b/WebApp/autoreduce_webapp/templates/overview.html
@@ -6,8 +6,8 @@
     <div class="instrument-btns-row row form-group">
     {% if instrument_list %}
         {% for instrument in instrument_list %}
-                <div class="instrument-btn col-md-4 text-center" id={{ instrument.name }}-instrument-btn>
-                    <a href="/instrument/{{ instrument.name }}" class="btn btn-primary btn-lg btn-block">{{ instrument.name }}</a>
+                <div class="instrument-btn col-md-4 text-center" id={{ instrument }}-instrument-btn>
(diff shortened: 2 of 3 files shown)


JEThompson97 (approved):

WebApp/autoreduce_webapp/reduction_viewer/views.py
```diff
@@ -103,7 +105,7 @@ def overview(_):
     Render the overview landing page (redirect from /index)
     Note: _ is replacing the passed in request parameter
     """
-    instruments = Instrument.objects.order_by('name')
+    instruments = VALID_INSTRUMENTS
```
rickybassom, line 108: Should we assume that `VALID_INSTRUMENTS` is ordered alphabetically? Maybe we could do a sort here?
```suggestion
    instruments = VALID_INSTRUMENTS.sort()
```

or where `VALID_INSTRUMENTS` is defined:

```python
VALID_INSTRUMENTS = ['ENGINX', 'GEM', 'HRPD', 'MAPS', 'MARI', 'MUSR',
                     'OSIRIS', 'POLARIS', 'POLREF', 'WISH'].sort()
```

rickybassom (commented):
The changes look good :+1:

I noticed in my testing that 'LARMOR' and 'MAPS' were missing on the overview page. Is that expected?

WebApp/autoreduce_webapp/reduction_viewer/views.py
Anders-Markvardsen: Nice suggest Ricky, but I would prefer not to sort, because we may want to control the order manually that these are listed

Anders-Markvardsen:
thanks, one less thing to get confused with in the future and deployed on dev environment

PR #826, opened 2020-10-15
Lowers IPython version pin to 7.16.1

dtasev:
### Summary of work
This is due to 7.16.1 being the highest one that pip can find on CentOS.
Also added comments to warn about this as travis runs Ubuntu and thus
if `dependabot` opens a PR with an update the builds will pass successfully,
but then fail to install on our environments

### How to test your work
On dev queue env:
```
virtualenv /tmp/pr
source /tmp/pr/bin/activate

# Go into autoreduction folder and just install
pip3 install -e .
pip3 install -r requirements.txt
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,9 @@ dash==1.16.3
 django-debug-toolbar==3.1.1
 django_plotly_dash==1.4.2
 gitpython==3.1.7
-ipython==7.18.1
+# this is the highest available version that pip can find on CentOS - be careful when updating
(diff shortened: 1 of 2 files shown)


Anders-Markvardsen (approved):
tested it on autoreducedev with

```
virtualenv /tmp/pr826
source /tmp/pr826/bin/activate
git clone  https://github.com/ISISScientificComputing/autoreduce.git testPR826
cd testPR826/
pip3 install -e .
git checkout -b lower_ipython_version origin/lower_ipython_version
pip3 install -e .
pip3 install -r requirements.txt
```

Installed and worked

PR #849, opened 2020-10-21
Catch the base exception

### LINKED ISSUE #845: Fix the exception handling in reduction viewer utils
Issue raised by: [developer]

### What?
line 189 in WebApp/autoreduce_webapp/reduction_viewer/utils.py there is an attempt to catch a module instead of an exception. This results in the root exception being swallowed and a new exception being thrown.

### Where?
line 189 in WebApp/autoreduce_webapp/reduction_viewer/utils.py

### How?
Beta testing

### Reproducible?
[Yes]
Rerun jobs for HRPD and GEM

Root exception should be logged and rethrown

keiranjprice101:
### Summary of work
We catch the base exception allowing it to be properly logged and reported instead of being swallowed and a new exception rethrown.

This should be made more specific in the future, but right now it is more important that we can identify the issue that is being hidden

### How to test your work
Resubmit a HRPD run. And the true exception is revealed.

### Additional comments
 

closes #845 

### DIFF
--- a/WebApp/autoreduce_webapp/reduction_viewer/utils.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/utils.py
@@ -185,8 +185,7 @@ class ReductionRunUtils(object):
         # Check record is safe to save
         try:
             new_job.full_clean()
-        # pylint:disable=catching-non-exception
-        except django.core.exceptions as exception:
+        except Exception as exception:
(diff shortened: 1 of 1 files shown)


Anders-Markvardsen (approved):
Agree with change

Anders-Markvardsen:
And backported into release branch, plus updated and restarted reduceddev2 accordingly

PR #851, opened 2020-10-21
Allow blank software

### LINKED ISSUE #850: Rerunning is not working
Issue raised by: [developer]

### What?
It is not possible to rerun a reduction. When the run is submitted model validation appears to fail because the software field on `ReductionRun` object is blank or null.

### Where?
`WebApp/autoreduce_webapp/reduction_viewer/utils.py`

### How?
Smoke testing

### Reproducible?
[Yes]
click rerun jobs, then submit. If #849 is not merged the exception will be hidden

IT will be possible to rerun jobs in the webapp

keiranjprice101:
### Summary of work
Quick fix to allow reruns will open an issue (#852) to reinforce this and identify why the software was not included originally.

### How to test your work
Verify that a run can be resubmitted in the WebApp

### Additional comments
 

closes #850 

### DIFF
--- a/WebApp/autoreduce_webapp/reduction_viewer/models.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/models.py
@@ -111,6 +111,8 @@ class ReductionRun(models.Model):
                                    on_delete=models.CASCADE)
     retry_run = models.ForeignKey('self', on_delete=models.SET_NULL, null=True, blank=True)
     status = models.ForeignKey(Status, blank=False, related_name='+', on_delete=models.CASCADE)
+    # Allowed software field to be black in code line below. Issued opened (#852) to later
+    # populate this field
     software = models.ForeignKey(Software, blank=True, related_name='reduction_runs', null=True,
                                  on_delete=models.CASCADE)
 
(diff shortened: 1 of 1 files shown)


WebApp/autoreduce_webapp/reduction_viewer/models.py
```diff
@@ -111,7 +111,7 @@ class ReductionRun(models.Model):
                                    on_delete=models.CASCADE)
     retry_run = models.ForeignKey('self', on_delete=models.SET_NULL, null=True, blank=True)
     status = models.ForeignKey(Status, blank=False, related_name='+', on_delete=models.CASCADE)
-    software = models.ForeignKey(Software, blank=False, related_name='reduction_runs', null=True,
+    software = models.ForeignKey(Software, blank=True, related_name='reduction_runs', null=True,
```
Anders-Markvardsen, line 116: A minor. I would suggest that adding a comment above this line saying something like "Allowed software field to be black for now and issued opened (#852) to later this field gets populated properly later".

JEThompson97 (approved):
So if I understand correctly: this is a quick fix pending a fuller fix (#852)?

Anders-Markvardsen (approved):
agree looks good

Anders-Markvardsen:
back ported to release branch and dev webapp server updated and restarted

PR #854, opened 2020-10-21
Cleanup run summary for skipped runs

### LINKED ISSUE #853: Tidy up run Summary page for skipped runs
Issue raised by: [developer]
### What?
From #844 originally
When a run is skipped the run summary page should:
 - Start time: Not run
- End time: Not run
- Duration: Not run
- Rerun section: Not visible
- Logs: Not shown

### Where?
`WebApp/autoreduce_webapp/templates/run_summary.html`

### How?
Smoke testing

### Reproducible?
[Yes]
View a skipped run to see the current state

The criteria above should be true when viewing a skipped run

keiranjprice101:
### Summary of work
Alters the template for a skipped run
Changes start and end times, duration and hides logs and rerun option

This PR:
image

Current:
image

### How to test your work
Skipped runs should match updated format

### Additional comments
 

Closes #853 

### DIFF
--- a/WebApp/autoreduce_webapp/reduction_viewer/views.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/views.py
@@ -260,6 +260,7 @@ def run_summary(_, instrument_name=None, run_number=None, run_version=0):
         run = ReductionRun.objects.get(instrument=instrument,
                                        run_number=run_number,
                                        run_version=run_version)
+        # run status value of "s" means the run is skipped
         is_skipped = run.status.value == "s"
         history = ReductionRun.objects.filter(run_number=run_number).order_by('-run_version')
         started_by = started_by_id_to_name(run.started_by)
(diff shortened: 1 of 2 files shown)


WebApp/autoreduce_webapp/reduction_viewer/views.py
```diff
@@ -260,6 +260,7 @@ def run_summary(_, instrument_name=None, run_number=None, run_version=0):
         run = ReductionRun.objects.get(instrument=instrument,
                                        run_number=run_number,
                                        run_version=run_version)
+        is_skipped = run.status.value == "s"
```
JEThompson97, line 263: So `run.status.value` uses a character to indicate the status?

JEThompson97 (approved):

WebApp/autoreduce_webapp/reduction_viewer/views.py
keiranjprice101: Yes, I think we should create an ENUM and some point then we can just do run.status == Statuses.SKIPPED.
Low priority though I think

JEThompson97: I like that idea 🙂

Anders-Markvardsen:
and thanks for additional comment. Back ported and reducedev2 updated accordingly

Issue #763, opened 2020-09-07
WebApp imports


keiranjprice101:
Issue raised by: [developer]

### What?
Some imports within the webapp are treating the `autoreduce_webapp` directory as the top-level directory of the project. We should either change this to be consistent with every other package within the project, document why this is done or if the WebApp does not use anything from a different package we could consider moving it to a separate repository (This would remove IDE warnings and slim down this repo)

### Where?
Various modules in `WebApp` e.g. `WebApp/autoreduce_webapp/reduction_variables/utils.py`

### How?
Possibly this is a result of how the webapp is deployed and the directory it is deployed from

### Reproducible?
[Yes]

### How to test the issue is resolved
This will depend on the solution we go with

dtasev:
I believe this is done because that's how Django apps usually work. I'm not sure if you can change it without also manually configuring lots of assumptions that Django has. 

Might be possible, but I agree that the best solution is to move it into a separate repository as it ought to be.

PR #868, opened 2020-10-26
846 disable rerun for instruments with novariables

### LINKED ISSUE #846: Rerun not possible for runs that do not have variables.

### What?
It is not possible to rerun a job for an instrument that does not have variables.
We should either hide the buttons for rerunning for an instrument like this. Or allow a rerun to be submitted for this scenario.

Currently this only effects WISH who do not use the WebApp anyway.

### How?
Smoke testing
### Reproducible?
[Yes]

Either button is not displayed for these instruments or it is possible to rerun these jobs.

keiranjprice101:
### Summary of work
Removes the option to rerun jobs through the WebApp if an instrument does not have variables. (As is currently the case only for WISH)

### How to test your work
 
Check that the rerun option in the run-summary and instrument-summary do not appear for WISH only.

### Additional comments
 

closes #846

### DIFF
--- a/WebApp/autoreduce_webapp/reduction_viewer/views.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/views.py
@@ -274,7 +274,8 @@ def run_summary(_, instrument_name=None, run_number=None, run_version=0):
         rb_number = Experiment.objects.get(id=run.experiment_id).reference_number
         has_variables = bool(
             InstrumentVariablesUtils().get_default_variables(run.instrument.name)
-            or run.run_variables.all())
+            or run.run_variables.all()) # We check default vars and run vars in case none exist
+                                        # for run but could exist for default
 
         context_dictionary = {'run': run,
                               'is_skipped': is_skipped,
@@ -342,7 +343,7 @@ def instrument_summary(request, instrument=None):
         if len(runs) == 0:
             return {'message': "No runs found for instrument."}
 
-        has_variables = InstrumentVariablesUtils().get_default_variables(instrument_obj.name)
+        has_variables = bool(InstrumentVariablesUtils().get_default_variables(instrument_obj.name))
 
         context_dictionary = {
             'instrument': instrument_obj,
--- a/WebApp/autoreduce_webapp/reduction_viewer/views.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/views.py
@@ -339,6 +339,8 @@ def instrument_summary(request, instrument=None):
         if len(runs) == 0:
             return {'message': "No runs found for instrument."}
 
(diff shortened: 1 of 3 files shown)


WebApp/autoreduce_webapp/reduction_viewer/views.py
```diff
@@ -339,6 +342,8 @@ def instrument_summary(request, instrument=None):
         if len(runs) == 0:
             return {'message': "No runs found for instrument."}
 
+        has_variables = InstrumentVariablesUtils().get_default_variables(instrument_obj.name)
```
Anders-Markvardsen, line 345: Is there a bool() missing here around InstrumentVariablesUtils().get_default_variables(instrument_obj.name)?

WebApp/autoreduce_webapp/reduction_viewer/views.py
```diff
@@ -273,12 +272,16 @@ def run_summary(_, instrument_name=None, run_number=None, run_version=0):
             reduction_location = reduction_location.replace('\\', '/')
 
         rb_number = Experiment.objects.get(id=run.experiment_id).reference_number
+        has_variables = bool(
```
Anders-Markvardsen, line 275: Could you add a comment explaining why you check for both InstrumentVariablesUtils().get_default_variables(run.instrument.name) and run.run_variables.all() ?

WebApp/autoreduce_webapp/reduction_viewer/views.py
```diff
@@ -16,25 +16,24 @@
...
-                                          require_admin, check_permissions)
-from plotting.plot_handler import PlotHandler
-from reduction_variables.utils import MessagingUtils
+from reduction_variables.utils import InstrumentVariablesUtils, MessagingUtils
 from reduction_viewer.models import Experiment, ReductionRun, Instrument, Status
 from reduction_viewer.utils import StatusUtils, ReductionRunUtils
 from reduction_viewer.view_utils import deactivate_invalid_instruments
 from utilities.pagination import CustomPaginator
 
+from plotting.plot_handler import PlotHandler
```
Anders-Markvardsen, line 36: Is the reason for moving the order of the above headers round to have them listed in alphabetic order?

Anders-Markvardsen (changes requested):
Just received the code in this review. See comments

WebApp/autoreduce_webapp/reduction_viewer/views.py
keiranjprice101: This was automatic by IDE.. I think the reason the order seems strange is because of #763

Anders-Markvardsen (approved):

PR #872, opened 2020-10-27
Implements fix for correctly updating the Python dicts

### LINKED ISSUE #870: Variables set in the web interface are not being used by the reduction
Issue raised by: user

### What?
The variables set in the web interface are not being used by the reduction.

It turned out that the webserver is passing the correct information to the python reduction process but that the dictionary update was not happening.

### Where?
https://github.com/ISISScientificComputing/autoreduce/blob/master/queue_processors/autoreduction_processor/post_process_admin.py#L120

### How?
Seems like leftover code during switch from Python 2 to 3! `map` is lazy in Python 3 and the result WILL NOT BE computed unless it's used in a for loop, or other sequence call (e.g. calling `list(map(merge_dicts, ["standard_vars", "advanced_vars"])`)

### Reproducible?
Yes

See comment below

dtasev:
### Summary of work
Replaces the map with the 2 manual calls.

### How to test your work
Run webapp locally with the fix and resubmit a job with changed parameters. The parameters from the WebApp should now override the standard parameters

Have a look at this comment https://github.com/ISISScientificComputing/autoreduce/issues/870#issuecomment-717188785

Fixes #870 

Release notes added to the release notes PR in commit https://github.com/ISISScientificComputing/autoreduce/commit/9f9f19f5d569a34780c8b79ab7e32faad71d41d0

### DIFF
--- a/queue_processors/autoreduction_processor/post_process_admin.py
+++ b/queue_processors/autoreduction_processor/post_process_admin.py
@@ -117,7 +117,8 @@ class PostProcessAdmin:
 
         if not hasattr(reduce_script, "web_var"):
             reduce_script.web_var = types.ModuleType("reduce_vars")
-        map(merge_dicts, ["standard_vars", "advanced_vars"])
+        merge_dicts("standard_vars")
+        merge_dicts("advanced_vars")
         return reduce_script
 
     @staticmethod


keiranjprice101 (approved):
Working on dev.
Amazing this was only recently found

dtasev:
Cherry-picked on release branch https://github.com/ISISScientificComputing/autoreduce/commit/8887db47bd4ca25f87872cfdb05f0cd479624147

Issue #875, opened 2020-10-29
DataArchiveCreator prevents changes to settings.py


keiranjprice101:
Issue raised by: [developer]

### What?
This is pretty convoluted to explain so I will try to bullet point it
- `queue_processors/autoreduction_processor/test_settings.py` has 2 settings `ceph_directory` and `scripts_directory`. (As of right now there is also a comment above them that shows the settings used for the dev node and probably prod)
- In `test_end_to_end.py` the class `DataArchiveCreator` is used to setup a fake data archive for storing datafiles/scripts/etc. 
- The `DataArchiveCreator` will only set up a directory that matches the current value in `test_settings.py` This means 2 different things: 
  * If those settings are ever changed `test_end_to_end.py` will break, as the `DataArchiveCreator` is not using the setting but is hardcoded to create a directory matching the setting as it exists right now.
  * The production and development builds would technically fail these 2 end to end test cases, more generally, the end to end test only tests the software in one specific configuration, that is only used in travis and not anywhere else
 - The docstring for `DataArchiveCreator` also claims to produce an archive identical to isis data archive. It doesn't

### Where?
Settings at  `queue_processors/autoreduction_processor/test_settings.py` 
Archive creator at `utils/data_archive/data_archive_creator.py`
tests at `systemtests/test_end_to_end.py`

### How?
A need to change the settings revealed this.

### Reproducible?
[Yes]
Change the setting and these 2 tests will fail.

### How to test the issue is resolved
The DataArchiveCreator should create a data archive based on the 2 settings from the settings file, not hard coded location. The end to end tests should then be able to pass on any environment (including dev and prod in theory)

dtasev:
Might be fixed via changes in https://github.com/ISISScientificComputing/autoreduce/tree/queue_processor_refactor

dtasev:
I touched this in #1022 but I'm not sure how far this has been fixed.

The DataArchiveCreator doesn't use any of the settings.py so the `test_end_to_end` will still fail (even with the rewrite) if the script_directory is changed.

Maybe making a new variable that holds the `ARCHIVE_ROOT={PROJECT_ROOT}/data-archive` path and builds `SCRIPTS_DIRECTORY = f"{ARCHIVE_ROOT}/NDX%s/user/scripts/autoreduction/"`. Then on dev/prod we just make `ARCHIVE_ROOT=/isis` and all should work.

PR #878, opened 2020-10-30
Split up PostProcessAdmin

### LINKED ISSUE #756: PostProcessAdmin is confusing and potentially broken.
Issue raised by: [developer]

This will most likely take a few issues/PR to fix/come up with solution

### What?
PostProcessAdmin is confusing and potentially broken. There are some parts which are chains of method calls that are hard to follow, and these methods are often confusing and/or misleading.
Summary below
- The name is unanimously disliked and misleading.
- The combination of its name and the large number of responsibilities it has makes it unclear what it is meant to represent as an object
- Too many responsibilities

The next section is split into confusing names, (method names that are not immediately clear on what the function of a method is), misleading names (method names that imply one thing is done, but something else is done, or an unexpected side effect would occur) and finally, methods that are broken in some way.

- Confusing method names:
    - `specify_instrument_directories`
    - `write_and_readability_checks`
    - `create_final_result_and_log_directory` - Name is ambiguous, does this create final result and log its directory or create 2 directories?
    - `reduction_as_module`
    - `additional_save_directories_check`
    - `_new_reduction_data_path`
    - `_copy_tree`
- Misleading method names:
    - `create_log_path` (This does not create anything, also does not return a string as the docstring claims)
    - `check_for_skipped_runs`
    - `create_directory` takes a list of paths and creates all of them, not a single directory as implied
    - `check_for_skipped_runs` does not check for skipped runs, it checks if the current run is skipped.
    - `check_for_skipped_runs` This method will also send a message, and can call `main` (unexpected based on name)
    - `reduction_as_module` This method also calls `check_for_skipped_runs` and could invoke the side effects mentioned above.
    - `validate_reduction_as_module` has the same side effects as above
- Broken methods or control flow
    - `verify_directory_access` This method will return true if a given location is readble or writable (based on arg). However if it is not readable or writable an exception will be raised. This method is called as if it would return `False` within `write_and_readability_checks` and is meant to raise an OSError, ofcourse this will never happen as the original exception is not caught.
    - `check_for_skipped_runs` will raise an `UnboundLocalError` if a run is skipped

It is a possibility there is more broken within `PostProcessAdmin` but this is from a quick look only. 

### Where?
`queue_processors/autoreduction_processor/post_process_admin.py`

### How?
Some of these problems are old and some are related to recent changes.

### Reproducible?
[Yes]

It should be clear exactly what PostProcessAdmin is, what its responsibilities are and it should be easy to follow.

### LINKED ISSUE #865: Wrong error displayed in webapp when reduce.py fails
Issue raised by: Observed end of Friday by @rickybassom, @dtasev and me

### What?
Observed LARMOR and MARI, but looks to be a bug in general,  e.g. see 

http://reducedev2.isis.cclrc.ac.uk:8080/runs/MARI/27437/2/
http://reducedev2.isis.cclrc.ac.uk:8080/runs/LARMOR/51568/1/

Take the MARI example above, which from viewing "Show reduction logs" states the reduce.py error is

```
RuntimeError: *** Invalid properties found. Can not run convert_to energy
```

however the error displayed to the user is 

```
Unable to copy to /instrument/MARI/RBNumber/RB2010450/autoreduced/ - [Errno 2] No such file or directory: '/autoreducetmp/instrument/MARI/RBNumber/RB2010450/autoreduced/'
```

which is misleading. The same wrong error is shown for the LARMOR example above. What should be displayed is the reduce.py error caught.

Also, with this issue: ensure that the reduction logs gets saved, i.e. the logs going into the sub-folder reduction_log/. For sanity please double check that the logs for a "Completed" run are outputted to reduction_log/. logs should be outputted to reduction_log/ for all cases except "Skipped" runs.

keiranjprice101:
@dtasev as promised. Probably best not for this release but to test in between on dev env

#### What is done:

- Introduced classes as per discussion with @dtasev and @rickybassom 
  - Due to #875 `TemporaryReductionDirectory` no longer extends `ReductionDirectory` and instead makes use of the `tempfile` module and is actually nicer as we can completely delegate responsibility of cleanup of temp dirs (Currently broken on dev)
- Introduced new exceptions within `reduction_exceptions.py` and moved the existing `SkippedRunException`
- Move reduction responsibility away from `PostProcessAdmin` to `reduction_service.py`
- More testing
- Remove explicit validation as it becomes inherent within design.

#### What is working:
Through limited testing on dev
- Runs successfully for some flat output and non flat instruments GEM/WISH/HRPD
- With logging as is reduction_log modal is still populated
- Data location is still populated

#### What needs to be done:
- Improve the way logging is setup in both `post_process_admin` and `reduction_service`
- Look at `channels_redirected` (Out of scope of this PR potentially)
- Rename `post_process_admin` as discussed (Again potentially out of scope.)
- `reduction_service` should really be in a package separate to `queue_processors`
- The reduce method within `post_process_admin` remains untested

### How to test your work
Deploy the branch to dev and check runs go through, including logs written to and displayed in WebApp

### Additional comments
 

Closes #756 Closes #865


queue_processors/autoreduction_processor/reduction_exceptions.py
```diff
@@ -0,0 +1,28 @@
...
+class ReductionScriptError(Exception):
+    """
+    Exception raised when an unhandled exception is produced within the reduction script
+    """
+
+
+class SkippedRunException(Exception):
+    """
+    Exception for runs that have been skipped
+    Note: this is currently only the case for EnginX Event mode runs at ISIS
```
Anders-Markvardsen, line 27:
```suggestion
    Exception for runs that have been skipped
```

Anders-Markvardsen: when this code first was written it was probably only an exception for EnginX. However now we label everything that is not an exceptable RB number as skipped

queue_processors/autoreduction_processor/reduction_exceptions.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+
+
+class ReductionScriptError(Exception):
+    """
+    Exception raised when an unhandled exception is produced within the reduction script
+    """
+
+
+class SkippedRunException(Exception):
```
Anders-Markvardsen, line 24: I like created a separate files for exceptions. However, is SkippedRunException and exception any more? I.e. now that we report skipped runs in webapp has skipped run become part of normal operation

queue_processors/autoreduction_processor/reduction_service.py
```diff
@@ -0,0 +1,210 @@
...
+    LOGGER.info("Temporary log dir: %s", temp_dir.log_path)
+    LOGGER.info("Final log dir: %s", reduction_dir.log_path)
+    LOGGER.info("Datafile: %s", datafile.path)
+    LOGGER.info("Reduction script: %s", script.script_path)
+    LOGGER.info("-------------------------------------------------------")
+    LOGGER.info("Starting reduction...")
+
+
+    script.load()
+    if run_number in script.skipped_runs:
```
Anders-Markvardsen, line 187: Hi @keiranjprice101 I looked and there is no reference to `skipped_runs` in any of our reduce.py and reduce_vars.py files, and neither on help page on webapp. Hence please delete this if statement. Further I am not sure there is any point in the SkippedRunException. Unless you see reason for keeping it could you delete this exception also please

Anders-Markvardsen (changes requested):
Overall looks good and thanks.

Requested one change concerning skipped_runs and exception

Anders-Markvardsen (approved):

Anders-Markvardsen:
Tested with GEM and HRPD and come through well

Anders-Markvardsen:
updating dev compute node to include this change

PR #885, opened 2020-11-02
Can't submit OSIRIS run with manual_submission

### LINKED ISSUE #884: Can't submit OSIRIS run with manual_submission
Issue raised by: developer

### What?
Manual submission fails with

```
python manual_submission.py OSIRIS 144378
```

because ICAT prefix is OSI however *.nxs files for this instrument have format OSIRIS00144378.nxs

extend script to handle this use case also.

Anders-Markvardsen:
### Summary of work
Enable OSIRIS to work with manual_submission.

Also tied up doc of method changed in recent PR.

### How to test your work
* test that can submit e.g. `python manual_submission.py OSIRIS 144376`
* test readability

### Additional comments
Not sure how easy it is to regression test this with the current mocking setup. Hence left this as is, but ideas welcome

Fixes #884

### DIFF
--- a/scripts/manual_operations/manual_submission.py
+++ b/scripts/manual_operations/manual_submission.py
@@ -72,6 +72,23 @@ def get_location_and_rb_from_database(database_client, instrument, run_number):
     return data_location, experiment_number
 
 
+def icat_datafile_query(icat_client, file_name):
+    """
+    Search for file name in icat and return it if it exist.
+    :param icat_client: Client to access the ICAT service
+    :param file_name: file name to search for in icat
+    :return: icat datafile entry if found
+    :raises SystemExit: If icat_client not connected
+    """
+    if icat_client is None:
+        print("ICAT not connected")  # pragma: no cover
+        sys.exit(1)  # pragma: no cover
+
+    return icat_client.execute_query("SELECT df FROM Datafile df WHERE df.name = '"
+                                         + file_name +
+                                         "' INCLUDE df.dataset AS ds, ds.investigation")
+
+
 def get_location_and_rb_from_icat(icat_client, instrument, run_number, file_ext):
     """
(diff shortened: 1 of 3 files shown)


scripts/manual_operations/manual_submission.py
```diff
@@ -87,12 +87,30 @@ def get_location_and_rb_from_icat(icat_client, instrument, run_number, file_ext)
         print("ICAT not connected")  # pragma: no cover
         sys.exit(1)  # pragma: no cover
 
-    icat_instrument_prefix = get_icat_instrument_prefix(instrument)
-    file_name = f"{icat_instrument_prefix}{str(run_number).zfill(5)}.{file_ext}"
+    # look for file-name assuming file-name uses full instrument name
+    file_name = f"{instrument}{str(run_number).zfill(5)}.{file_ext}"
     datafile = icat_client.execute_query("SELECT df FROM Datafile df WHERE df.name = '"
                                          + file_name +
                                          "' INCLUDE df.dataset AS ds, ds.investigation")
```
JEThompson97, line 94: Given we use this same query 4 times by my count, I wonder if it'd be worth refactoring.
The first idea that comes to mind would be a simple method:
```python
def create_query(file_name):
    return f"SELECT df FROM Datafile df WHERE df.name = '{file_name}' INCLUDE df.dataset AS ds, ds.investigation"
```

JEThompson97 (approved):

scripts/manual_operations/manual_submission.py
Anders-Markvardsen: thanks and implemented and with a few smaller updates to tests also

Anders-Markvardsen:
I have taken the backport-PR label off since this fix does not concerns that system in operation, and left it as optional to backport into release.

dtasev:
Looks good

PR #891, opened 2020-11-04
Fix manual_remove multiple version bug

### LINKED ISSUE #669: Allow for Manual Removal of One or More Versions of a Run When Multiple Versions exist
Issue raised by: developer

### What?
It has been brought to light that the functionality to manually remove a run for a given instrument when multiple versions exists longer works. 

We should aim to fix this. 

Actions performed and Error Produced:
```
python manual_remove.py GEM 90374                                                                                                                                                                                                                                     
Discovered multiple reduction versions for GEM90374:
        v10 - 
        v9 - 
        v8 - 
        v7 - 
        v6 - 
        v5 - 
        v4 - 
        v3 - 
        v2 - 
        v1 - 
        v0 - 
Which runs would you like to delete (e.g. 1,2,3): 9 
Traceback (most recent call last):
  File "manual_remove.py", line 260, in <module>
    main()
  File "manual_remove.py", line 256, in main
    remove(instrument, run)
  File "manual_remove.py", line 191, in remove
    manual_remove.process_results()
  File "manual_remove.py", line 58, in process_results
    self.multiple_versions_found(run_number=key)
  File "manual_remove.py", line 92, in multiple_versions_found
    self.to_delete[run_number].remove(reduction_job)
AttributeError: 'QuerySet' object has no attribute 'remove'
```

### Where?
scripts/manual_operations/manual_remove.py.ManualRemove.multiple_versions_found

### How?
Attempting to remove a run with many versions

### Reproducible?
Yes

* Code review
* No drop in coverage
* Manual testing:  be able to remove one or many versions of a run for a given instrument if many version exist

rickybassom:
### Summary of work
* Fixed functionality to manually remove a run for a given instrument when multiple versions exists

### How to test your work
* Code review
* All tests pass
* Try removing a run using the `manual_remove.py` script for 1 or multiple versions

Fixes #669

### DIFF
--- a/scripts/manual_operations/manual_remove.py
+++ b/scripts/manual_operations/manual_remove.py
@@ -91,7 +91,7 @@ class ManualRemove:
         # Remove runs that the user does NOT want to delete from the delete list
         for reduction_job in self.to_delete[run_number]:
             if not int(reduction_job.run_version) in user_input:
-                # this excludes the versions that the user DID NOT type
+                # this removes the versions that the user DID NOT enter
                 # from the list of run versions that will be DELETED
                 self.to_delete[run_number].remove(reduction_job)
 
--- a/scripts/manual_operations/manual_remove.py
+++ b/scripts/manual_operations/manual_remove.py
@@ -43,7 +43,7 @@ class ManualRemove:
             .filter(instrument=instrument_record.id) \
             .filter(run_number=run_number) \
             .order_by('-created')
-        self.to_delete[run_number] = result
+        self.to_delete[run_number] = list(result)
         return result
 
     def process_results(self):
@@ -93,8 +93,7 @@ class ManualRemove:
             if not int(reduction_job.run_version) in user_input:
                 # this excludes the versions that the user DID NOT type
                 # from the list of run versions that will be DELETED
-                self.to_delete[run_number] = self.to_delete[run_number].exclude(
-                    run_version=reduction_job.run_version)
+                self.to_delete[run_number].remove(reduction_job)
 
     def delete_records(self):
         """
--- a/scripts/manual_operations/manual_remove.py
+++ b/scripts/manual_operations/manual_remove.py
@@ -91,6 +91,8 @@ class ManualRemove:
         # Remove runs that the user does NOT want to delete from the delete list
         for reduction_job in self.to_delete[run_number]:
             if not int(reduction_job.run_version) in user_input:
+                # this excludes the versions that the user DID NOT type
+                # from the list of run versions that will be DELETED
                 self.to_delete[run_number] = self.to_delete[run_number].exclude(
                     run_version=reduction_job.run_version)
 
--- a/scripts/manual_operations/manual_remove.py
+++ b/scripts/manual_operations/manual_remove.py
@@ -91,7 +91,8 @@ class ManualRemove:
         # Remove runs that the user does NOT want to delete from the delete list
         for reduction_job in self.to_delete[run_number]:
             if not int(reduction_job.run_version) in user_input:
-                self.to_delete[run_number].remove(reduction_job)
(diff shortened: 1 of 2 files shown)


scripts/manual_operations/manual_remove.py
```diff
@@ -91,7 +91,8 @@ def multiple_versions_found(self, run_number):
         # Remove runs that the user does NOT want to delete from the delete list
         for reduction_job in self.to_delete[run_number]:
             if not int(reduction_job.run_version) in user_input:
-                self.to_delete[run_number].remove(reduction_job)
+                self.to_delete[run_number] = self.to_delete[run_number].exclude(
```
dtasev, line 94:
```suggestion
                # this excludes the versions that the user DID NOT type
                # from the list of run versions that will be DELETED
                self.to_delete[run_number] = self.to_delete[run_number].exclude(
```

scripts/manual_operations/manual_remove.py
```diff
@@ -91,6 +91,8 @@ def multiple_versions_found(self, run_number):
         # Remove runs that the user does NOT want to delete from the delete list
         for reduction_job in self.to_delete[run_number]:
             if not int(reduction_job.run_version) in user_input:
+                # this removes the versions that the user DID NOT enter
+                # from the list of run versions that will be DELETED
                 self.to_delete[run_number].remove(reduction_job)
```
dtasev, line 96:
```suggestion
        self.to_delete[run_number] = [
            reduction_job for reduction_job in self.to_delete[run_number]
            if reduction_job.run_version in user_input
        ]
```
Maybe we should change this to remove the double negation, and instead make to_delete contain only the run_versions that the user has specified

dtasev:
This is a patch for that adds parsing a range in the run version, but it doesn't amend any of the user-facing messages to say that this is available. Maybe we can keep this to a follow up PR, but it's not a huge change either

```
diff --git a/scripts/manual_operations/manual_remove.py b/scripts/manual_operations/manual_remove.py
index 0616c160..ad5f6b22 100644
--- a/scripts/manual_operations/manual_remove.py
+++ b/scripts/manual_operations/manual_remove.py
@@ -175,6 +175,13 @@ class ManualRemove:
                     processed_input.append(number)
                 except ValueError:
                     return False, []
+        elif "-" in user_input:
+            range_of_versions_to_delete = user_input.split('-')
+            if len(range_of_versions_to_delete != 2:
+                return False, []
+            start = int(range_of_versions_to_delete[0])
+            end = int(range_of_versions_to_delete[1])
+            return True, list(range(start, end))
         else:
             try:
                 user_input = int(user_input)
```

rickybassom:
> This is a patch that adds parsing a range in the run version, but it doesn't amend any of the user-facing messages to say that this is available. Maybe we can keep this to a follow up PR, but it's not a huge change either
> 
> ```
> diff --git a/scripts/manual_operations/manual_remove.py b/scripts/manual_operations/manual_remove.py
> index 0616c160..ad5f6b22 100644
> --- a/scripts/manual_operations/manual_remove.py
> +++ b/scripts/manual_operations/manual_remove.py
> @@ -175,6 +175,13 @@ class ManualRemove:
>                      processed_input.append(number)
>                  except ValueError:
>                      return False, []
> +        elif "-" in user_input:
> +            range_of_versions_to_delete = user_input.split('-')
> +            if len(range_of_versions_to_delete != 2:
> +                return False, []
> +            start = int(range_of_versions_to_delete[0])
> +            end = int(range_of_versions_to_delete[1])
> +            return True, list(range(start, end))
>          else:
>              try:
>                  user_input = int(user_input)
> ```

Nice! I'll open another PR for that change.

rickybassom:
Tested this and it works well

JEThompson97 (approved):
I like a good list comprehension 🙂

dtasev (approved):

PR #906, opened 2020-11-11
Use GitHub Actions for integration testing

### LINKED ISSUE #911: Coveralls is broken
Issue raised by: [developer]

### What?
Coveralls is broken

image

### Where?
https://coveralls.io/github/ISISScientificComputing/autoreduce?branch=master

No errors are logged on Travis for PRs.

### Reproducible?
[Yes]

* Can view code coverage again

rickybassom:
### Summary of work
* Replace Travis CI with GitHub actions

### How to test your work
* The same tests are run in same environment as Travis
* All tests pass

### Additional comments
* Opened this issue about what Python version to use #916 
* **Remember to squash commits when merging!**

Also fixes #911


keiranjprice101:
I know this is in progress, but I notice the python version is 3.8. The dev node runs 3.6.8 and I am not sure from memory about production. But It would be nice if the actions/dev/prod all ran the same version

rickybassom:
> I know this is in progress, but I notice the python version is 3.8. The dev node runs 3.6.8 and I am not sure from memory about production. But It would be nice if the actions/dev/prod all ran the same version

Yeah, I wasn't sure what version to use. I'm not sure which dev node uses 3.6.8?

Here's what we are using across servers using `python3 --version`:
- reducedev2: v3.7.4
- reducequeuedev: v3.7.4
- reduce: v3.7.8
- autoreduce: v3.6.8

Others:
- Travis-CI: v3.7
- Wiki recommends: v3.6

We need to move to using one version. I'll open an issue about this. But which version do you think I should use for GH actions?

keiranjprice101:
> But which version do you think I should use for GH actions?

The lowest version you have found is also 3.6.8 and since every forward version should be backwards compatible, I guess 3.6.8 is the safest choice

rickybassom:
Strangely Python 3.6.8 has dependency problems.
https://github.com/ISISScientificComputing/autoreduce/runs/1390816875

rickybassom:
* Had to use 127.0.0.1 instead of localhost for the db
* This also includes a fix for Coveralls. This issue was caused by a dependency problem in `requirements.txt`

requirements.txt
```diff
@@ -15,7 +13,7 @@ pandas==1.1.4
 plotly==4.12.0
 pylint==2.6.0
 pytest-cov==2.10.1
-python-coveralls==2.9.3
+coveralls==2.1.2
```
keiranjprice101, line 16: This is very nitpicky, but can you maintain the alphabetical order in `requirements.txt`

rickybassom: Didn't spot the alphabetical order! Thanks, will update.

keiranjprice101:
> I tried to make some jobs run in parallel to speed it up but with no luck. Any suggestions welcome.

I don't know the specs of the vm but `pylint` can take a `-j <num_threads>` flag to allow multiple threads. In the selenium work I am doing I have also added `pytest-xdist` to allow parallel testing. You could try adding that here too if you wanted then you can call:
`pytest (whatever) -n <num_cpus>`

Anders-Markvardsen (changes requested):
@rickybassom please keep the travis-ci enabled with this PR, unless there is a technical reason that means travis-ci and github-actions cannot be run side-by-side. This reasons for this are: 1) it is more clean this way, i.e. one PR for just adding github-action and another for removing travis-ci 2) if github-action turns out to be terrible we just turn it off and look for another solutions, which we can identify easily by comparing the build times while running these in parallel 3) perhaps the github-action builds will not generate exactly the same output as travis-ci builds and it important to identifying why before killing travis-ci. Does this makes sense?

rickybassom:
> @rickybassom please keep the travis-ci enabled with this PR, unless there is a technical reason that means travis-ci and github-actions cannot be run side-by-side. This reasons for this are: 1) it is more clean this way, i.e. one PR for just adding github-action and another for removing travis-ci 2) if github-action turns out to be terrible we just turn it off and look for another solutions, which we can identify easily by comparing the build times while running these in parallel 3) perhaps the github-action builds will not generate exactly the same output as travis-ci builds and it important to identifying why before killing travis-ci. Does this makes sense?

Yes that makes sense. travis-ci and github-actions can run side by side. We can compare them over the next month or so before support runs out. Also travis-ci currently takes a while to complete (sometimes over an hour) which is a bit annoying but github-actions should complete before it so you won't have to wait as long for a result.

rickybassom:
I think the drop in coveralls should be ignored. We were unaware of the drops between when coveralls broke and now.

Anders-Markvardsen (approved):
Looks good

PR #913, opened 2020-11-11
Now uses the name of the data file to look for plot files

### LINKED ISSUE #794: Plot finding seems limited to be only for MARI and WISH

### What?
The code that generates the regex for the filename of the saved plot image seems to expect only MARI and WISH.
This [dictionary](https://github.com/ISISScientificComputing/autoreduce/blob/master/plotting/plot_handler.py#L23) provides the patterns used to find the plot files. It only contains MARI and WISH, so any other instrument with a plot will immediately `return None` for the filenames [here](https://github.com/ISISScientificComputing/autoreduce/blob/e68548a915716354f308dcf8f44add32e7aae95e/plotting/plot_handler.py#L53-L57)

### How?
Trying to retrieve plots for instruments was causing #760 

### Reproducible?
Yes, go to `/runs/HRPD/78938/30/` - no plot will be shown, although one is available due to testing for #793 

Plots should show for `/runs/HRPD/78938/30/` (it's just a simple sine wave plot, not anything from the data itself)

### To fix
Either:
- add a key for every supported instrument in autoreduction
- or add a wildcard using the instrument name assuming the pattern `<INSTRUMENT_NAME><RUN_NUMBER>.png` to look for filenames

### Thoughts
The 2nd option shouldn't be an issue anymore, as after #793 the page will be rendered even if plot files are missing or encounter an error during retrieving.

We likely want to handle the case where plot files are expected BUT missing, which could be made harder with a wildcard

### LINKED ISSUE #900: Instructions on how to create plot output from reduce.py
Issue raised by: developer

### What?
Currently plotting is supported but it's not clear what the users need to do to get it working.

We need some documentation in the help pages to describe what needs to be done (e.g. needs to be saved out as a PNG in the reduction dir, file naming conventions (if any?), multiple plots, etc).

It would be good if there are code examples with `matplotlib`.

A help page containing answers to the points above (and probably more)

dtasev:
### Summary of work

Previously it relied on a manual dictionary mapping between instrument name
and regex that will be used to find the name.

This has been removed in favour of just looking for image files of the same
name as the name of the input data file.

This change shouldn't break plotting for MARI or WISH - and no other instruments
currently use plotting at all, so they can pick up this naming scheme.

### How to test your work
Change the SFTP in your `credentials.ini` to the development reduce machine. This should allow you to pick up plots even when the web app is running locally.

- Run web app locally with this change
- Go to http://127.0.0.1:8000/runs/MARI/27037/0/ (assuming you use the default django runserver settings)
- Plots should show, or if they don't there shouldn't be a "encountered error while retrieving plots" error
  - If there is an error - check the SFTP credentials
  - If there are no plots it's possible they were cleaned up (as that data is cleaned daily for dev). Manually remove some runs and resubmit them (I used MARI 27037, perhaps try a WISH run that plots as well, I tried 48104 but there weren't any plots in the reduction output) 

Fixes #794 
Fixes #900 (although we may want more information that the bit I've put in)

Not sure we need release notes - we have stated that plotting should work for all instruments and with this PR it should!


dtasev:
I am not sure whether to update the release notes - are we backporting this into 20_3? @Anders-Markvardsen

Anders-Markvardsen:
Tested with local webapp and sftp setup and encountered error "encountered error while retrieving plots" with freshly run

```
python manual_submission.py MARI 27037
```

Also tried with

```
python manual_submission.py OSIRIS 144370
```

which is setup to create `TEST_PLOT.png` and here also message displayed "encountered error while retrieving plots". In this case I would not have expected to see any message because this *.png does not conform to plotting file format

Anders-Markvardsen:
Now above works with adjusted sftp config 👍

Anders-Markvardsen (approved):
looks good

Anders-Markvardsen:
and back ported into release branch and on production reduce server

Anders-Markvardsen:
and back ported to dev env also web server

PR #918, opened 2020-11-12
Add manual remove by range

### LINKED ISSUE #912: Manual remove add remove versions by range feature
Issue raised by: [developer]

### What?
Add remove run versions by range feature. Instead of 1, 2, 3 use 1 - 3.

### Where?
`/scripts/manual_operations/manual_remove.py`

### How?
We are trying to remove many repeated OSIRIS runs

### Reproducible?
[Yes]

The script should allow you to input a range of versions instead of individually.

rickybassom:
### Summary of work
* Added feature to input a range (e.g 0-5) to remove runs
* Code cleanup
* Added new tests
* Improved comments
* Improved text output

### How to test your work
* Find a run with multiple versions and try removing
* Code review
* Test pass 

### Additional comments
- [x] Add feature to remove version numbers for both: `small - large` and `large - small`
- [x] Add tests for range
- [x] Remove the many repeated OSIRIS runs on production using this tool

Fixes #912

### DIFF
--- a/scripts/manual_operations/manual_remove.py
+++ b/scripts/manual_operations/manual_remove.py
@@ -174,6 +174,13 @@ class ManualRemove:
                     processed_input.append(number)
                 except ValueError:
                     return False, []
+        elif "-" in user_input:
+            range_of_versions_to_delete = user_input.split('-')
+            if len(range_of_versions_to_delete) != 2:
+                return False, []
+            start = int(range_of_versions_to_delete[0])
+            end = int(range_of_versions_to_delete[1])
+            return True, list(range(start, end+1))
         else:
             try:
                 user_input = int(user_input)
--- a/scripts/manual_operations/manual_remove.py
+++ b/scripts/manual_operations/manual_remove.py
@@ -180,7 +180,7 @@ class ManualRemove:
                 return False, []
             start = int(range_of_versions_to_delete[0])
             end = int(range_of_versions_to_delete[1])
-            return True, list(range(start, end+1))
+            return True, list(range(start, end + 1))
         else:
             try:
                 user_input = int(user_input)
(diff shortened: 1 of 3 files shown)


keiranjprice101 (approved):
Working nicely. Does this also remove the reduced files or is it from the database only?

rickybassom:
@keiranjprice101 but that's a good point about removing reduced files. I will have to check.

scripts/manual_operations/manual_remove.py
```diff
@@ -75,22 +75,22 @@ def multiple_versions_found(self, run_number):
...
-                               'Please provide a comma separated list of values:')
+            user_input = input('Input was invalid. '
+                               'Please provide a comma separated list or a range of values: ')
             input_valid, user_input = self.validate_csv_input(user_input)
 
         # Remove runs that the user does NOT want to delete from the delete list
-        self.to_delete[run_number] = [
-            reduction_job for reduction_job in self.to_delete[run_number]
+        self.to_exclude_from_deletion[run_number] = [
+            reduction_job for reduction_job in self.to_exclude_from_deletion[run_number]
```
dtasev, line 93: I think this name is misleading now because what is left in `self.to_exclude_from_deletion` after this list comprehension is the stuff that will actually be deleted

dtasev (changes requested):

dtasev:
I think it removes everything, based on these lines - https://github.com/ISISScientificComputing/autoreduce/blob/master/scripts/manual_operations/manual_remove.py#L107-L110

scripts/manual_operations/manual_remove.py
rickybassom: Good point, will change that soon

Anders-Markvardsen (changes requested):
Tested this PR and successfully used it to delete HIFI runs including `python manual_remove.py HIFI 174353 175038`

Please also update `scripts/manual_operations/README.md` accordingly thanks

Anders-Markvardsen (approved):

dtasev (approved):

PR #921, opened 2020-11-12
Adds a script that will backup all reduction scripts from the archive

### LINKED ISSUE #379: New Github Repository to Store Reduction Scripts for Archiving Using a Cron Job
Issue raised by: [developer]

### What?
Design a cron job to push corresponding reductions scripts for each instrument using Autoreduction into an ordered format in a new GitHub repository for archiving purposes. 

### Where?
A new appropriately named GitHub repository containing directories for each Instrument

- Test if the cron job can create a new directory for a reduction script for a new instrument if not already present on repository.
- Test that the cron job can push a new reduction script to GitHub for a given time period (every 5 minutes) only if a change has been made to the reduction script. 
- Test if the Cron Job can handle Merge Conflicts

dtasev:
### Summary of work
Adds a script that will backup all reduction scripts from the archive - it checks grabs all folders starting with NDX and tries to get a `/user/scripts/autoreduction/reduce.py` and `reduce_vars.py` if it finds one it copies it. If it doesn't it skips the directory.

- Added an account `autoreduce-bot` with write permissions to the repository where the scripts are uploaded. This is currently my fork https://github.com/dtasev/AutoreductionScripts, but as soon as I get access to the main repo I will move it there.
- Updated the VALID_INSTRUMENTS list on the development compute node to match the production ones - the backup script uses it to ensure all are captured.
  - And then I removed the dependency on VALID_INSTRUMENTS but I haven't reverted the change on the dev node as it doesn't affect anything

This is currently deployed and running on the dev node, but pointing to the wrong repository.

### How to test your work
Check that https://github.com/dtasev/AutoreductionScripts updates

### Additional comments
TODOs

- [x] Move to https://github.com/ISISSoftwareServices/AutoreductionScripts
- [x] Figure out why scripts dir is not in the repository on the dev compute node
  - [x] Update the cronjob to use the script from the actual repository - this will prevent the need to manually copy changes over
- [x] Consider if documentation in the Wiki is needed for this: added here https://github.com/ISISScientificComputing/autoreduce/wiki/Daily-backing-up-of-reduce.py-and-reduce_vars.py
- [ ] Can probably add a test just for sanity

Fixes #379 

Perhaps this can be a nice 'feature' that we tell people about in `21_1` release?

### DIFF
--- a/scripts/backup_reduction_scripts.py
+++ b/scripts/backup_reduction_scripts.py
@@ -43,6 +43,7 @@ ISIS_MOUNT_PATH = Path("/isis")
 AUTOREDUCTION_PATH = Path("user/scripts/autoreduction")
 REDUCE_FILES_TO_SAVE = ["reduce.py", "reduce_vars.py"]
 
+# STORAGE_DIR is the git repository dir that has been configured to point to the correct remote
 STORAGE_DIR = Path("~/autoreduction_scripts").expanduser().absolute()
 
 logging.basicConfig(filename=get_log_file('backup_reduction_scripts.log'),
--- a/scripts/backup_reduction_scripts.py
+++ b/scripts/backup_reduction_scripts.py
@@ -16,8 +16,15 @@ remote - otherwise this script will fail to commit/push.
 There is also a --dry-run option that will just show which files are traversed,
 but will not do anything.
 
-"""
+The cronjob should be:
 
+0 1 * * * python3 /home/reduce/backup_reduction_scripts.py
+
+The reason it runs at 1AM is that the queue_processor restarting happens at midnight and
+there might be some interference. There shouldn't be, but just in case it doesn't hurt
+to be scheduled for a bit later.
+"""
+import os
 import sys
 import argparse
 import datetime
@@ -32,7 +39,6 @@ from git import Git
 
 from utils.project.static_content import LOG_FORMAT
 from utils.project.structure import get_log_file
-from utils.settings import VALID_INSTRUMENTS
 
 ISIS_MOUNT_PATH = Path("/isis")
 AUTOREDUCTION_PATH = Path("user/scripts/autoreduction")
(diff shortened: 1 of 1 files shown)


keiranjprice101:
This looks good. It is very strange, that the scripts folder is not on the dev node even after pulling. Where did you test this from with this in mind?

Unrelated. It would be nice if in the future this script and the reset script and any other cron jobs ran through jenkins instead. To save faffing around with ssh/remote desktop. Plus we could trigger this script on the completion of the restart script and get alerted. A long way off for now though potentially, but worth making a record of when you next meet with the devops guys

dtasev:
The script dir is now available on the dev env (it has been un-sparsed).

> It would be nice if in the future this script and the reset script and any other cron jobs ran through jenkins instead. To save faffing around with ssh/remote desktop. Plus we could trigger this script on the completion of the restart script and get alerted. A long way off for now though potentially, but worth making a record of when you next meet with the devops guys

I agree but there's an issue that travis/actions aren't going to have access to our servers so we need some access point to run those jobs - and proper authentication + rate limiting etc.

keiranjprice101:
This appears to be working in terms of copying what does exist and attempting to commit and push. I am guessing the reason the it fails to commit and push is because there are no changes

scripts/backup_reduction_scripts.py
```diff
@@ -0,0 +1,150 @@
...
+from git import Git
+
+from utils.project.static_content import LOG_FORMAT
+from utils.project.structure import get_log_file
+
+ISIS_MOUNT_PATH = Path("/isis")
+AUTOREDUCTION_PATH = Path("user/scripts/autoreduction")
+REDUCE_FILES_TO_SAVE = ["reduce.py", "reduce_vars.py"]
+
+STORAGE_DIR = Path("~/autoreduction_scripts").expanduser().absolute()
```
Anders-Markvardsen, line 46:
```suggestion
# STORAGE_DIR is the git repository dir that has been configured to point to the correct remote
STORAGE_DIR = Path("~/autoreduction_scripts").expanduser().absolute()
```

Anders-Markvardsen: An extra comment to help, and where someone has not read the script intro blurb in detail

dtasev:
@keiranjprice101 Yeah if all the copied files are the same then the commit and push do nothing at all, but it seems some files have changed so we can see at least see that it works (e.g. NDXLARMOR)

keiranjprice101 (approved):

Anders-Markvardsen (approved):
Looks good.

As discussed created issue for moving this new script to production out of cycle #928 . Unit test will be added later

Anders-Markvardsen:
Also note made a few cosmetic updates to wiki

PR #932, opened 2020-11-19
Use composite GitHub action to build Autoreduction

rickybassom:
### Summary of work
* Reduce future repeated code by splitting build steps into separate action

### How to test your work
* Tests pass


.github/actions/build-autoreduction/action.yml
```diff
@@ -0,0 +1,40 @@
...
+      shell: bash
+
+    - name: Set up database
+      run: mysql --host 127.0.0.1 --port 3306 -u root < build/database/reset_autoreduction_db.sql
+      shell: bash
+
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
```
dtasev, line 19: I don't think you can have this in the Action - that's what it's complaining about. But you should be able to stick this in the workflow before the action call

keiranjprice101:
@rickybassom  changed base from and back to master to hide commits

rickybassom:
> @rickybassom changed base from and back to master to hide commits

Thanks! That cleans it up a lot

Anders-Markvardsen (approved):
Look good to me.

One minor detail which can be address when it becomes relevant, the name `build-autoreduction/action.yml` is generic and could perhaps be more specific and ref test env, but as said that can be adjusted when more build envs

rickybassom:
> Look good to me.
> 
> One minor detail which can be address when it becomes relevant, the name `build-autoreduction/action.yml` is generic and could perhaps be more specific and ref test env, but as said that can be adjusted when more build envs

Thanks for the review and for merging. An action yaml can either be `action.yml` or `action.yaml` so the directory name (`build-autoreduction`) is essentially the name of the action.

PR #933, opened 2020-11-19
rename travis file

keiranjprice101:
### Summary of work

Rename the travis file correctly.
So travis runs again.
As shown below


rickybassom (approved):
Oops! Yep, ready to merge.

rickybassom:
Interestingly this PR passed :thinking:

keiranjprice101:
@rickybassom Interesting and scary. It seems the list is in reverse order in that test in the build that failed. I think there is something really simple or really nasty happening. But the test in theory should pass regardless of order (from the brief skim of it I did earlier)

keiranjprice101:
@rickybassom Even more interesting. It passed in the PR but the test failed on the merge commit

PR #936, opened 2020-11-20
Fix the failing test

### LINKED ISSUE #931: test_filter_files_by_time_add_multi
Issue raised by: [developer]

### What?
`test_filter_files_by_time_add_multi` This test is failing on the github test job on the master branch.

It is not failing locally for me

### Where?
`utils/data_archive/tests/test_file_filter.py`

### How?
See https://github.com/ISISScientificComputing/autoreduce/actions/runs/372579810
### Reproducible?
[Yes]

keiranjprice101:
### Summary of work
From what I understand from reading the docs `os.listdir` will return in an arbitrary order. The test assumes the order will remain the same, the order is sometimes differing in the workflow vm. So instead we compare the lists ignoring order using a `Counter`. Since the order is irrelevant for the purpose of the function anyway (based on its docstring). 

### Additional comments
 

Fixes #931

### DIFF
--- a/utils/data_archive/tests/test_file_filter.py
+++ b/utils/data_archive/tests/test_file_filter.py
@@ -120,6 +120,7 @@ class TestTimeFilterFiles(unittest.TestCase):
             with open(file_path, 'w+') as new_file:
                 new_file.write("test_file")
         actual = filter_files_by_time(self.test_output_directory, cut_off)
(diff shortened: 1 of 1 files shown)


rickybassom (approved):
Nice find! Looks good :+1:

PR #941, opened 2020-11-23
Drop paths package

### LINKED ISSUE #754: Investigate if we can drop the paths package
Issue raised by: [developer]

### What?
The paths package appears to reimplement a small portion of the built-in and more widely used `pathlib` library while appearing to only be used 4 times. We should double-check all usages and possibly remove this package and just use `pathlib` instead, alternatively this could be implemented everywhere, but I don't see the benefit of this package over `pathlib`.

As far as I can tell from a quick look, the `paths` package offers this functionality:
- check if a path is a windows path
- add a separator to the end of a path e.g. `/home` -> `/home/
- split a path e.g. `/home/mydir/myfile` -> `[home, mydir, myfile]
- get the separator for a path e.g. `/home` -> `/`
- validate a path object based on some parameters set in the constructor of the object.

I have briefly searched for usages of this package (outside of this package) and have found 4 usages of the `append_path` function.

If these are the only usages I believe we should remove this entire package and use `pathlib` instead.

### Where?
`/paths`

### How?
I am guessing this came about as a way to have a nice API for dealing with all the filepaths, that are used within autoreduction. But it was never implemented everywhere.

Either paths will be removed and we should move to using path objects from `pathlib` or we should implement this everywhere.

keiranjprice101:
### Summary of work
Drops the paths package that has only 3 usages. We should continue to move to using `pathlib` whenever string paths are encountered within the actual code

### Additional comments
 

Fixes #754 


dtasev (approved):
I think this looks good - I can't see these paths used anywhere either.

Annoying the compute node production settings.py also uses `append_path` so at some point we'll break that - hopefully we remember to remove the `append_path` usage there.

Coverage is down (I didn't realise we had that working again!) but we are only removing things so shouldn't matter

keiranjprice101:
@dtasev Should be easy to fix on prod as, as soon as the paths are read in from settings they are converted to `Path` objects anyway.

PR #942, opened 2020-11-23
Use ReduceRun.delete() to delete a run, adds `--delete-all-versions` flag

### LINKED ISSUE #937: Manual remove does too much manual work?
Issue raised by: developer

### What?
It seems like manual remove is _literally_ manually removing all parts of the `ReductionRun` from the database - even though it seems like all the tables that have it as a foreign key also have `on_delete=models.CASCADE`. 

Is all that manual work unnecessary?
### Where?
https://github.com/ISISScientificComputing/autoreduce/blob/master/scripts/manual_operations/manual_remove.py#L107-L110

### Proposed fix
Just do `version.delete()` instead of the 4 lines that are there now

`manual_remove.py` should be fast

dtasev:
### Summary of work
- Tries to call the `.delete()` on the `ReducedRun` model. Everything else should get deleted via the on delete cascade.
	- It seems like this isn't always the case! Sometimes the DB raises an integrity error - as a workaround the code will revert to it's old behaviour then. This needs further investigation.

- Adds `--delete-all-versions` flag - specifying this means the script will no longer ask you to enter version range. It will just delete all versions for that run. 

### How to test your work
- Check that the related foreign key models aren't missing `on_delete=models.CASCADE` (I had a look but possible I missed something)
- This is a huge pain to test manually. The easiest (but not particularly easy) way that I recommend is to:
	- Run the script with a debugger attached. 
	- Breakpoint before or on `version.delete()` call
	- Run some of the previous queries in the debug terminal before the delete() gets executed
		- This should net you a QuerySet of some RunVariables / DataLocation (depending on which one you run)
	- Let the `delete()` execute
	- Run the previous query again - it should return you an empty QuerySet now

I checked that this happens while removing some runs from production today - it seems everything is being deleted as expected. If it wasn't deleting properly I'd also expect the DB to complain about consistency, but it might not if we don't have our relations set up correctly.

### Additional comments
 

Fixes #937

### DIFF
--- a/scripts/manual_operations/manual_remove.py
+++ b/scripts/manual_operations/manual_remove.py
@@ -46,7 +46,7 @@ class ManualRemove:
         self.to_delete[run_number] = list(result)
         return result
 
-    def process_results(self):
+    def process_results(self, delete_all_versions: bool):
         """
         Process all the results what to do with the run based on the result of database query
         """
@@ -56,7 +56,7 @@ class ManualRemove:
                 self.run_not_found(run_number=key)
             if len(value) == 1:
                 continue
-            if len(value) > 1:
+            if len(value) > 1 and not delete_all_versions:
                 self.multiple_versions_found(run_number=key)
 
     def run_not_found(self, run_number):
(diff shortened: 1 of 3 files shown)


dtasev:
I've tested this twice now to perform maintenance on the DB by removing a large amounts of run from each instrument. Haven't encountered any issues, and generally performs a lot faster. The workaround in b104737 can probably be removed once we fix our variable duplication.

PR #951, opened 2020-11-26
Fix contact us in footer

### LINKED ISSUE #679: Autoreduction Webapp - Fix contact us mailto links and Add "Contact Us" to footer
Issue raised by: developer

### What?
The Autoreduction mailto links currently don't work, we should fix this.

Currently, the Autoreduction Webapp does not include any functionality to contact the team for support quickly from all pages. We should add this to the footer.

We should ideally log/create a tally of questions asked so we can use the frequency of repeated questions asked directly as justification to add to the help page permanently under a commonly asked questions category. 
I suggest designing and then putting forward a process which we could use to do this for the whole team to review.

### Where?
* WebApp/autoreduce_webapp/templates/footer.html

### How?
To allow for easier accessibility to direct support from the Autoreduction team. 

* Contact us link works as expected (opens up email)
* Mutually agreed process to categorise the same or similar questions together to calculate a frequency from is put in place (this should be agreed upon by the whole team)

keiranjprice101:
### Summary of work
Creates custom context processor to make the support email available in all templates and mailto link functional

### How to test your work
Check that 
image
Is correctly displayed in all pages.

### Additional comments
 

Fixes #679 


WebApp/autoreduce_webapp/autoreduce_webapp/tests/test_context_processors.py
```diff
@@ -0,0 +1,25 @@
...
+class TestContextProcessors(TestCase):
+    """
+    Test cases for custom context processors
+    """
+    def test_support_email_processor(self):
+        """
+        Test: Support email dict returned
+        When: support email context processor called.
+        """
+        self.assertEqual({"support_email": "isisreduce@stfc.ac.uk"},
```
rickybassom, line 24: I'm not sure if this test would be useful. Can we ignore `support_email_processor` in test coverage since the function doesn't contain any logic? We are also repeating the same email address in two locations?

rickybassom (commented):
Code looks good :+1:. Just was't sure how to get the email to appear in the footer?

keiranjprice101:
@rickybassom the test setting needs to be applied

rickybassom:
> @rickybassom the test setting needs to be applied

Thanks, works well now

WebApp/autoreduce_webapp/autoreduce_webapp/tests/test_context_processors.py
keiranjprice101: Good shout, ill add this to the footer selenium tests though

rickybassom (approved):
The test looks good

PR #964, opened 2020-11-30
Use Ubuntu 16.04 for GH Actions

rickybassom:
### Additional comments
 

Fixes #xyz


rickybassom:
I think we should stick with 18.04 and test Autoreduction with it because support for 16.04 runs out in 2021-04.

PR #966, opened 2020-11-30
Use dash_core_components==1.12.1 for GitHub Actions

rickybassom:
### Summary of work
* Use dash_core_components==1.12.1 because of dependency clash seen [here](https://github.com/ISISScientificComputing/autoreduce/runs/1474250570):
```
The conflict is caused by:
    autoreduction 20.3 depends on dash_core_components==1.13.0
    dash 1.16.3 depends on dash-core-components==1.12.1
```

### How to test your work
* Tests pass

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ from build.commands.start import Start
 setup_requires = ['attrs==20.3.0',
                   'dash==1.16.3',
                   'dash_html_components==1.1.1',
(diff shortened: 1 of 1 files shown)


rickybassom:
@keiranjprice101 this seems to work with GH actions

keiranjprice101:
@rickybassom can you add a comment in the requirements just mentioning the clash

PR #1002, opened 2020-12-22
Removes variable duplication when submitting runs

### LINKED ISSUE #285: Remove data redundancy in the database

dtasev:
### Summary of work
Does a big refactor of `queue_processors/queue_processor/queueproc_utils/instrument_variable_utils.py`. The Web App copy of the same code will be handled in another PR

Effects of this PR:
Variables should no longer be duplicated.

### How to test your work
To test you'll have to deploy the migrated MySQL changes while running the DB locally. Then update your credentials accordingly.

Throughout this I recommend checking the number of rows in `InstrumentVariables` and `Variables`.

- Submit GEM 89880
	- Check that 5 variables were created as per it's reduce_vars.py
- Submit MARI 27858
	- Check that 10 variables have been created
- Submit GEM 89881
	- Check that no new variables have been created, but all have been re-used
- Change GEM's reduce_vars value, and resubmit
	- One new variable should have been created, containing the new value
	- Revert GEM's variable value
- Submit GEM 89881 again and MARI 27858
- Variable count shouldn't have changed, all should have been re-used

### Additional comments

Note: this requires a database migration due to the change in `RunVariable` (making `InstrumentVariable` a FK relation, to allow re-use).

Fixes #285


queue_processors/queue_processor/queueproc_utils/instrument_variable_utils.py
```diff
@@ -45,187 +51,124 @@ def log_error_and_notify(message):
...
+        variables.extend(
+            self.find_or_make_variables(possible_variables, run_number, instrument_id,
+                                        reduce_vars_module, 'advanced_vars', True))
+
+        logger.info('Creating RunVariables')
+        # Create run variables from these instrument variables, and return them.
+        return VariableUtils().save_run_variables(variables, reduction_run)
 
-        if not reduce_vars_module:
+    def find_or_make_variables(self, possible_variables, run_number, instrument_id,
```
keiranjprice101, line 92: - Can this take a reduction run and get the args instead from the run?
- Are the last 2 args necessary as (from what I can see within the PR) `find_or_make_variables` is only ever called twice and both within `create_variables_for_run`, meaning we could instead call this method once and have the list extension happen within the method and it returns both advanced and standard in one go?
This would reduce the number of args and be a lot clearer e.g. 
```python
def find_or_make_variables(self, possible_variables, reduction_run, reduce_vars_module):
```

keiranjprice101 (changes requested):
Very nice steps in the right direction. Just a few minor comments.
Also did you plan to look at the variable utils within the WebApp as well.

dtasev:
> Also did you plan to look at the variable utils within the WebApp as well.

Not in this PR. I think after #1022 would be a good time to merge the two variable utils

queue_processors/queue_processor/queueproc_utils/instrument_variable_utils.py
dtasev: I was passing stuff as arguments to avoid db hits, but I think everything is cached anyway on the `ReductionRun` object already. In any case I've already got the variables for the filter above so I'd rather use them. We can change that if we need more fields from the ReductionRun

dtasev:
Test failure might be due to `wait=False` on the QP connect, try setting it to `wait=True`

dtasev:
@keiranjprice101 Finally passes when I disabled a bunch of tests - the reason for that is that a lot of them will be refactored in #1022 so I don't want to spend a lot of time fixing them now, only to rewrite them in a day

queue_processors/queue_processor/queueproc_utils/script_utils.py
```diff
@@ -0,0 +1,79 @@
+# ############################################################################### #
+# Autoreduction Repository : https://github.com/ISISScientificComputing/autoreduce
+#
+# Copyright &copy; 2021 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+# ############################################################################### #
+"""
+Module that provides script utils
```
keiranjprice101, line 8: I'm conscious of the size of this PR growing, but I think this whole module should be combined into the `ReductionScript` class in `reduction_service.py` Though that could be opened as an issue after this and QP refactor

queue_processors/queue_processor/queueproc_utils/tests/test_script_utils.py
```diff
@@ -0,0 +1,56 @@
...
+# Copyright &copy; 2021 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+# ############################################################################### #
+import os
+from unittest.mock import Mock, patch
+from queue_processors.queue_processor.queueproc_utils.script_utils import import_module
+from queue_processors.queue_processor.queueproc_utils.tests.module_to_import import TEST_DICTIONARY
+
+
+def assert_raises(exc, func, args):
```
keiranjprice101, line 13: Can we not just use:
```python
with pytest.raises(ExceptionClass):
```
And remove custom assert.

As a side note we have a mixture of unittest testcase classes and just test modules written the pytest way. We should probably decide on one way and move towards making them all the same in the future. Not a big deal for now though.

keiranjprice101 (changes requested):
Looks good. 1 minor nitpick about assertions

queue_processors/queue_processor/queueproc_utils/script_utils.py
dtasev: I will look into it - they do look like they're doing the same thing, more or less.

queue_processors/queue_processor/queueproc_utils/tests/test_script_utils.py
dtasev: Thing is both unittest and pytest do some things better. As long as they can be used without interfering with each other I think a mix is OK. Though I lean towards pytest more and more

queue_processors/queue_processor/queueproc_utils/script_utils.py
dtasev: Actually on a second look I won't do it in this PR as the ReductionScript itself needs a small change to work with `reduce_vars.py`. I think it's fine for now but they'll need to be merged in.

dtasev:
@keiranjprice101 Let me know what you think on my responses

keiranjprice101 (approved):

Issue #915, opened 2020-11-11
Enforce code formatting repo-wide and add CI job


dtasev:
Issue raised by: developer

### What?
Currently there is no formatting style enforced. This can be annoying when you do a 1 line change, then save and the auto formatter changes another 150 lines!

If we had a preset code-style formatter this can be avoided.

My personal preference is YAPF, which is similar to C++'s clang-format in that it doesn't care about any previous formatting and reformats to the rules that are chosen. This is different from other formatters (e.g. PyCharm's formatter will sometimes respect manual formatting of arguments, etc). Overall I don't care whether or not we chose YAPF as long as we have one!

### Reproducible?
[Yes/No]

### How to test the issue is resolved
- [ ] Tool added to dependencies
- [ ] CI Job that enforces formatting and fails if formatting doesn't match

PR #1022, opened 2021-01-15
Queue processor refactor

### LINKED ISSUE #72: Queue Processor Logging
Issue raised by: Developer

### What?
Currently there's differences in how you setup logging for the queue processors. Since the queueprocessor and autoreductionprocessor are setup at the same time there should be some continuity in how/where logging settings are stored. 

### Where?
* [`Autoreduction processor logging`](https://github.com/ISISScientificComputing/autoreduce/blob/master/QueueProcessors/AutoreductionProcessor/autoreduction_logging_setup.py)
* [`Queue processor logging`](https://github.com/ISISScientificComputing/autoreduce/blob/master/QueueProcessors/QueueProcessor/test_settings.py)

* Ensure the new logging functionality is combined for both the Queue Processor and Autoreduction processor.
* Ensure some testing is done to make sure the right information goes to the correct log files

### LINKED ISSUE #74: Harmonise QueueProcessors
Issue raised by: Developer

### What?
On a similar note to #72, we should also harmonise the settings files for AutoreductionProcessor and QueueProcessor. Additionally we should look if there can be shared code between the two Processors with a view to simplify them.
This should not be attempted until there is more suffeicent testing for the QueueProcessors

### Where?
`Autoreduction Processor` & `Queue Processor`

* Verify that the Queue Processors still work as intended
* Code review for logic and simplification

### LINKED ISSUE #967: Queue Processor redesign
Issue raised by: develper

### What?
Discuss whether all 5 AMQ stages are necessary in the Queue Processor

It seems that not all 5 are required and there seems to be some unnecessary complication of the code in the queue processor (QP).

Currently it keeps moving messages in different queues, all while updating the DB on: `/DataReady, /ReductionStarted, /ReductionSkipped, /ReductionError, /ReductionComplete`.

It seems one of those updates is unnecessary: either keep the state to queues or keep it to the DB. Currently our state is mostly processed from the Queues and we get stale messages often. This is expected from Queues so it's probably better to move towards keeping the state in the DB

If we change the QP to have a simpler approach of:
- Get message from `/DataReady`
- Check if suitable for reduction - if so start a sub-processes performing the reduction
	- The subprocess is responsible for updating its state in the DB
	- All but `/DataReady` queues won't be used and can be removed

This change seems to be contained to the QP and ActiveMQ, and we don't necessarily even need to change the 2nd, but it would be good to clean up.

Pros:
- Less chance for losing job state - unless the subprocess magically disappears we can ensure that we have a job outcome
	- This should prevent stale jobs
- Potentially less DB hits - if possible to keep the job record reference we can queue the DB ONCE at the start, then just update state and save multiple times as we go through. 
	- So far it seems that saving a record whose reference we have is much cheaper than queueing for it again, especially when the DB grows
- This should make it easier to scale - especially if less DB hits turns out to be true
	- All we need then is multiple `/DataReady` consumers on different machines. This is also the case now, but I am not confident it will work well with the current arrangement, as we might just end up losing more messages!
- Should significantly reduce the code complexity in the QP package while doing the same thing

Time estimates:
- This doesn't look like more than a 2 week task, based on:
	- The rewrite itself is probably a couple of days or so, as most of it is a refactor
	- The testing will take more time, and we want to ensure the performance is OK
		- If possible to also test scalability that would be excellent, but I'm not putting that in the 2 week estimate

### LINKED ISSUE #1045: Extend ReductionScript to be able to load reduction_vars

### What?
`ReductionScript` has hard-coded `reduce.py` into the constructor. If it is extended to a parameter it can be used in `instrument_variable_utils` to load the reduce_vars module.

dtasev:
### Summary of work

This has grown quite a bit since the description below. At this point it basically rewrites the Queue Processor. Highlights of the works are:

#### `PostProcessAdmin` [PPA] is now called `ReductionRunner` [RR].
	- Added logging call when `ReductionScriptError` is encountered
	  - Communication between the PPA/RR still uses a Message object. The message is now written to a `tmpfile` from the subprocess. This file is read back into the RR and the Message is re-parsed.
		  - Perhaps this can be improved somehow, but the streams seem to be cluttered with a lot of output from the subprocess' reduce software. A tmpfile seemed the least code-heavy fix.

#### Removes the use of all queues but `/DataReady`. Now runs are processed as part of `on_message` handling. This is beneficial because:
- No chance of losing messages - once a message is consumed it is guaranteed to have _a_ processing outcome, unless the process gets killed or the system crashes
	- We can also scan the DB for hanging runs and rerun them
- Less DB hits. Very few (or even 0) DB reads after the initial `ReducedRun` & variables creation
	- A reference to the `ReducedRun` is kept so we do a few DB writes to update the state as it gets further into processing

#### Forces AMQ subscribe to prefetchSize=1, meaning that consumers will now only take 1 message off the queue
- With this change the ActiveMQ Dashboard's message dequeue time now actually has a real meaning: how fast are our consumers taking up the messages
	- This is a real statistic that we can use to gauge necessity for scaling up our processing nodes!
	- We can set real measurable targets for this depending on how responsive we want to be - e.g. at most 1 second before a message gets processed, if we go over this time, upscale the backend

- Due to allowing only 1 message at a time, the job queue might be less 'live' now, as there isn't going to be a big stack of "Queued" jobs sitting around. Arguably, whenever there was a big stack of queued jobs it meant that the DB has become too slow to process correctly, so this might be a good thing
	- This will also improve as we scale up the number of processing nodes. Because the process is now synchronous it means that 1 node = 1 reduction running at a time. Mantid usually assumes that it has the whole machine so this is a good thing, running 2 reduction simultaneously only slows everything down

The actual reduction is still done by the PPA/ReductionRunner

#### Queue Processor / Listener graceful exit
The queue listener daemon will now handle SIGTERM and SIGINT, and instead of stopping immediately will start a client disconnect. The disconnect waits until the current message is finished processing.

Also adds `kill-dangerous.sh` to use the old behavours. It's dangerous because killing a reduction will leave the ReductionRun object hanging in Processing or Queued or whatever state it got to.

#### Line length was changed to 120 in `.pylintrc`. 
A lot of formatting changes are there because we don't enforce a formatter as an Action PR check. That will change in #915 

#### LOGGER config is not moved to `queue_processor/__init__.py` 
- This allows easy enforcement throughout the whole QP module, but we still need another change for repo-wide logging configuration.

#### ReduceScript now takes a file name parameter
- And it's used in create variables
- Fixes #1045

<hr>

### How to test your work

Before each:
- Run MySQL locally and migrate it to the latest migrations
	- `sudo mysql < build/database/reset_autoreduction_db.sql; and python setup.py database`
- Have ActiveMQ running (`sudo python setup.py start`)

#### Running the queue_processors tests and systemtests
- `pytest queue_processors/`
- `pytest systemtests`

#### Running the full stack
- The `systemtests` do this, but you can also run the queue listener manually and submit some runs with manual_submission
	- Note that on your dev env the reductions will fail if you don't have the archive at /isis and Mantid installed
	- Failing the reductions is OK - just check the variable count and that they show up in the webapp at all
- Run the Queue Listener locally - I recommend on a new terminal run `python queue_processors/queue_processor/queue_listener.py`. It doesn't need to be run through the daemon
- Change the `credentials.ini` ActiveMQ creds to:
```user=admin
password=admin
host=127.0.1.1
port=61613
```
- Manually submit a run - the terminal that is running queue listener should start printing stuff out.
- Also keep an eye on the logs with `tail -f logs/queue_processor.log` - it should start printing stuff

### Additional comments
This PR has become pretty chunky and rewrites most of the queue processor. The PPA/RR stuff is only marginally refactored (e.g. taking sending messages out)

Fixes #967 
Fixes #1045
Fixes #74
Fixes #72 

### Follow up issues
#1042
#1073


queue_processors/queue_processor/reduction/runner.py
```diff
@@ -0,0 +1,154 @@
...
+from queue_processors.queue_processor.reduction.exceptions import DatafileError, ReductionScriptError
+from queue_processors.queue_processor.reduction.utilities import windows_to_linux_path
+from queue_processors.queue_processor.reduction.service import (Datafile, ReductionDirectory, ReductionScript,
+                                                                TemporaryReductionDirectory, reduce)
+from queue_processors.queue_processor.settings import MANTID_PATH, TEMP_ROOT_DIRECTORY
+
+logger = logging.getLogger("reduction_runner")
+
+
+class ReductionRunner:
```
dtasev, line 28: This is just the PPA, but with all AMQ messaging ripped out

queue_processors/queue_processor/tests/test_handle_message.py
```diff
@@ -9,492 +9,360 @@
...
+from functools import partial
+import random
 import unittest
 from unittest import mock
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
-from model.database import access
+from django.db.utils import IntegrityError
+from parameterized import parameterized
```
keiranjprice101, line 19: Is this a new dependency, it shows as not available for me and if so this will need to go into the `requirements.txt`

queue_processors/queue_processor/reduction/tests/test_process_manager.py
```diff
@@ -0,0 +1,66 @@
...
+
+        def side_effect(args, check):
+            raise CalledProcessError(1, args)
+
+        subprocess_run.side_effect = side_effect
+        rpm = ReductionProcessManager(self.message)
+        rpm.run()
+
+        subprocess_run.assert_called_once()
+        assert "Processing encountered an error" in rpm.message.message
```
keiranjprice101, line 52: Can we use `self.assertIn` and `self.assertEquals` for unittest test cases?

keiranjprice101 (changes requested):
This is a lot easier to follow, and a lot more sane. 👍 

I have some minor nitpicks and still don't like `HandleMessage` as a whole but this is looking really promising.

Still need to look through the rest of the tests.

queue_processors/queue_processor/queue_listener.py
```diff
@@ -10,74 +10,81 @@
...
+            # the connection is configured with client-individual, meaning that each client
+            # has to submit an acknowledgement for receiving the message
+            # (otherwise I think that it is not removed from the queue but I am not sure about that)
+            self.client.ack(headers["message-id"], headers["subscription"])
+            try:
+                if destination == '/queue/DataReady':
+                    self.message_handler.data_ready(message)
+                else:
+                    self.logger.error("Received a message on an unknown topic '%s'", destination)
+            except InvalidStateException as exp:
```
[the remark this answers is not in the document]
dtasev: Good catch, that exception is never raised in the calls below so I don't think we'll ever capture it

queue_processors/queue_processor/handle_message.py
```diff
@@ -37,201 +36,179 @@ class HandleMessage:
...
+        skip_reason = self.should_skip(reduction_run, message, instrument)
+        if skip_reason is not None:
+            message.message = skip_reason
+            self.reduction_skipped(reduction_run, message)
+        else:
+            instrument = self.activate_db_inst(instrument)
+            self.do_reduction(reduction_run, message)
+
+    @staticmethod
+    def should_skip(reduction_run, message: Message, instrument) -> Optional[str]:
```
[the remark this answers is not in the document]
dtasev: `find_reason_to_skip_run`? 😄

queue_processors/queue_processor/tests/test_handle_message.py
dtasev: It's added but you'll have to rerun `pip install -r requirements.txt`

queue_processors/queue_processor/reduction/tests/test_process_manager.py
dtasev: TBH I find just the `assert` much easier to read, and with pytest as a runner they have the same behaviour I think

keiranjprice101: Fair enough

model/database/access.py
```diff
@@ -20,25 +20,19 @@ def start_database():
     return database
 
 
-def get_instrument(instrument_name, create=False):
+def get_instrument(instrument_name):
     """
     Find the instrument record associated with the name provided in the database
     :param instrument_name: (str) The name of the instrument to search for
     :param create: (bool) If True, then create the record if it does not exist
```
Anders-Markvardsen, line 27: please delete this line since `create` been removed

Anders-Markvardsen: And similarly where delete parameter elsewhere

model/database/access.py
```diff
@@ -104,30 +89,21 @@ def get_reduction_run(instrument, run_number):
...
+    return database.data_model.ReductionRun.objects.filter(instrument=instrument_record.id, run_number=run_number)
 
 
-def find_highest_run_version(experiment, run_number):
+def find_highest_run_version(experiment, run_number) -> int:
     """
     Search for the highest run version in the database
     :param experiment: (str) The experiment number associated
     :param run_number: (int) The run number to search for
     :return: (int) The highest known version number for a given reduction job
```
Anders-Markvardsen, line 100:
```suggestion
    :return: (int) The highest known run version for a given run number
```

model/message/validation/validators.py
```diff
@@ -23,14 +22,6 @@ def validate_run_number(run_number):
...
-    """
-    Assert an instrument exists in the valid set of instruments
-    :param instrument: The instrument to validate
-    """
-    return isinstance(instrument, str) and instrument in VALID_INSTRUMENTS
-
-
 def validate_rb_number(rb_number):
     """
     Detects if the RB number is valid e.g. (above 0 and not a string)
```
Anders-Markvardsen, line 27:
```suggestion
    Detects if the RB number takes the format RBxxxxxxx where xxxxxxx is the 7-digit RB number
```

queue_processors/queue_processor/queue_listener.py
```diff
@@ -95,7 +98,7 @@ def setup_connection(consumer_name):
 
     # Subscribe to queues
     activemq_client.subscribe_autoreduce(consumer_name, listener)
-    return activemq_client
+    return activemq_client, listener
```
Anders-Markvardsen, line 101: please update `:return:` accordingly, or using `->`

queue_processors/queue_processor/reduction/process_manager.py
```diff
@@ -0,0 +1,49 @@
+# ############################################################################### #
+# Autoreduction Repository : https://github.com/ISISScientificComputing/autoreduce
+#
+# Copyright &copy; 2020 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+# ############################################################################### #
+"""
+Module that reads from the reduction pending queue and calls the python script on that data.
```
Anders-Markvardsen, line 8: Is `reads from the reduction pending queue` correct?

queue_processors/queue_processor/reduction/runner.py
```diff
@@ -0,0 +1,152 @@
...
+# Autoreduction Repository : https://github.com/ISISScientificComputing/autoreduce
+#
+# Copyright &copy; 2020 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+# ############################################################################### #
+# !/usr/bin/env python
+# pylint: disable=broad-except
+# pylint: disable=bare-except
+"""
+Post Process Administrator. It kicks off cataloging and reduction jobs.
```
Anders-Markvardsen, line 11: Should all references to `Post Process Administrator` and `Post Process admin` be removed. The name of this file is `runner.py` perhaps a good name, but the connection it name to comment `It kicks off cataloging and reduction jobs` is not clear to me

utils/clients/queue_client.py
```diff
@@ -70,71 +69,55 @@ def _create_connection(self, listener=None):
...
             self._connection = connection
 
-    def subscribe_queues(self, queue_list, consumer_name, listener, ack='auto'):
+    def subscribe_queues(self, queue_list, consumer_name, listener):
         """
         Subscribe a listener to the provided queues
         :param queue_list: The queue(s) to subscribe to as a string or list of strings
         :param consumer_name: A name to assign to the consumer
         :param listener: A ConnectionListener object to assign to the stomp connection, optionally
         :param ack: The acknowledge type
```
Anders-Markvardsen, line 89: please delete :param ack:

Anders-Markvardsen (changes requested):
tested and wfm

A lot of work! And very good.

Some small suggested for modifications, good with follow up issues, and with this big change good with time to operate before the April cycle:)

model/message/validation/validators.py
dtasev: I agree the original is now wrong, but I think the suggestion is a bit confusing. Maybe `Detects whether the RB number is a 7-digit number`?

keiranjprice101 (approved):

Anders-Markvardsen (approved):

PR #1026, opened 2021-01-19
Fix adding signing key comes before adding repo

RebeccaBecky:
### Summary of work
Swapped the order of two lines. Before, adding the repo came before adding the mantid signing key. Now, adding the signing key comes before adding the repo.

### How to test your work
Test by installing mantid:
sudo python setup.py externals -s mantid

### Additional comments

### DIFF
--- a/build/install/mantid.sh
+++ b/build/install/mantid.sh
@@ -10,10 +10,11 @@ fi
 
 set -e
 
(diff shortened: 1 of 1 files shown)


rickybassom:
This works without any issues on a fresh install of Ubuntu 18.04. Tested yesterday when setting up WSL.

PR #1030, opened 2021-01-20
Fixes retrieving plots from CEPH

dtasev:
Instead of joining the paths it was just appending, leading to malformed URI
such as: `[20/Jan/2021 16:25:05] ERROR [app:127] File '/instrument/MARI/RBNumber/RB2000159/autoreducedMAR27877_Ei40.10meV.png' does not exist` - we've dropped a `/` somewhere in #793 

### Summary of work
Just adds the dropped `/`. 

`os.path.join` doesn't work as it's on Windows so it makes the path like this: `'/instrument/MARI/RBNumber/RB2000159/autoreduced\MAR27877_Ei40.10meV.png'`

### How to test your work
Hard one to test as this only happens when on a Windows machine and accessing the SFTP client which is Linux. Code review should be enough

### DIFF
--- a/plotting/plot_handler.py
+++ b/plotting/plot_handler.py
@@ -115,7 +115,7 @@ class PlotHandler:
             client = SFTPClient()
             for plot_file in _existing_plot_files:
                 # Generate paths to data on server and destination on local machine
-                _server_path = self.server_dir + plot_file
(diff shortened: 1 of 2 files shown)


keiranjprice101 (approved):

rickybassom (approved):
Looks good :+1:

PR #1067, opened 2021-02-03
Display more information in admin page

### LINKED ISSUE #898: Increase usability of the  Admin WebApp tab
Issue raised by: developer

### What?
The admin gives full view in Autoreduction database entries. But lacks easy way to see what these are. Ideally also add better search functionality, although this may be done in a separate issue.

A suggestion is to add more columns (but better options may be available). For example for Notifications:

image

which as can be seen at present needs to entry each of these to what the notification actually are!

keiranjprice101:
### Summary of work

Adds str dunders to  models meaning the admin page will show more useful text

e.g.

image

### How to test your work
View the admin page and check the models show their name instead of X Object (x)

Fixes #898

### DIFF
--- a/WebApp/autoreduce_webapp/reduction_viewer/models.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/models.py
@@ -27,7 +27,7 @@ class Instrument(models.Model):
         """
         :return: str representation of instrument
         """
-        return f"{name}"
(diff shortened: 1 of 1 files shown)


rickybassom (approved):
Looks good. This might simplify some things elsewhere in the code too.

PR #1070, opened 2021-02-03
Remove populate_reduction_viewer, test_db_generation, DatabaseClient and mandatory MySQL backend

### LINKED ISSUE #758: Drop SQLAlchemy dependency
Issue raised by: [developer]

### What?
SQLAlchemy is only used in the setup. As part of #730, #736 we should look to drop this dependency 

### Where?
`utils/clients/database_client.py`

### How?
SQLAlchemy is used to create a test database, but we should just use Django/containerised DB

### Reproducible?
[Yes]

SQLAlchemy no longer a dependency

### LINKED ISSUE #1066: Remove test_db_generation and populate_reduction_viewer.sql - use text fixtures instead

### What?
test_db_generation is a test to see whether `populate_reduction_viewer.sql` was executed properly. It is a bad way of doing a global database fixture that _some_ of the tests depend on.

Remove this global fixture and adjusts the tests that use it to themselves seed the database with the required data.

### LINKED ISSUE #1072: Remove DatabaseClient (SQLAlchemy)

### What?
In our tests we have an annoying mixture of SQLAlchemy usage via DatabaseClient and Django ORM. This means that we have a duplication of how the database can be accessed, but most importantly prevents us from using a simpler DB backend for testing - sqlite3

Pick either SQLAlchemy or Django ORM and stick with it for the apps outside of the Django Web App.

The DatabaseClient usage doesn't seem to be that much - mostly in tests, so the removal should be fairly simple.

dtasev:
*Should be merged after #1022*

### Summary of work
- Removes populate_reduction_viewer.py from `python setup.py database`.
- Removes test_db_generation.
- Removes DatabaseClient that uses SQLAchemy, and seems to be unused
- Removes MySQL as the default backend and makes it SQLite3
- Adjusts tests to not expect initial data, and instead prepare the DB objects in setUp and cleanup in tearDown

### How to test your work
Tests should pass.

Fixes #1066
Fixes #1072
Fixes #758


build/database/tests/test_generate_database.py
```diff
@@ -1,63 +0,0 @@
...
-
-# pylint:disable=missing-docstring
-def raise_operational_error(_):
-    raise OperationalError(statement='Fail',
-                           params=None,
-                           orig=None)
-
-
-# pylint:disable=missing-docstring
-class TestGenerateDatabase(unittest.TestCase):
```
dtasev, line 29: The functions tested here have been deleted as they're no longer used and the one that's left over in generate_database.py doesn't have a test

WebApp/autoreduce_webapp/autoreduce_webapp/settings.py
```diff
@@ -182,17 +187,11 @@ def get_str(section, key):
...
 ACTIVEMQ = {
-    'topics': [
-        '/queue/DataReady', '/queue/ReductionStarted', '/queue/ReductionComplete',
-        '/queue/ReductionError'
-    ],
-    'username':
-    get_str('QUEUE', 'user'),
-    'password':
-    get_str('QUEUE', 'password'),
+    'topics': ['/queue/DataReady', '/queue/ReductionStarted', '/queue/ReductionComplete', '/queue/ReductionError'],
```
keiranjprice101, line 191: Queues can go now?

dtasev: Good catch - done

keiranjprice101 (approved):

PR #1075, opened 2021-02-05
Adds YAPF to requirements and as a CI step

### LINKED ISSUE #915: Enforce code formatting repo-wide and add CI job

dtasev:
### Summary of work
Adds YAPF to requirements and as a CI step

### How to test your work
Tests should pass

Fixes #915

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -25,6 +25,9 @@ jobs:
 
       - uses: ./.github/actions/build-autoreduction
 
+      - name: YAPF
+        run: python -m yapf --exclude '**/migrations' --parallel --diff --recursive .
+
       - name: Run unit tests
         run: >-
           pytest --ignore=systemtests --cov=build --cov=monitors --cov=queue_processors
--- a/.unibeautifyignore
+++ b/.unibeautifyignore
@@ -1,0 +1,1 @@
+**/migrations
--- a/.unibeautifyrc.json
+++ b/.unibeautifyrc.json
@@ -1,0 +1,8 @@
+{
+    "Python": {
+        "beautifiers": [
+            "YAPF"
+        ],
+        "wrap_line_length": 120
+    }
+}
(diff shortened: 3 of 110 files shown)


keiranjprice101:
Possibly wishful thinking, but is it possible to add an action that will run yapf and commit the output ?

dtasev:
I don't mind that tbh - do you think it might be annoying if you decide to add more changes but then you can't push & have to rebase/merge the formatting changes?

keiranjprice101:
You can just pull the formatting change and continue from there no?

dtasev:
Yeah but if you don't notice and continue working after pushing, then at some point you might get a lot of conflicts. Perhaps some bot would be better, i.e. if you comment @format or something it just does it and commits?

keiranjprice101:
That could be nice, especially if we can combine it with a branch protection rule that wont allow merging until the `@Format` has been triggered (Or whatever the manual trigger event is

rickybassom:
You can have a look at this [code](https://github.com/ISISScientificComputing/autoreduce/blob/2e8e6bcc4b8284fa0077c4500a3b422f2e3f7547/.github/workflows/codeql-analysis.yml#L17) I used to trigger code scanning by commenting on a PR. In the end I decided to go for a `workflow_dispatch` event that can be triggered from Actions > Code Scanning > Run Workflow. 

I suppose both methods would work for triggering a YAPF action.

dtasev:
Hmm I did some looking around and https://unibeautify.com/ looks like exactly what we were discussing above. I will try to integrate it in to this repo.

dtasev:
Unibeautify should provide that ability to merge the formatted changes, however it doesn't seem to be working right now and I'm trying to get some help from the CI maintainers at https://github.com/Unibeautify/ci-issues/issues/5

In the meantime I've added back the YAPF CI job and I don't think this should hold up this PR

PR #1086, opened 2021-02-08
Modify STATICFILES_DIRS so web app can be started from any directory

RebeccaBecky:
### Summary of work

Changed STATICFILES_DIRS to 'BASE_DIR, static' so that the web app can be run from any directory (rather than having to run from autoreduction/WebApp/autoreduce_webapp)

### How to test your work
 

- [ ] `cd autoreduction/WebApp/autoreduce_webapp` then `python manage.py runserver` then check web app works correctly
- [ ] `cd autoreduction` then `python WebApp/autoreduce_webapp/manage.py runserver` then check web app works correctly

### Additional comments
 

Fixes #xyz

### DIFF
--- a/WebApp/autoreduce_webapp/autoreduce_webapp/settings.py
+++ b/WebApp/autoreduce_webapp/autoreduce_webapp/settings.py
@@ -139,7 +139,7 @@ USE_TZ = True
 STATIC_URL = '/static/'
 STATIC_PATH = os.path.join(BASE_DIR, 'static')
 STATICFILES_DIRS = [
-    os.path.join('BASE_DIR, static'),
+    os.path.join(BASE_DIR, 'static'),
 ]
 
 # Logging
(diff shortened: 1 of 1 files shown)


WebApp/autoreduce_webapp/autoreduce_webapp/settings.py
```diff
@@ -139,7 +139,7 @@ def get_str(section, key):
 STATIC_URL = '/static/'
 STATIC_PATH = os.path.join(BASE_DIR, 'static')
 STATICFILES_DIRS = [
-    os.path.join('static'),
+    os.path.join('BASE_DIR, static'),
```
dtasev, line 142:
```suggestion
    os.path.join(BASE_DIR, 'static'),
```
Just a small typo because the ' was making this whole thing a string!

dtasev (changes requested):

dtasev (approved):
Works well!

PR #1092, opened 2021-02-09
Adds Web App migrations

### LINKED ISSUE #1068: Unignore DB migrations

### What?
The migrations folder is currently ignored - this is not great as it means we can never migrate from state 1 to state 2, and to apply a schema change we must delete the whole database.

This hasn't been an issue so far because we delete the dev/prod DBs at the end of a cycle anyway, but if that is going to change we'll have to keep the migrations in

dtasev:
### Summary of work
There's no reason not to have the migrations in the app - this avoids having to make them with `makemigrations` every time, but now you can just migrate all the apps.

This is also a step towards not having to wipe the DB on schema change, but instead migrate the data so that we keep the production DB alive for longer. But we'll see how next cycle goes before touching that issue further.

### How to test your work
Nothing really, just comment if there's any reason not to have the migrations in the codebase.

Fixes #1068


keiranjprice101 (approved):
Makes sense to me and clearly intended.
From Django doc:
> The migration files for each app live in a “migrations” directory inside of that app, and are designed to be committed to, and distributed as part of, its codebase.

Issue #987, opened 2020-12-16
Add missing page tests


keiranjprice101:
Issue raised by: [developer/user/project requirement]

### What?
We need to add the missing tests for other pages. And as suggested ideally everyone should do some tests for at least one page.

The pages that need doing.
- [x] Run summary page
- [x] Rerun Page #1042
- [x] Job queue page #1042
- [x] Configure new runs page #1042
- [x] See instrument variables page #1042
- [x] Help Page 
- [ ] Graphs (maybe)

Tick off as we go.

### Where?
`webtests/`

dtasev:
I'll do the rerun page as I'm having a look at it now for #1042, though I'll cover it as both view and selenium tests

rickybassom:
I'll do the help page

rickybassom:
I think we still need the Graphs tests

Issue #857, opened 2020-10-21
Rerun past jobs  & configure new runs will produce an error if only skipped runs exist


keiranjprice101:
Issue raised by: [developer]

### What?
This is an edge case, but currently observable on LARMOR.
If there exists <em>only</em> skipped runs for an instrument. Clicking Re-run past jobs on the instrument summary page will produce an error as django will fail to find the correct URL to go to.

### Where?
Error can be triggered from `WebApp/autoreduce_webapp/templates/run_summary.html`

The relevant code is in `WebApp/autoreduce_webapp/instrument/views/runs.py::configure_new_runs_GET`, specifically the `except AttributeError`

### How?
Smoke testing

### Reproducible?
[Yes]

### How to test the issue is resolved
Either hide the button in this case or change the way the url for this button is generated.

Issue #778, opened 2020-09-23
Display reduction variables in the rerun page


keiranjprice101:
Issue raised by: [developer]

### What?
It should be possible to see the current reduction variables in the rerun page. It would also be a good idea to veirfy  whether the "reset variables" "additional action" is working, and reformat it so the help is displayed. (similar to #775)
### Where?
`WebApp/autoreduce_webapp/templates/submit_runs.html`

### Reproducible?
[Yes]

### How to test the issue is resolved
The reduction variables should be visible and the reset variable action should be reformatted.

dtasev:
Isn't this the default behaviour - the latest reduction variables' values is shown? Whether reset variables works is a good question though

dtasev:
I'll close this until requested by users

PR #1094, opened 2021-02-09
Refactor WebApp variable creation

### LINKED ISSUE #739: Duplicated Code
Issue raised by: [developer]

### What?
There is a lot of duplicated code throughout autoreduction. The biggest culprits I found are:

- ~~`WebApp/autoreduce_webapp/autoreduce_webapp/daemon.py` - `queue_processors/daemon.py`~~
- `WebApp/autoreduce_webapp/reduction_variables/utils.py` - `queue_processors/queue_processor/queueproc_utils/variable_utils.py` (`convert_variable_to_type`) in both
- `WebApp/autoreduce_webapp/reduction_variables/utils.py` - `queue_processors/queue_processor/queueproc_utils/instrument_variable_utils.p` (large portion of `create_variables_for_runs()` and `create_variables_for_run()`

I am sure more exist though. Might not be the best idea to tackle all at once.

### Where?
The Big ones are listed above.

### Reproducible?
[Yes]

Functionality unchanged and tests pass

### LINKED ISSUE #852: Software is blank when attempting to rerun through WebApp
Issue raised by: [developer]

### What?
In #851 the software field on a `ReductionRun` was set to allow blank values to allow reruns through the WebApp
We need to identify why the software was never set previously and reenable this non blank constraint.

### Where?
Constraint is set here `WebApp/autoreduce_webapp/reduction_viewer/models.py`
Run is triggered in the WebApp here `WebApp/autoreduce_webapp/reduction_viewer/utils.py`

### How?
Smoke Testing

Non blank software constraint should remain and it should be possible to resubmit runs via the WebApp

### LINKED ISSUE #938: Production webapp log is in the wrong place

### What?
The log should be in `Autoreduction/logs`

### Where?
Seems to be in `Autoreduction/WebApp/autoreduce_webapp/autoreduction.log`

### LINKED ISSUE #1098: Remove StatusUtils duplication from WebApp

### What?
What is the issue

### Where?

### How?
How did the issue come about

### Reproducible?
[Yes/No]

### LINKED ISSUE #1108: Remove REDUCTION_DIRECTORY variable

### What?
What is the issue

### Where?

### How?
How did the issue come about

### Reproducible?
[Yes/No]

### LINKED ISSUE #1112: Variables not displayed on runs/<instrument>/<run>/<version>/

### What?
What is the issue

### Where?

### How?
How did the issue come about

### Reproducible?
[Yes/No]

### LINKED ISSUE #1119: `runs:summary` needs a way back to `runs:list`

### What?
There isn't a path back from the page for summary of a run to the runs list.

A simple back button will suffice

dtasev:
#### Refactors the WebApp variable creation 

Refactors the WebApp variable creation code to re-use the new queue processor (and the real code, rather than a copy-pasted version of it!). This removes a lot of the utils. Changes touch these parts:
- Rerunning from a run summary
- Rerunning past jobs
- Configuring new jobs
- Seeing instrument variables

#### Rewrites rerun approach
Previously the rerun created ReductionRun object in the DB (via the copy pasted QP code), then submitted to the `/ReductionPending` queue. That approach was removed due to that queue not being used anymore, and to reduce the code duplication.

The new approach just submits the message to the `/DataReady` queue. 

There is a drawback here - it may not be obvious to the user whether the Re-run has been submitted, as the re-run might not show up immediately after it has been sent. This can happen if all workers are busy, and is a known limitation of the 1 queue design.

This might need an additional message in the `run_confirmation.html` or a restructuring of how the information is shown page.

#### Renames `reduction_variables` app to `instrument`
All the URLs in the ex-`reduction_variables` were in the namespace `instrument`, so it made sense that the whole thing be named instrument.

This makes for some nicer URL names such as `instrument:submit_runs`, `instrument:variables` etc.

Note the `instrument:` syntax, for when referring to a URL included from an app.

There's a few more bad URL names left in the `reduction_viewer`, but the views were too entangled to deal with.

----------
TODOs

- [x] #1098
- [x] ~~Investigate why creating experiment variables makes ALL variables, rather than just the changed ones~~
	- Left unchanged, because actually, that makes sense. When you set variables for the whole experiment it will make all variables again, some duplication but only once per experiment number, which shouldn't cause issues. We can look into it again in the future if it does
- [x] ~~"Configure New Runs" doesn't put the correct defaults when editing an experiment number/start run~~
	- This will be tackled in a follow up issue  #1133
- [x] View & selenium tests for re-run page
	- [x] Run summary (only left is sending an actual rerun)
	- [x] Rerun jobs page, part of #987
	- [x] Configure new jobs
	- [x] See instrument variables

- [x] #778
- [x] use_current_script = False path, done by removing that whole thing. The script is ignored by the reduction runner. This is captured in #1115 if we want the functionality back.
- [x] #1108

### How to test your work
The hardest part of testing this PR is setting up the env!

Before testing anything you will have to migrate the DB (`python setup.py database`) and install the basic fixtures (`python WebApp/autoreduce_webapp/manage.py loaddata super_user_fixture status_fixture two_runs` or `python setup.py fixtures`).

Make a `reduce_vars.py` for the test instrument with:
```
mkdir -p data-archive/NDXTestInstrument/user/scripts/autoreduction
echo 'standard_vars = {"variable1":"value1"}' > data-archive/NDXTestInstrument/user/scripts/autoreduction/reduce_vars.py
```

Then run ActiveMQ locally with `sudo python setup.py start` - you need to be admin for this.

Check the `credentials.ini` and make sure all is pointing locally, you can just use the values from the `test_credentials.ini` I think.

Once you've got the DB & fixtures installed you can run the queue processor & Web App, by:
- Going to `Run` tab (CTRL+SHIFT+D default keybind) in VSCode and starting the `Queue Listener` and then `Web app` - all you need to do afterwards is CTRL-click the `http://127.0.0.1:8000/` in the terminal

if not using VSCode

- Start QP with `python queue_processors/queue_processor/queue_listener.py` - this starts it without a daemon
- Start webapp with `python WebApp/autoreduce_webapp/manage.py runserver`

#### Instructions for `Rerun from Run summary`
- Navigate to the instrument runs list page
- Click on run
- Try the reset to defaults button
- Rerun it with different parameters
- See what runs come up, and with what variable values.
	- If nothing new comes in check that the QP is running, and what Queue it's been connected too.

#### Instructions for `Rerun from Rerun page`
- Similar to rerunning from run summary, but try rerunning a range of runs

#### Instructions for `Configuring new Runs` & `See instrument variables`
- You might want to use a real instrument here (e.g. MARI 27877 run is good, 27878)
	- The runs don't have to Complete, they just need to not be Skipped!
	- This means you can submit any runs
- Configure a bunch of variables for a future run range
- Then run a run inside that run range and check that they are used correctly
- Repeat for experiment number

--------

Fixes #938
Fixes #739
Fixes #852
Fixes #1098
Fixes #1112
Fixes #1119
Fixes #1108 

### Follow ups / known issues
#1133
https://github.com/ISISScientificComputing/autoreduce/issues/1154
#857


Anders-Markvardsen:
Started testing and done

```
pip uninstall Autoreduction
pip install Autoreduction
python setup.py database
python WebApp/autoreduce_webapp/manage.py loaddata super_user_fixture status_fixture two_runs
```

Unhappy about the last step and error `ModuleNotFoundError: No module named 'utils'`

WebApp/autoreduce_webapp/selenium_tests/pages/page.py
```diff
@@ -37,5 +37,9 @@ def launch(self):
         Navigate the webdriver to this page
         :return: The page object
         """
+        self.driver.get(configuration.get_url())
         self.driver.get(self.url())
         return self
+
+    def get_id(self, id_str: str):
```
keiranjprice101, line 44: Doesnt appear to be ever used.

build/commands/database.py
```diff
@@ -41,3 +41,19 @@ def run(self):
...
+
+    def initialize_options(self):
+        pass
+
+    def finalize_options(self):
+        pass
+
+    def run(self):
+        """ Run the setup scripts required for localhost database """
+        # pylint:disable=import-outside-toplevel
```
keiranjprice101, line 58: Is this disable needed?

Anders-Markvardsen:
Observed the following issue when clicking on a run `A server error occurred.  Please contact the administrator.`

Anders-Markvardsen:
Also observe this one

webapp_bug

dtasev:
> Also observe this one
> 
> webapp_bug
I've changed how a missing `reduce_vars` is handled, not much can be done without one, so now the buttons are disabled:

image

I think the views themselves handle that case OK

Anders-Markvardsen:
Following approximately these steps in venv

```
pip uninstall Autoreduction
pip install Autoreduction
python setup.py database
python setup.py fixtures
python WebApp/autoreduce_webapp/manage.py loaddata two_runs
mkdir -p test-archive/NDXTestInstrument/user/scripts/autoreduction
echo 'standard_variables = {"variable1":"value1}' > test-archive/NDXTestInstrument/user/scripts/autoreduction/reduce_vars.py
export PYTHONPATH=$PYTHONPATH:/home/anders/autoreduction/
python WebApp/autoreduce_webapp/manage.py runserver
```

`reduce_vars.py` still not picked up. Have tried with `test-archive` and `data-archive` and with ref to 

```
if "RUNNING_VIA_PYTEST" in os.environ:
    ARCHIVE_ROOT = os.path.join(PROJECT_ROOT, 'test-archive')
else:
    ARCHIVE_ROOT = os.path.join(PROJECT_ROOT, 'data-archive')

# Variables that get changes less
# %(instrument, cycle, experiment_number, run_number)
CYCLE_DIRECTORY = os.path.join(ARCHIVE_ROOT, 'NDX%s', 'Instrument', 'data', 'cycle_%s')
SCRIPTS_DIRECTORY = f"{ARCHIVE_ROOT}/NDX%s/user/scripts/autoreduction/"
```

in  `queue_processors/queue_processor/settings.py`. 

Is PRODUCT_ROOT just within the git folder?

Ran out of time now, will come back later

WebApp/autoreduce_webapp/selenium_tests/tests/test_run_summary_integration.py
```diff
@@ -0,0 +1,187 @@
...
+        Submit after a reset button has been clicked.
+
+        Sticks the submission in a loop in case the first time doesn't work. The reason
+        it may not work is that resetting actually swaps out the whole form using JS, which
+        replaces ALL the elements and triggers a bunch of DOM re-renders/updates, and that isn't fast.
+        """
+        while str(self.run_number) in self.driver.current_url:
+            # NOTE that we MUST wait first and then try to submit, otherwise the processing
+            # may be finished before wait_for_result is called, causing it to be stuck forever
+            time.sleep(1)
```
keiranjprice101, line 99: Can you use a selenium wait until expected condition here instead of a sleep?

WebApp/autoreduce_webapp/selenium_tests/tests/test_rerun_jobs_range_integration.py
```diff
@@ -0,0 +1,149 @@
...
+
+        Sticks the submission in a loop in case the first time doesn't work. The reason
+        it may not work is that resetting actually swaps out the whole form using JS, which
+        replaces ALL the elements and triggers a bunch of DOM re-renders/updates, and that isn't fast.
+        """
+        page_url = self.page.url()
+        while page_url in self.driver.current_url:
+            # NOTE that we MUST wait first and then try to submit, otherwise the processing
+            # may be finished before wait_for_result is called, causing it to be stuck forever
+            time.sleep(1)
```
keiranjprice101, line 95: Can this use a selenium wait for expected condition?

keiranjprice101 (changes requested):
I have only briefly looked over the selenium part since it seems you are still making a lot of changes. But initial comments below

A majority of the selenium tests are not using page object model pattern correctly.  There are places in tests where there are calls to the webdriver that interact with the page.

e.g. calls to 
`.is_displayed()`
`.text`
`.find_element_by_x()`
`.click()`

These tpes of calls should be in page methods. This means as pages change in the future we only need to change stuff in one place.

Elements as properties is very nice though, they should probably be protected and only accessed via methods within the page.

dtasev:
> A majority of the selenium tests are not using page object model pattern correctly. There are places in tests where there are calls to the webdriver that interact with the page.
> 
> e.g. calls to
> `.is_displayed()`
> `.text`
> `.find_element_by_x()`
> `.click()`
> 
> These types of calls should be in page methods. This means as pages change in the future we only need to change stuff in one place.

I'll see how much I can clean up, there are places where the number of elements changes dynamically (such as the submitting/editing/deleting tests in `test_configure_new_runs_integration`) so encapsulating that is very hard, as elements' position changes based on the state of the page.

For most of the `.is_displayed()` and `.click()` I don't really see a difference between 

```
self.page.submit_button.click()
assert not self.page.form_validation_message.is_displayed()
```

and

```
self.page.click_submit_button()
assert not self.page.form_validation_message_is_displayed()
```

which then does the action internally. My reasoning is that if element changes completely from being a button to something else, then the tests around them are likely to do quite different things too, and I don't think it being encapsulated helps.

That said I agree that `.find_element_by_x()` and `.text` are better encapsulated and I'll try to do that

WebApp/autoreduce_webapp/selenium_tests/tests/test_run_summary_integration.py
dtasev: I'll look into that! These potentially infinite while's already caused the tests to deadlock for 6 hours, so it would be good to have the wait until with a timeout

keiranjprice101:
To be more clear sorry. I do not think for a complex case like this we should have something so simple as `click_submit_button()` vs `submit_button.click()` I am suggesting it should be `submit_variables()` Then the mechanism of if its a button or something else is irrelevant, since the button aspect is hidden.

dtasev:
I see, I think this is the step in the right direction then: https://github.com/ISISScientificComputing/autoreduce/pull/1094/commits/c0a4f51a31647a55e8acf6fc29772154539d7bdb#diff-077f3e6a971defd01bf2123b1d3c4997ca0d99b1759a598c6a24cb9649acd6a0R30-R49 ! I'll see if there's more complex cases like this one to wrap into utils

dtasev:
If submitting doesn't work with this error:

```
  File "/home/dtasev/dev/autoreduce/queue_processors/queue_processor/instrument_variable_utils.py", line 100, in merge_arguments
    InstrumentVariablesUtils.set_with_correct_type(message_reduction_arguments, reduction_args, 'standard_vars')
  File "/home/dtasev/dev/autoreduce/queue_processors/queue_processor/instrument_variable_utils.py", line 121, in set_with_correct_type
    real_type = VariableUtils.get_type_string(reduction_args[dict_name][name])
KeyError: 'variable1'
```

it's because I've named the variable wrong in the reduce_vars - it should be `standard_vars` not `standard_variables`. You can fix that with

```
echo 'standard_vars = {"variable1":"value1"}' > data-archive/NDXTestInstrument/user/scripts/autoreduction/reduce_vars.py
```

Anders-Markvardsen:
Many thanks for fixes. Observe now after reinstall

webapp_bug

Anders-Markvardsen:
Separately I wonder if `.vscode` should be in `.gitignore` - to avoid accident unintentional local settting changes being committed?

dtasev:
@Anders-Markvardsen You'll have to add a `reduce.py` next to the `reduce_vars.py` and click the logo.  (you can find the `reduce.py` echo command in https://github.com/ISISScientificComputing/autoreduce/wiki/PR-testing-quickstart)

Strange, I know, but the index view checks for `reduce.py` and enabled/disables instruments based on whether `reduce.py` exists. I've made an issue to improve the error message for the user: https://github.com/ISISScientificComputing/autoreduce/issues/1176

keiranjprice101:
Can you improve the spacing on this button?
image

image

keiranjprice101:
image
These tooltips are not accessible. The information is not displayed when the links are selected by tabbing.
It works here:
image
so perhaps a similar approach would be better.

dtasev:
> image
> These tooltips are not accessible. The information is not displayed when the links are selected by tabbing.
> It works here:
> image
> so perhaps a similar approach would be better.

I agree, but won't be in this PR. The JS code around that is a mess, I'd rather touch it separately. Issue: https://github.com/ISISScientificComputing/autoreduce/issues/1178

keiranjprice101:
@dtasev is rerun description meant to be working. That is the only thing I am having issue with

dtasev:
> @dtasev is rerun description meant to be working. That is the only thing I am having issue with

Thank you for raising that, it should now (c4639d5). It was also now used when creating the run in the queue processor - I must have missed it because the `ReductionRun` field is `run_name` rather than `description`. That should be fixed too now.

keiranjprice101:
Yep seems to be working now. As a separate issue it would be nice if we added description to the summary page inside the summary box

keiranjprice101 (approved):
I have tried to be as thorough as possible with review and testing, but the size may mean I have missed something.

But it seems to be working nicely for me.

Thanks @dtasev this is a lot of nice changes

dtasev:
> I have tried to be as thorough as possible with review and testing, but the size may mean I have missed something.
> 
> But it seems to be working nicely for me.
> 
> Thanks @dtasev this is a lot of nice changes

Thank you for sparing the time to review it. I promise this is the last big PR and I'll be splitting stuff better in the future!

Anders-Markvardsen (approved):
Tested and thanks

As a reminder don't hesitate to copy any modified setup instructions to the wiki, and separately asked someone else to check over

dtasev:
I've copied the instructions to https://github.com/ISISScientificComputing/autoreduce/wiki/PR-testing-quickstart

Issue #1042, opened 2021-01-25
Refactor variable creation in Web App


dtasev:
Issue raised by: [developer/user/project requirement]

### What?
The whole variable creation code was, at some point, copy pasted into the web app (or the other way around, not sure). Regardless, the duplication needs to go and the two need to be merged into a common version with #1002

This should be done after #1022

dtasev:
Closed by #1042

PR #1102, opened 2021-02-09
Refactors url declarations from url to path. Moves most runs/ and instrument/ to apps and includes them instead

### LINKED ISSUE #1100: Refactor URLs to use django.urls.path and be in their own respective apps

### What?
What is the issue

### Where?

### How?
How did the issue come about

### Reproducible?
[Yes/No]

dtasev:
### Summary of work
- Renames `reduction_variables` to `instrument` to match the urls and what it actually does
- Moves URLs for runs/ and instrument/ into `reduction_viewer.urls` and `instrument.urls`

### How to test your work
Run the web app and check that it starts and all links work.

It's known that the `instrument/MARI/variables/` shows and error but that will be fixed in #1042

### Additional comments
 

Fixes #1100


rickybassom (approved):
Looks good 👍 . Tested on dev.

PR #1126, opened 2021-02-16
Replace data archive creator

### LINKED ISSUE #875: DataArchiveCreator prevents changes to settings.py

keiranjprice101:
### Summary of work
Remove the old `DataArchiveExplorer` and `DataArchiveCreator`
Combine their logic into a single class.
Move this class to the systemtests package as a util
Move away from unsafe `__del__` implementations.

### How to test your work
Tests should still pass 

Fixes #875


queue_processors/queue_processor/settings.py
```diff
@@ -37,7 +37,8 @@
 # ONLY USED IN THE WebApp and should be removed in https://github.com/ISISScientificComputing/autoreduce/issues/1042
 REDUCTION_DIRECTORY = os.path.join(ARCHIVE_ROOT, 'NDX%s', 'user', 'scripts', 'autoreduction')
 # %(instrument, cycle, experiment_number, run_number)
-ARCHIVE_DIRECTORY = os.path.join(ARCHIVE_ROOT, 'NDX%s', 'Instrument', 'data', 'cycle_%s', 'autoreduced', '%s', '%s')
+CYCLE_DIRECTORY = os.path.join(ARCHIVE_ROOT, 'NDX%s', 'Instrument', 'data', 'cycle_%s')
+ARCHIVE_DIRECTORY = os.path.join(CYCLE_DIRECTORY, 'autoreduced', '%s', '%s')
```
dtasev, line 41: I can't find this to actually be used anywhere. Same for `TEST_REDUCTION_DIRECTORY` and `TEST_ARCHIVE_DIRECTORY`. Perhaps we can just remove them?

systemtests/test_end_to_end.py
```diff
@@ -153,7 +142,7 @@ def test_end_to_end_wish_invalid_rb_number_skipped(self):
         # Create and send json message to ActiveMQ
         data_ready_message = Message(rb_number=self.rb_number,
                                      instrument=self.instrument,
-                                     data=file_location,
+                                     data=str(file_location),
```
dtasev, line 145: Is there a case where this is not a string? I briefly looked at `_setup_data_structures` and it seems like it always come out as a str

dtasev (changes requested):
Just a couple minor comments and it's ready to go

PR #1128, opened 2021-02-18
Adds pytest config for running django tests and use django.urls reverse instead of hard-coded URLs

dtasev:
### Summary of work
Adds pytest config for running django tests - this means you can now run them with `pytest WebApp/autoreduce_webapp/selenium_tests`. You will need to install the `pytest-django` plugin. Note that tests don't pass in parallel and that will need some more work.

Uses django.urls reverse instead of hard-coded URLs. This makes us more flexible as to the actual URL names and they will be dynamically generated from what is declared in django's urls.py

### How to test your work

CI tests should pass. 

Code review.

Try running the tests locally
- `pip install pytest-django`
- `pytest WebApp/autoreduce_webapp/selenium_tests`


keiranjprice101 (commented):
Looks good, but we will need to re add the removed classes when we add the tests for those pages. 

If we move from static `url` etc. it would be nicer to make them `@property` as well as abstract.

It is also slightly more confusing to call `reverse` within an assert, optionally it would be nice if we could add the path name for resolve as str constants to the corresponding page objects. Just to make it more clear which pages url we are looking for. Alternatively it could be more simple to return the reversed url from the original static method?

dtasev:
> Looks good, but we will need to re add the removed classes when we add the tests for those pages.

Yep I think that's fine. We'll add the pages back when we have tests for them. 

> If we move from static `url` etc. it would be nicer to make them `@property` as well as abstract.
> 
> It is also slightly more confusing to call `reverse` within an assert, optionally it would be nice if we could add the path name for resolve as str constants to the corresponding page objects. Just to make it more clear which pages url we are looking for. Alternatively it could be more simple to return the reversed url from the original static method?

The url method being static causes an inconvenience when you have a URL with parameters, because you can't reverse without passing them in, but the page object already knows the information, so it just gets duplicated on the call.

If we have a mixture of static and instance methods, I'm not sure whether that just makes it more confusing, as now some require.

As to the reverse in assert, I can see that, but I don't think it's that much of an issue for reverses without params, and the others we can capture in a url func

keiranjprice101:
> As to the reverse in assert, I can see that, but I don't think it's that much of an issue for reverses without params, and the others we can capture in a url func

I like this. 

I guess we can just disable pytest-django for the pytest action and all builds will pass. Looks good then 👍

dtasev:
Yeah it should be fine now, I made the commit yesterday but never pushed the branch 😄

keiranjprice101 (approved):

PR #1134, opened 2021-02-19
Moves driver start and quit in setUpClass/tearDownClass

dtasev:
### Summary of work
- Moves driver start and quit in setUpClass/tearDownClass
- Adds atexit register that will close the driver window

This means that only one Chrome/WebDriver instance will be started per test class. Should make tests run a bit faster. I wonder if this will cause any issues for parallel running though? Is it worth worrying about when it's not currently working anyway?

### How to test your work
CI tests should pass.

### Additional comments
 

Fixes #xyz

### DIFF
--- a/WebApp/autoreduce_webapp/selenium_tests/driver.py
+++ b/WebApp/autoreduce_webapp/selenium_tests/driver.py
@@ -8,6 +8,7 @@
 Module containing functions for obtaining webdrivers
 """
 import os
+import atexit
 
 from selenium import webdriver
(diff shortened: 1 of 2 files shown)


keiranjprice101:
This will, as you said make parallelisation harder in the future and is also not recommended. 
See https://www.selenium.dev/documentation/en/guidelines_and_recommendations/avoid_sharing_state/
Potential for headaches with browser caching etc. seems not worth the risk in my opinion

PR #1138, opened 2021-02-22
Fix JavaScript error

### LINKED ISSUE #1020: JavaScript exception on instrument not found page
Issue raised by: developer

### What?
JavaScript exception on instrument summary page when instrument is not found.
```js
instrument_summary.js:17
Uncaught TypeError: Cannot set property 'onchange' of null
    at init (instrument_summary.js:17)
    at instrument_summary.js:28
    at instrument_summary.js:29
```
Errors like this should be caught so that other JS code can execute on the page.

### Where?
On webapp go to `/instrument/randomstring/`
Template: `/WebApp/autoreduce_webapp/templates/instrument_summary.html`

### Reproducible?
Yes

* Load the webapp and go to `/instrument/randomstring/`, the exception should not appear in the console.

keiranjprice101:
### Summary of work

Check the element exists before setting onChange and onClick
Prevent automatic submission of filtering. (@dtasev I also found this confusing, and also jarring when the page would reload automatically)

### How to test your work
Verify no error on page `/runs/randomstr`
Verify no more automatic submission of filtering
Verify can still be filtered with apply

### Additional comments
 

Fixes #1020

### DIFF
--- a/WebApp/autoreduce_webapp/static/javascript/runs_list.js
+++ b/WebApp/autoreduce_webapp/static/javascript/runs_list.js
@@ -17,7 +17,7 @@
         if (select !== null) {
             select.onchange = function () {
                 setProgressCursor($(this));
-                document.getElementById('filter_options').submit();
+                document.getElementById('filter_options');
             };
         }
 
--- a/WebApp/autoreduce_webapp/static/javascript/runs_list.js
+++ b/WebApp/autoreduce_webapp/static/javascript/runs_list.js
@@ -8,22 +8,26 @@
         $('body').css('cursor', '');
     }
 
-    var init = function init(){
-        document.addEventListener("DOMContentLoaded", function(){
+    let init = function init() {
+        document.addEventListener("DOMContentLoaded", function () {
             setDefaultCursor($(this));
         });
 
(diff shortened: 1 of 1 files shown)


WebApp/autoreduce_webapp/static/javascript/runs_list.js
```diff
@@ -8,22 +8,26 @@
...
-        apply.onclick = function(){
-            setProgressCursor($(this));
+        let apply = document.getElementById('apply_filters');
+        if (apply !== null) {
+            apply.onclick = function () {
+                setProgressCursor($(this));
+            };
         }
-    };
+    }
```
rickybassom, line 30: I think the semicolon needs to stay because the start of this statement is `let int = ...`

rickybassom (changes requested):
Works well 👍

WebApp/autoreduce_webapp/static/javascript/runs_list.js
keiranjprice101: My mistake

rickybassom (approved):

PR #1139, opened 2021-02-23
Docker dependency check

### LINKED ISSUE #1111: Make a simple Docker CentOS7 image to make Dependabot PR testing easier

### What?
Currently the dependabot package version updates need to be tested locally and specially on CentOS7, because as we've found from IPython, some of the packages are NOT the same versions across all OS'es.

This testing can be made easier if we just install Autoreduction in a Docker image with CentOS7 base - which would test that deployment environment.

An overly complicated for AR docker image can be seen at https://github.com/mantidproject/mantidimaging/blob/master/docker/CentOS7_base.Dockerfile, we only really need the FROM centos7:base and then maybe preconfigure the venv for AR

keiranjprice101:
### Summary of work
Adds a Centos docker image for building to check dependencies are availble

### How to test your work
See action below

### Additional comments
 

Fixes #1111 


dependency-check/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+USER root
+
+COPY . /autoreduce
+
+WORKDIR /autoreduce
+
+RUN pip install --upgrade pip
+
+RUN  pip install -r requirements.txt
```
dtasev, line 11:
```suggestion
RUN pip install -r requirements.txt
```
Excess space

dtasev (changes requested):

PR #1148, opened 2021-02-24
Remove retry cancellation

### LINKED ISSUE #1084: Remove UI options to cancel retrying of runs because we never retry runs

### What?
There's a bunch of mentions & code regarding the option to cancel retrying of runs, but as far as I have seen so far we never actually schedule runs for retrying the reduction.

Additionally, looking at the implementation I am not sure if it ever worked properly as it didn't have integration with the queue processor where this would be handled.

If we ever need it again it would be easier to reimplement it within the new queue processor.

### Where?
There is a cancel option in `runs/failed/`

keiranjprice101:
### Summary of work
Removes retry cancellation from:
- Templates
- JavaScript
- view
- utils
- Message model
- ReductionRun model

### How to test your work
Should not be possible to cancel a retry from the fail page and all tests should remain passing

### Additional comments
 

Fixes #1084 


dtasev (approved):
Looks good!

PR #1171, opened 2021-03-03
Enable pylint on webapp

### LINKED ISSUE #1165: Turn on pylint for Web App

### What?
`pylint` is not being ran on the Web App.

keiranjprice101:
### Summary of work

A few things are done here:
 - Use pylint django in the webapp
 - Disable import related checks when running in the WebApp (Until the WebApp is separate these will remain problematic)
 - Removes a lot of disables that didnt need to be there
 - Added some to the WebApp that would require too large changes for this PR.
 - Fixes various warnings in the WebApp

To make reviewing this easier I recommend first reviewing these commits:
-  44489a02f3f
- 139f047
- 1788c0f
- fa27a9d16cdd9

These are where the bigger changes are. The rest is best to check by Files changed and skim as a lot of the changes are very very simple, just there is a lot of them

### Additional comments
 

Fixes #1165


keiranjprice101:
Master merge broke tests. Will investigate

WebApp/autoreduce_webapp/autoreduce_webapp/backends.py
```diff
@@ -8,16 +8,17 @@
 UOWS authentication backend
 """
 import logging
+
+# pylint:disable=imported-auth-user
```
dtasev, line 12: https://github.com/PyCQA/pylint-django/issues/244

I read through this a couple of times but I am still not 100% clear what the reason for this warning is

> Django recommends that you do not use the User model directly. Neither to issue queries against it, nor to use it as a foreign key reference. Instead you should be using the get_user_model() function and the AUTH_USER_MODEL setting.
The reason for all of this is that this can be overridden and applications which do not hard-code the values are in theory more portable downstream. That is they can be used directly inside an environment where the User model has been overridden.

Maybe an alternative is to just replace the `User` usage with `get_user_model()`

WebApp/autoreduce_webapp/autoreduce_webapp/management/commands/add_super.py
```diff
@@ -7,8 +7,9 @@
 """
 Custom manage.py command to add a super user to the database
 """
-from django.core.management.base import BaseCommand
+# pylint:disable=imported-auth-user
```
dtasev, line 10: Same as other comment for this imported-auth-user

WebApp/autoreduce_webapp/autoreduce_webapp/models.py
```diff
@@ -16,9 +16,9 @@ class Cache(models.Model):
     """
     created = models.DateTimeField(auto_now_add=True, blank=False)
 
-    def __unicode__(self):
-        # pylint: disable=no-member
-        return u'%s' % self.id_name
+    # pylint:disable=no-member
```
dtasev, line 19: Do we need this when it's __str__ and not __unicode__?

WebApp/autoreduce_webapp/autoreduce_webapp/uows_client.py
```diff
@@ -13,20 +13,17 @@
...
-class UOWSClient(object):
+class UOWSClient:
     """
     A client for interacting with the User Office Web Service
     """
     def __init__(self, **kwargs):
-        url = UOWS_URL
-        if 'URL' in kwargs:
-            url = kwargs['URL']
+        url = kwargs.get("URL", default=UOWS_URL)
```
dtasev, line 26: 👍

WebApp/autoreduce_webapp/autoreduce_webapp/views.py
```diff
@@ -9,15 +9,19 @@
...
 import os
 import sys
+
 from django.shortcuts import render
 from django.template import RequestContext
-# pylint: disable=no-member
 
 from utils.project.structure import get_project_root
+
+# pylint: disable=no-member
```
dtasev, line 18: I don't think we need that sys.path.append at all. I've removed a few before and it seemed to make no difference.

Does the import after it work without it? Note that this might also depend on where you start the webapp from:

- `python WebApp/autoreduce_webapp/manage.py runserver`, with cwd in the root of the repository
- `python manage.py runserver`, with cwd in `WebApp/autoreduce_webapp`

WebApp/autoreduce_webapp/instrument/views/runs.py
```diff
@@ -70,7 +69,7 @@ def submit_runs(request, instrument=None):
         return context_dictionary
 
 
-# pylint:disable=too-many-return-statements,too-many-branches,too-many-statements
+# pylint:disable=too-many-return-statements,too-many-branches,too-many-statements,too-many-locals
```
dtasev, line 72: I've got a feeling pylint is trying to tell us something 🤔

WebApp/autoreduce_webapp/selenium_tests/pages/component_mixins/rerun_form_mixin.py
```diff
@@ -9,6 +9,7 @@
 
 
 class RerunFormMixin:
+    # pylint:disable=no-self-use
```
dtasev, line 12: Rather we just made this `@staticmethod`

WebApp/autoreduce_webapp/selenium_tests/pages/configure_new_runs_page.py
```diff
@@ -19,6 +19,8 @@ class ConfigureNewRunsPage(Page, RerunFormMixin, NavbarMixin, FooterMixin, TourM
     """
     Page model class for run summary page
     """
+
+    # pylint:disable=too-many-arguments
```
dtasev, line 23: 5 is too many?! Guess it counts self, but still 6?!

WebApp/autoreduce_webapp/selenium_tests/pages/overview_page.py
```diff
@@ -40,6 +40,7 @@ def launch(self):
         WebDriverWait(self.driver, 10).until(presence_of_element_located((By.CLASS_NAME, "instrument-btn")))
         return self
 
+    # pylint:disable=no-self-use
```
dtasev, line 43: Can this be `@staticmethod`? I think it might have been an oversight from #1042

WebApp/autoreduce_webapp/selenium_tests/tests/test_configure_new_runs_integration.py
```diff
@@ -78,6 +75,25 @@ def _submit_var_value(self, value, start=None, end=None, experiment_number=None)
         self.page.variable1_field = value
         self.page.submit_button.click()
 
+    @staticmethod
+    def assert_expected_var(var: InstrumentVariable, expected_run_number, expected_reference, expected_value):
```
dtasev, line 79: 👍

scripts/manual_operations/tests/test_manual_remove.py
```diff
@@ -43,7 +43,7 @@ def make_test_run(experiment, instrument, run_version: str):
     return run
 
 
-# pylint:disable=invalid-name,too-many-public-methods
+# pylint:disable=invalid-name
```
dtasev, line 46: Is this name still invalid?

dtasev (changes requested):

WebApp/autoreduce_webapp/autoreduce_webapp/backends.py
keiranjprice101: I added this ignore as there is an issue for this .

WebApp/autoreduce_webapp/selenium_tests/pages/configure_new_runs_page.py
keiranjprice101: It's funny how everyones opinion differs I am in the 2 is pushing it camp 😄

WebApp/autoreduce_webapp/autoreduce_webapp/models.py
keiranjprice101: Yes we do

.pylintrc
```diff
@@ -61,7 +61,7 @@ disable=
       R1705, # no-else-return
       C0114, # missing-module-docstring
       C0115, # missing-class-docstring
-#      C0116, # missing-function-docstring
+      C0116, # missing-function-docstring
```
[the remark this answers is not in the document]
keiranjprice101: I wasn't sure why it was in the pylintrc at all and commented out. I assumed it was mistakenly commented so readded it. But I would prefer it removed also

WebApp/autoreduce_webapp/autoreduce_webapp/views.py
keiranjprice101: Yeah there are quite a few of these. I was going to open a separate issue to look at this, but don't mind doing it here if the size is alright.

WebApp/autoreduce_webapp/instrument/views/runs.py
keiranjprice101: Agree, but I felt refactoring this large function was out of scope I will open an issue for it.

scripts/manual_operations/tests/test_manual_remove.py
keiranjprice101: Good catch

keiranjprice101:
@dtasev in  d083e6e21b we both missed the problem on line 279 of the first file. Was the intention to assign `experiment_reference` to `int(request.POST.get("experiment_reference_number", 1))` before the if check?

WebApp/autoreduce_webapp/instrument/views/runs.py
```diff
@@ -272,7 +272,7 @@ def configure_new_runs_post(request, instrument_name, start=0):
             # Makes the variables that will be active for the range END + 1 -> onwards
             InstrumentVariablesUtils.find_or_make_variables(possible_variables, instrument.id, post_range_args, end + 1)
 
-    elif experiment_reference != 0:
```
dtasev, line 279: I think we should leave this in, and have an else that throws a RuntimeError, because if this gets POSTed without any info about start/end or experiment reference run then something has gone wrong!

dtasev (changes requested):

WebApp/autoreduce_webapp/instrument/views/runs.py
keiranjprice101: We cant leave this in. It is referenced before assignment

keiranjprice101: I commented above about this

keiranjprice101: I commented this above, if you didn't see

dtasev:
> @dtasev in [d083e6e](https://github.com/ISISScientificComputing/autoreduce/commit/d083e6e21b5c0e355d7d851a06e0aa209f5ed298) we both missed the problem on line 279 of the first file. Was the intention to assign `experiment_reference` to `int(request.POST.get("experiment_reference_number", 1))` before the if check?

That's a good catch, it should currently work but there's a mismatch between the:
- `experiment_reference` in the URL
- `experiment_reference_number` in the POST data

The mismatch will happen IF the user changes the experiment number manually. Functionally it still works because we update the value with that questionable `int`

.pylintrc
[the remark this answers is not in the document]
dtasev: 👍 I agree, I think it's good to have the function docstrings enforced. I remember ignoring it, but then changed my mind. Seems like I never removed it from the `.pylintrc`!

WebApp/autoreduce_webapp/instrument/views/runs.py
dtasev: Yep definitely not in the scope of this PR

WebApp/autoreduce_webapp/selenium_tests/pages/configure_new_runs_page.py
```diff
@@ -48,42 +50,54 @@ def url_path(self) -> str:
...
     @property
     def run_start(self) -> WebElement:
+        """
+        Return the run start input WebElement
+        """
         return self.driver.find_element_by_id("run_start")
 
     @property
     def run_start_val(self) -> WebElement:
+        """Return the value  run start WebElement"""
```
dtasev, line 60:
```suggestion
        """Return the value of the run start WebElement"""
```

dtasev (changes requested):
Small typo and it's ready to go!

PR #1188, opened 2021-03-05
Fixes configuring new variables when there are only skipped runs

### LINKED ISSUE #857: Rerun past jobs  & configure new runs will produce an error if only skipped runs exist

dtasev:
### Summary of work

Configuring new variables should work fine as long as there is a `reduce_vars.py`, as at no point are the previous run's variables used.

Changes the "See instrument variables" view to show "No current variables" if there are none.

Also adds test.

### How to test your work
Make sure to use these fixtures instead of the ones in the Quickstart page:
```
loaddata super_user_fixture status_fixture skipped_run
```

Follow the rest of the instructions from here to set up the rest: https://github.com/ISISScientificComputing/autoreduce/wiki/PR-testing-quickstart 

Fixes #857


keiranjprice101:
This only fixes the issue for configuring new jobs. You can still click rerun and get a server error.

Specifically:
```
File "/autoreduce/WebApp/autoreduce_webapp/instrument/views/runs.py", line 47, in submit_runs
    kwargs = ReductionRunUtils.make_kwargs_from_runvariables(last_run)
  File "/autoreduce/WebApp/autoreduce_webapp/reduction_viewer/utils.py", line 54, in make_kwargs_from_runvariables
    if reduction_run.run_variables.count() == 0:
AttributeError: 'NoneType' object has no attribute 'run_variables'
```

dtasev:
That should fix it. I'll add a test for the rerun page

WebApp/autoreduce_webapp/instrument/views/helper.py
```diff
@@ -0,0 +1,21 @@
...
+# Autoreduction Repository : https://github.com/ISISScientificComputing/autoreduce
+#
+# Copyright &copy; 2021 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+# ############################################################################### #
+
+from queue_processors.queue_processor.status_utils import STATUS
+
+
+def get_last(runs_for_instrument):
```
keiranjprice101, line 11: Since django is active record pattern, wouldn't this make more sense as either a method on  the `ReductionRun` model that takes an instrument as arg or a method on the `Instrument` model. Plus we dont end up with even more utils and helpers

dtasev: Yes that makes more sense. I thought about it but decided not to as it was on the FK relationship, and forgot I can just put it in the Instrument model!

DanNixon (approved):
:+1: Looks good to me, rerun and configure new runs worked as expected with no errors.

PR #1190, opened 2021-03-08
Improve help page

### LINKED ISSUE #699: Improve Webapp help page search
Issue raised by: [developer]

### What?
The webapp help page search bar doesn't work very well. It states that it is found by key word search but some obvious keywords that are even  in the text do not discover content some examples of this are: `ISIS`, `Accessing`, `Compute`, `vars`

We should investigate how this is normally done - I would imagine there are very common framework/libraries for this - and implement one.

### Where?
Webapp help pages

### How?
Manual testing

### Reproducible?
Yes

* Use more generic keyword to search and ensure sane results
* Search for words that exist in the text and ensure they are found

### LINKED ISSUE #987: Add missing page tests

rickybassom:
### Summary of work
- [x] Add fuzzy search
- [x] Styling improvements (making sure divs line up etc)
- [x] `<code>` syntax highlighting
- [x] Selenium tests
- [x] Remove some redundant launch methods in other page tests 

### How to test your work
* Code review
* Tests pass
* Start up webapp and check that the search, categories field and sidebar work as expected. Also test with different screen sizes.
* Check that styling doesn't look off

Fixes #699 
Fixes #987

### DIFF
--- a/WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
+++ b/WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
@@ -79,21 +79,21 @@ class HelpPage(Page, NavbarMixin, FooterMixin):
         Should be run post JS topic link generation.
         :return (List) A list of topic headers
         """
-        return [x.find_element_by_xpath("/h3/a").text.strip() for x in self.get_help_topic_header_elements()]
+        return [x.find_element_by_xpath("./h3/a").text.strip() for x in self.get_help_topic_header_elements()]
 
     def _get_category_filter_elements(self) -> List[WebElement]:
         """
         Get the filter elements in #category-filter
         :return: (List) A list of filter elements
         """
-        return self.driver.find_elements_by_xpath("//div[id='category-filter']/label")
+        return self.driver.find_elements_by_xpath("//div[@id='category-filter']/label")
 
     def get_topic_filters(self) -> List[str]:
         """
         Get the topic filters from the category filter
         :return: (List) A list of categories
         """
-        return [x.text.strip().lower() for x in self._get_category_filter_elements()]
+        return [x.get_attribute("data-category").strip().lower() for x in self._get_category_filter_elements()]
 
     def get_valid_topics(self) -> List[str]:
         filters = self.get_topic_filters()
@@ -143,10 +143,3 @@ class HelpPage(Page, NavbarMixin, FooterMixin):
         :return:
         """
         return [x for x in self.get_all_help_topic_elements() if x.is_displayed()]
-
-    def get_mock_help_topic_element(self) -> WebElement:
-        """
-
-        :return:
-        """
-        return self.driver.find_element_by_id("mock-help-topic")
--- a/WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
+++ b/WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
@@ -8,17 +8,18 @@
 Module for the help summary page model
 """
 from __future__ import annotations
+from typing import List
 
 from django.urls.base import reverse
+from selenium.webdriver.remote.webelement import WebElement
+
 from selenium_tests import configuration
 from selenium_tests.pages.component_mixins.footer_mixin import FooterMixin
 from selenium_tests.pages.component_mixins.navbar_mixin import NavbarMixin
-from selenium_tests.pages.component_mixins.rerun_form_mixin import \
-    RerunFormMixin
 from selenium_tests.pages.page import Page
 
 
-class HelpPage(Page, RerunFormMixin, NavbarMixin, FooterMixin):
+class HelpPage(Page, NavbarMixin, FooterMixin):
     """
     Page model class for help page
     """
@@ -43,3 +44,17 @@ class HelpPage(Page, RerunFormMixin, NavbarMixin, FooterMixin):
         self.driver.get(configuration.get_url())
         self.driver.get(self.url())
         return self
+
+    def get_sidenav_contents(self) -> List[WebElement]:
+        """
+        Get the contents of the collapsible sidebar
+        :return: (List) A list of <li> WebElements in #sidebar-contents
+        """
+        return self.driver.find_element_by_id("sidenav-contents").find_elements_by_tag_name("li")
+
+    def get_help_topics(self) -> List[WebElement]:
+        """
+        Get the help topics on the page
+        :return: (List) A list of <section> WebElements in .help-topic
+        """
+        return self.driver.find_elements_by_class_name("help-topic")
--- a/WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
+++ b/WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
@@ -49,11 +49,14 @@ class HelpPage(Page, NavbarMixin, FooterMixin):
         Get the contents of the collapsible sidebar
         :return: (List) A list of <li> WebElements in #sidebar-contents
         """
-        return self.driver.find_element_by_id("sidenav-contents").find_elements_by_tag_name("li")
+        print("sidenav-contents test")
+        print(list(self.driver.find_elements_by_xpath("//ul[@id='sidenav-contents']/li")))
+        return self.driver.find_elements_by_xpath("//ul[@id='sidenav-contents']/li")
 
     def get_help_topics(self) -> List[WebElement]:
         """
         Get the help topics on the page
         :return: (List) A list of <section> WebElements in .help-topic
         """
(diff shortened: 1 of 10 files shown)


WebApp/autoreduce_webapp/selenium_tests/tests/test_help_page.py
```diff
@@ -0,0 +1,23 @@
...
+        """
+        super().setUp()
+        self.page = HelpPage(self.driver)
+
+    def test_all_sidebar_contents_generated(self):
+        """
+        Test that the number of elements in the sidebar equals the number of
+        topics on the page
+        """
+        WebDriverWait(self.driver, 10).until(len(self.page.get_sidenav_contents()) == len(self.page.get_help_topics()))
```
dtasev, line 23:
```suggestion
        WebDriverWait(self.driver, 10).until(lambda _: len(self.page.get_sidenav_contents()) == len(self.page.get_help_topics()))
```
The `until` condition needs to be a callable

WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
```diff
@@ -42,6 +43,7 @@ def launch(self) -> HelpPage:
         # Navigates to / first to force a login. Check the README and
         # the "index" view for more details
         self.driver.get(configuration.get_url())
+        self.driver.get("http://localhost:0000/overview")
```
dtasev: This shouldn't hard code the URL but I don't think overview does anything special, it's `/` that we want to force login

WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
```diff
@@ -0,0 +1,62 @@
...
+        print(list(self.driver.find_elements_by_xpath("//ul[@id='sidenav-contents']/li")))
+        return self.driver.find_elements_by_xpath("//ul[@id='sidenav-contents']/li")
+
+    def get_help_topics(self) -> List[WebElement]:
+        """
+        Get the help topics on the page
+        :return: (List) A list of <section> WebElements in .help-topic
+        """
+        print(self.driver.find_elements_by_class_name("help-topic"))
+        return self.driver.find_elements_by_class_name("help-topic")
```
dtasev, line 62: Leftover prints?

WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
```diff
@@ -0,0 +1,62 @@
...
+    def launch(self) -> HelpPage:
+        """
+        Open the page with the webdriver
+        :return: (HelpPage) The HelpPage object model
+        """
+        # Navigates to / first to force a login. Check the README and
+        # the "index" view for more details
+        self.driver.get(configuration.get_url())
+        self.driver.get(self.url())
+        return self
```
dtasev, line 45: This seems identical to the implementation already in [Page](https://github1s.com/ISISScientificComputing/autoreduce/blob/699-improve-help-page-search/WebApp/autoreduce_webapp/selenium_tests/pages/page.py#L19)

dtasev (changes requested):
Works well, just a few comments

WebApp/autoreduce_webapp/selenium_tests/tests/test_help_page.py
```diff
@@ -0,0 +1,24 @@
+from selenium_tests.pages.help_page import HelpPage
+from selenium_tests.tests.base_tests import NavbarTestMixin, BaseTestCase, \
+    FooterTestMixin
+from selenium.webdriver.support.wait import WebDriverWait
+
+
+class TestHelpPage(NavbarTestMixin, BaseTestCase, FooterTestMixin):
```
dtasev, line 7: It would be nice to have some extra tests, perhaps ones that enter some text in the search and assert that the right boxes remain shown?

This helper for writing to a field should make it easier to do that:https://github.com/ISISScientificComputing/autoreduce/blob/699-improve-help-page-search/WebApp/autoreduce_webapp/selenium_tests/pages/component_mixins/rerun_form_mixin.py#L12-L15

WebApp/autoreduce_webapp/selenium_tests/tests/test_help_page.py
```diff
@@ -20,4 +20,5 @@ def test_all_sidebar_contents_generated(self):
         Test that the number of elements in the sidebar equals the number of
         topics on the page
         """
-        WebDriverWait(self.driver, 10).until(len(self.page.get_sidenav_contents()) == len(self.page.get_help_topics()))
+        WebDriverWait(self.driver,
```
keiranjprice101, line 23: This will still pass even if no help topics or contents exist on the page

WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
rickybassom: Selenium doesn't want to work on WSL so pushing up to Github is the only way I can debug! I need will remove them once finished with this PR.

WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
rickybassom: I didn't notice that, thanks for spotting. In that case the launch methods here should be removed also: 
- `rerun_jobs_page.py`
- `run_summary_page.py`
- `runs_list_page.py`

WebApp/autoreduce_webapp/selenium_tests/tests/test_help_page.py
rickybassom: I'm planning on adding these tests:
```py
# def test_all_topics_have_headers
# def test_all_topics_have_a_category
# def test_all_topics_have_content
# def test_filter_by_categories
# def test_scroll_to_header
```

WebApp/autoreduce_webapp/selenium_tests/tests/test_help_page.py
rickybassom: I'm on the fence about adding `== constant` because every time someone want to add another topic they will have to change the test. You shouldn't have to change the tests for something that is allowed.

I've added a test that checks that `topics>0` because I agree a Help page should logically include help info. But I'm interested to hear which way you think is _good practice_ for unit tests.

keiranjprice101: asserting > 0 does not mean they are all loaded. If the test changes dynamically it is not deterministic and is only testing that at least 1 condition hasn't gone wrong and not that everything is working as should.

E.g. if lets say 12 topics should load, but because something has been broken only the first 6 load, the test passes and we wont know there is an issue unless we notice or it is reported to us.

 Increasing the number in one line is also a very small overhead

WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
dtasev: Yep I think we can remove them. If not the tests will fail 😄

WebApp/autoreduce_webapp/selenium_tests/pages/help_page.py
dtasev: Not even headless? That doesn't sound good! You'll have a very hard time writing the tests without being able to see what the page actually looks like (for which you also need to run headless=false), from my experience writing them at least

WebApp/autoreduce_webapp/selenium_tests/tests/test_help_page.py
dtasev: As the help is currently hard coded in the HTML file I think we can be confident that they are all rendered. 

It's useful to have both `topics > 0` and `len(self.page.get_sidenav_contents()) == len(self.page.get_help_topics())` checks IMO (maybe in the same test), as that will cause a failure if nothing is rendered (e.g. an error in the view that causes a 500 will render nothing).

I don't think we need to check for the exact number of topics shown until we are adding them dynamically (e.g. via a django templated for loop)

keiranjprice101: The sidebar is generated dynamically though I thought?

dtasev:
Accidentally committed a fix for #1199 on this branch. I've reverted the commit to this branch.

dtasev (approved):

PR #1199, opened 2021-03-11
Makes the message.description default = ""

### LINKED ISSUE #1198: Message.description cannot be default=None anymore

### What?
The default for `Message.description` is `description = attr.ib(default=None)`.

This doesn't work well anymore because we *always* pass in `message.description` on `db_records.create_reduction_run_record` - see [here](https://github1s.com/ISISScientificComputing/autoreduce/blob/master/model/database/records.py)

This is because the ReductionRun field is `run_description = models.CharField(max_length=200, blank=True)`, but not `null=True`!

### Solution
Make the message.description default to "" and fix the tests.

### Testing
Submitting runs without description should no longer show this error in logs:
```
django.db.utils.IntegrityError: NOT NULL constraint failed: reduction_viewer_reductionrun.run_description
```

- Start Queue Listener
- Start Web App (to see result)
- Submit `python scripts/manual_operations/manual_submission.py MARI 27878`
- Check that error is like "No such file or directory ... reduce_vars.py"

dtasev:
### Summary of work
Makes the message.description default = ""

This does not break the Run Summary description visualisation in the web app as empty string is does not pass the `{% if run_description %}` check and that field is not rendered, just like before.

### How to test your work
Check https://github.com/ISISScientificComputing/autoreduce/issues/1198

Fixes #1198

### DIFF
--- a/model/message/message.py
+++ b/model/message/message.py
@@ -21,7 +21,7 @@ class Message:
     A class that represents an AMQ Message.
     Messages can be serialized and deserialized for sending messages to and from AMQ
     """
-    description = attr.ib(default=None)
+    description = attr.ib(default="")
     facility = attr.ib(default="ISIS")
     run_number = attr.ib(default=None)
     instrument = attr.ib(default=None)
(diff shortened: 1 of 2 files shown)


keiranjprice101 (approved):

keiranjprice101:
I guess that is a minor test change as its only the init test failing

PR #1210, opened 2021-03-15
Makes the info boxes more accessible

### LINKED ISSUE #1207: Make actions in configure new jobs and rerun jobs accessible

### What?
What is the issue

### Where?

### How?
How did the issue come about

### Reproducible?
[Yes/No]

dtasev:
### How to test your work

These images capture the changes well.

Rerun jobs:
Before:
image

After 
image

Configure new runs:
Before:
image

After:
image

### Additional comments
 

Fixes #1207

### DIFF
--- a/WebApp/autoreduce_webapp/templates/configure_new_runs.html
+++ b/WebApp/autoreduce_webapp/templates/configure_new_runs.html
@@ -111,7 +111,7 @@
                     <div class="row">
                         <ul class="js-form-actions">
                             <li>
-                                <a href="#currentScript" id="currentScript">Reset to values in the current reduce_vars script</a>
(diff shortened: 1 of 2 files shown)


dtasev:
@Anders-Markvardsen this is the change I described in our call, can you have a quick look? It's just a visualisation change and the pictures capture it well

keiranjprice101:
I think there is also an accessibility issue with `Reset to values in current script` being a link and not a button. And that it should either be changed to a button or have the `role="button"` attribute since it doesn't navigate

dtasev:
I will add `role="button"` because changing it to a real button messes up the styling

keiranjprice101 (approved):

PR #1232, opened 2021-03-22
Removes DB call from reduction runner

### LINKED ISSUE #1231: Selenium tests local fail when DB is accessed from Reduction runner

### What?
It seems accessing the DB during testing causes the Selenium tests to operate on 2 different DBs - one seems to be from Pytest, the other one the default from the settings. The issue seems to be that when the reduction runner starts it is a new process, so it doesn't inherit any environment settings that pytest might have changed.

I wasn't able to figure out how to fix that so the faster solution seemed to remove the DB access from the reduction runner, and move the required call inside handle_message instead.

dtasev:
### Summary of work
_copied from issue text_

It seems accessing the DB during testing causes the Selenium tests to operate on 2 different DBs - one seems to be from Pytest, the other one the default from the settings. The cause seems to be that when the reduction runner starts it is a new process, so it doesn't inherit any environment settings that pytest might have changed.

I wasn't able to figure out how to fix that so the faster solution seemed to be to remove the DB access from the reduction runner, and move the required call inside handle_message instead, then pass that information through the Message object

### How to test your work
CI tests should pass

### Additional comments
 

Fixes #1231

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -404,7 +404,7 @@ valid-metaclass-classmethod-first-arg=mcs
 [DESIGN]
 
 # Maximum number of arguments for function / method
-max-args=5
+max-args=8
 
 # Maximum number of attributes for a class (see R0902).
 max-attributes=7
--- a/WebApp/autoreduce_webapp/reduction_viewer/utils.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/utils.py
@@ -57,7 +57,6 @@ class ReductionRunUtils:
 
         return {"standard_vars": standard_vars, "advanced_vars": advanced_vars}
 
-    # pylint: disable=too-many-arguments
     @staticmethod
     def send_retry_message(user_id: int, most_recent_run: ReductionRun, run_description: str, script_text: str,
                            new_script_arguments: dict, overwrite_previous_data: bool) -> None:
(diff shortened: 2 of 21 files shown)


keiranjprice101:
I may be missing something, but I don't see where flat output would be set in the initial message

keiranjprice101:
I am also slightly confused about the issue, where has this been introduced if the build passes on master

dtasev:
> I may be missing something, but I don't see where flat output would be set in the initial message

In the initial message it defaults to False, but then gets overwritten with the instrument's value right before reduction here: https://github.com/ISISScientificComputing/autoreduce/pull/1232/files#diff-d835da45e97324b3c741476ec64bc50b5603a9a2eca8eb11b81b56d1dcf9b228R139

> I am also slightly confused about the issue, where has this been introduced if the build passes on master

I don't see the issue before https://github.com/ISISScientificComputing/autoreduce/pull/1213 and I'm not sure why it's not failing on the buildservers. 

I saw a strange behaviour where the subprocess would raise an exception (and trigger a breakpoint in VSCode) but if I just continue anyway the test passes. Looking at some of the assertions, it's likely they are checking parts of the reduction run object that are created regardless whether the reduction passes or fails (i.e. checking run version, variables etc)

It's possible this happens in the test server too but I'm not sure how to verify

keiranjprice101:
If there arent too many before merging could you grep for too-many-arg disables and remove them since we have upped the max args?

keiranjprice101 (approved):

PR #1235, opened 2021-03-23
Add fail tests AR-1235

keiranjprice101:
### Additional comments
 

Fixes #xyz


dtasev:
Is this ready for review?

keiranjprice101:
@dtasev  No

WebApp/autoreduce_webapp/selenium_tests/tests/test_failed_jobs_page.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        Test rerunning failed run creates rerun
+        """
+
+        archive, db_client, queue_client, listener = setup_external_services("TestInstrument", 21, 21)
+        try:
+            listener._processing = True  # pylint:disable=protected-access
+            self.page.toggle_run("99999").retry_runs()
+            WebDriverWait(self.driver, 30).until(lambda _: not listener.is_processing_message())
+            runs = access.get_reduction_run("TestInstrument", "99999")
```
dtasev, line 55: Could this use [submit_and_wait_for_result](https://github.com/ISISScientificComputing/autoreduce/blob/625245cc80fe55005a5b977cb7d0a0fed05142a5/WebApp/autoreduce_webapp/selenium_tests/utils.py)?

dtasev (changes requested):
Looks good overall, just a question

WebApp/autoreduce_webapp/selenium_tests/tests/test_failed_jobs_page.py
keiranjprice101: It can't as the url doesnt change and it redirects back to the original list and not a run confirmation.

dtasev: I think you could move the `expected_url` to be a function argument with `None` default, and `if expected_url is None` then it gets the reverse of `"run_confirmation"`. Would that work?

PR #1243, opened 2021-03-25
Further improve error capture in runner.py

dtasev:
### Summary of work
Stops usage of `message.message` as primary error carrier, because it causes this:

image

Now properly adds error in `message.reduction_log` which makes it look neater, and allows you to view the error in more detail after clicking:
image

### How to test your work
Tests should pass


keiranjprice101 (approved):
This is much more sane 👍

PR #1246, opened 2021-03-26
Resubscribe on disconnect AR-88

### LINKED ISSUE #88: Queue processors don't reconnect if ActiveMQ goes down
If ActiveMQ goes down both of the queue processors will not reconnect when it comes back up, in fact they don't even realise it's gone down and will sit idle until restarted.

Useful resources to fix this: 
https://jasonrbriggs.github.io/stomp.py/api.html#dealing-with-disconnects
https://stomp.github.io/stomp-specification-1.2.html#Heart-beating

### LINKED ISSUE #1245: Simplify QueueClient

### What?
Now that there is only 1 queue the queue client is needlesly complicated. Its api offers ways to connect to multiple queues etc. When we only have 1 queue.

### Where?
`QueueClient`

### How?
How did the issue come about

### Reproducible?
[Yes/No]

Refactor into a more simple api

### LINKED ISSUE #1274: Add a timeout to systemtests

### What?
It is possible to create a situation where the systemtests will never terminate if `is_processing` never becomes False.

We should add a timeout to prevent this.

### Where?
`test_end_to_end.py`

### Reproducible?
[Yes] It is simple to create a test where this would happen intentionally.

It should no longer be possible to get stuck.

keiranjprice101:
### Summary of work
- Add additional tests to `QueueClient` 
- Remove the `QueueListener` from `HandleMessage` (It was unused)
- Move `handle.connected` to within `HandleMessage` 
- [ ] Add similar to `setup_connection` back in to call from test/systemtest/daemon to remove repeated connect_and_subscribe
- [ ] Simplify subscribe in queue client
- [ ] Add test for amq connection lost
### Additional comments
 

Fixes #1245 
Fixes #88
Fixes #1274


dtasev:
Just saw this happen yesterday because the AMQ server was rebooted over the weekend

keiranjprice101:
Replaced by #1314

PR #1251, opened 2021-03-26
Minor fixes discovered during deploying smoke testing

dtasev:
### Summary of work
Adds a missing f to an f-string
Adds facility="ISIS" to the run_detection
Use the hostname as connection ID (though I'm not sure we can see that anywhere)
### Additional comments
 

Fixes #xyz

### DIFF
--- a/model/message/message.py
+++ b/model/message/message.py
@@ -83,7 +83,7 @@ class Message:
                     # Set the value of the variable on this object accessing it by name
                     setattr(self, key, value)
             else:
-                raise ValueError("Unexpected key encountered during Message population: '{key}'.")
+                raise ValueError(f"Unexpected key encountered during Message population: '{key}'.")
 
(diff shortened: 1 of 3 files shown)


model/message/message.py
```diff
@@ -83,7 +83,7 @@ def populate(self, source, overwrite=True):
                     # Set the value of the variable on this object accessing it by name
                     setattr(self, key, value)
             else:
-                raise ValueError("Unexpected key encountered during Message population: '{key}'.")
+                raise ValueError(f"Unexpected key encountered during Message population: '{key}'.")
```
keiranjprice101, line 86: very nice catch

dtasev: Only after spending 15 minutes sending messages and wondering why the key name is "key" 😄

keiranjprice101 (approved):

Issue #1271, opened 2021-03-30
Duplicated requirements


keiranjprice101:
Issue raised by: [developer/user/project requirement]

From: https://github.com/ISISScientificComputing/autoreduce/pull/1273#issuecomment-810556133

The following are duplicated:
- dash (With different versions too!)
- django_plotly_dash
- plotly
- requests
- stomp.py

### What?
THe same packages are in both in `requirements.txt` and `setup.py`

### Where?
`requirements.txt`

### How to test the issue is resolved
Remove from `requirements.txt` Everything should still pass

PR #1273, opened 2021-03-30
Extract requirements from setup.py into requirements.txt 

dtasev:
### Summary of work

Extracts requirements from setup.py into `requirements.txt`.

Move the dev requirements into `dev-requirements.txt`

This is done because for whatever reason `pip install -e autoreduce` will upgrade packages that autoreduce **DOES NOT** depend on.

`pip install -r` does not have this behaviour.

This was causing issues on production by thrashing the `h5py` needed by Mantid.

### How to test your work
- Code review the changes to the actions

The fastest way to see the result is to pull the docker container (adjust autoreduce code location).
```
docker run --network=host -v ~/dev/autoreduce/:/autoreduce -v /isis:/isis -it autoreduction/qp /bin/bash
```

To test the installation:
- Create a new virtual or conda env
- `pip install -r requirements.txt && pip install -r dev-requirements.txt`
- AR should run as normal


dtasev:
Turns out the `mantid-framework` from conda is missing bits we need (the plotting additions to matplotlib) so I had to revert to using a native Mantid installation.

This made the h5py issue come up again regardless of the changes in this PR. I fixed it with `python3 -m pip uninstall -y h5py && apt install --reinstall -y python3-h5py` in the container buildfile instead.

Sadly #1271 won't be fixed here but I've updated the issue to include your comment so that we don't miss them

PR #1289, opened 2021-04-08
Check if the `?next` URL is safe before using it as redirecturl from UOWS

### LINKED ISSUE #1288: Severe Web App vulnerability with ?next parameter
Issue raised by: developer

### What?
It's exploited by setting any redirect URL in `http://reducedev2.isis.cclrc.ac.uk:8080/?next=<url>`. So someone could send you an email saying look at this `http://reducedev2.isis.cclrc.ac.uk:8080/?next=fakereducewebapp.com` and it would send you to the fake site and steal your credentials etc.

The easy fix is to jump in on that `?next` parameter handling and verify that the value starts with one of `ALLOWED_HOSTS`. Perhaps use Django's `is_safe_url` and pass in `ALLOWED_HOSTS`

### Where?
Webapp ?next handling

### Tests
Check that going to `http://reducedev2.isis.cclrc.ac.uk:8080/?next=fakereducewebapp.com` will show some kind of error or just be ignored

dtasev:
### Summary of work
Check if the `?next` URL is safe before using it as `redirecturl` from UOWS.

The safe URLs are the one listed in `ALLOWED_HOSTS`, which are usually localhost & the host domain.

### Additional comments
 
Fixes #1288

### DIFF
--- a/WebApp/autoreduce_webapp/reduction_viewer/views.py
+++ b/WebApp/autoreduce_webapp/reduction_viewer/views.py
@@ -48,8 +48,8 @@ def index(request):
     Render the index page
     """
     return_url = UOWS_LOGIN_URL + request.build_absolute_uri()
-    if request.GET.get('next'):
-        next_url = request.GET.get('next')
+    next_url = request.GET.get('next')
+    if next_url:
         if url_has_allowed_host_and_scheme(next_url, ALLOWED_HOSTS, require_https=True):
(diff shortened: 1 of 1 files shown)


rickybassom (changes requested):
Works well 👍 . Code looking good apart from the code inspection error.

rickybassom (approved):

PR #1291, opened 2021-04-08
Remove SFTP Client from plot_handler, add whitenoise to serve staticfiles AR-1291

dtasev:
### Summary of work
Removes SFTP Client from plot_handler - a benefit of hosting the webapp on CentOS is that we can just copy the files from CEPH without a need for SFTPing. This could be bad if we move away from hosting on Linux/SCD Cloud but that's probably not worth worrying about.

Also adds Whitenoise as Django middleware to serve staticfiles (CSS/JS/plot images). This is a workaround to:
- Plot files being copied locally to render in the web app. 
- Apache not being able to access that location anymore, as the web app is on a different machine

The proper fix would be to move the staticfiles storage on a separate, common location. This would be necessary when we have >1 hosts for the web app.

### How to test your work
These changes are currently live on the production env. 

Check that the production website loads assets and plots OK. Do Shift + F5 to clear the cache.

### DIFF
--- a/.github/actions/build-autoreduction/action.yml
+++ b/.github/actions/build-autoreduction/action.yml
@@ -24,4 +24,5 @@ runs:
         cat build.log
         _LOCAL_PYTHON_PATH=$(which python)
         sudo $_LOCAL_PYTHON_PATH setup.py start
+        python WebApp/autoreduce_webapp/manage.py collectstatic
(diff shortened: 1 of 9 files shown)


dtasev:
Converting to draft because I need to fix a lot of tests with this change.

PR #1308, opened 2021-04-14
Keep real types in the database objects' values AR-1283 AR-1329

dtasev:
### Summary of work
Stops the queue processor from converting everything to string before saving in the DB - this means the original types are now kept. Adds tests for the different supported variable types to check that they are properly shown on the web app and stored in the DB as expected

**Adds back replacement of the reduction arguments** - currently only the values in reduce_vars are used and are never replaced. The fix will be deployed in production after this PR is merged. Adds test for this so it doesn't happen again.

### How to test your work
Tests should pass

### Additional comments
 

Fixes AR-1283


keiranjprice101:
Makes sense, tests still failing though.

keiranjprice101 (approved):

PR #1314, opened 2021-04-21
Reconnect on disconnect AR-88 AR-1245 AR-1246 AR-1274

keiranjprice101:
### Summary of work
The listener now implements the stomp listener base class as it should
The queue client now removes its listeners when it disconnects as it should
The queue client is refactored slightly as well as the listener
A timeout has been added to the wait part of the end to end test.
Adds a system test for activemq going down 

### How to test your work
Test should pass
Stop ActiveMQ, start ActiveMQ submit run, should still be picked up

### Additional comments
 

Fixes #xyz


queue_processors/queue_processor/queue_listener.py
```diff
@@ -63,10 +73,8 @@ def on_message(self, headers, message):
             self.logger.info("Destination: %s Priority: %s", destination, priority)
             # Load the JSON message and header into dictionaries
             try:
-                if not isinstance(message, Message):
```
keiranjprice101, line 66: It is not possible for the listener to recieve a message object so this if check was always `True`

utils/clients/tests/test_queue_client.py
```diff
@@ -148,3 +143,44 @@ def test_create_connection_bad_production(self):
 
         client.credentials.host = real_host
         del os.environ["AUTOREDUCTION_PRODUCTION"]
+
+    def test__test_connection_not_connected(self):
```
dtasev, line 147: Extra `test__`?

utils/clients/tests/test_queue_client.py
```diff
@@ -148,3 +143,44 @@ def test_create_connection_bad_production(self):
...
+
+        """
+        client = QueueClient()
+        mock_connection = MagicMock()
+        mock_connection.is_connected.return_value = False
+        client._connection = mock_connection
+        with self.assertRaises(ConnectionException):
+            client._test_connection()
+
+    def test__test_connection_connected(self):
```
dtasev, line 160: Extra `test__`?

dtasev (changes requested):
Question about test names, will test locally now

dtasev (changes requested):
Question about test names, will test locally now

utils/clients/tests/test_queue_client.py
keiranjprice101: Autogenerated from method name. Will remove as its clear anyway.

queue_processors/queue_processor/tests/test_queue_listener.py
```diff
@@ -111,11 +93,22 @@ def test_on_message_sends_acknowledgement(self):
...
+        """Test correct calls on disconnect"""
+        mock_con = MagicMock()
+        mock_sub = MagicMock()
+        self.mocked_client.connect = mock_con
+        self.mocked_client.subscribe = mock_sub
+
+        self.listener.on_disconnected()
+
+        mock_con.assert_called_once()
+        mock_sub.assert_called_once()
```
dtasev, line 114: Need to add 1 more test - when `self.mocked_client.connect.side_effect = ConnectionError`.

This might get stuck in an infinite loop though, you might need a more "advanced" side effect, or another way to break it:

```
def connect_side_effect(connect_mock):
	connect_mock.side_effect = None
	raise ConnectionError

self.mocked_client.connect.side_effect = partial(connect_side_effect, self.mocked_client.connect)
```

dtasev (changes requested):

dtasev:
Tested and works excellent. 

Just for reference a local ActiveMQ can be started with:

```
docker run -p 61613:61613 -p 8161:8161 -it rmohr/activemq:5.15.9
```

[rmohr/activemq](https://hub.docker.com/r/rmohr/activemq)

(61613 is the queue itself, 8161 is `api/jolokia`)

I don't think it's worth changing the systemtests at this point, but if it proves a nuisance we may move to docker (which comes with other drawbacks, such as needing sudo access)

keiranjprice101:
Good point that I overlooked, funny considering I have used it a lot of times lol. I will update the test now

dtasev:
Wondering if we may want this in the release branch instead

keiranjprice101:
@dtasev As long as we remember to merge it back into master I dont see why not.

PR #1319, opened 2021-04-26
Autoreduction v21.1

keiranjprice101:
### Summary of work
Merge back to master for QP change. Not to delete branch on merge.


dtasev:
Some Github extension I have automatically deleted it but I've restored it 😄

PR #1356, opened 2021-06-09
Read in file as non-binary UTF-8 | AR-1420

dtasev:
### Summary of work
Removes using `chardet` to try and find the encoding as the implementation was broken - it just checked one character `chardet.detect(file_raw.read(32))["encoding"]`. For this to work properly, that specific character had to be non-ASCII.

This may be leftover from Python 2.7

It seems that there is no need for this anymore. Just doing `open` and reading in the file works, and can handle any UTF-8 characters.

Adds some tests with emojis to ensure utf-8 compatibility.

### How to test your work
Tests should pass. Code review


DanNixon (approved):
:+1:

PR #1378, opened 2021-08-05
Reduction directory version now uses the run_version from the message

dtasev:
### Summary of work
Makes the reduction directory version use the `run_version` field from the message, rather than `the number of folders in the output location + 1`. This should make the `run_version` shown in the webapp and on the folder always be consistent, regardless of the presence of the run version folders in the output folder

### How to test your work
1. To be able to submit the runs, you'll need some prep:
    - Start ActiveMQ with the task `Run ActiveMQ Docker Daemon`
    - Start debugging the queue processor with the `Run queue listener` run configuration
2. Open a separate terminal and activate the Autoreduce environment, and check that you can run `autoreduce-manual-submission`, it should print a help command, if not you'll have to `pip install -e ./autoreduce-scripts`
3. To be able to run a reduction two files are needed
    - A datafile, that we can fake with `mkdir -p /isis/NDXINTER/Instrument/data/cycle_21_1/ && touch /isis/NDXINTER/Instrument/data/cycle_21_1/INTER00063125.nxs` - it's fine if it's empty, it just means it will fail the reduction, but it must exist to start
    - A reduce.py script, that we can make with `echo 'def main(input_file, output_dir): print("Hello")' > ~/.autoreduce/dev/data-archive/NDXINTER/user/scripts/autoreduction/reduce.py`
4. With that you can now run `autoreduce-manual-submission INTER 63125` (note the number here must match the ending of the datafile name, or it won't be found

**What to look for**

In the `Run queue processor` terminal output look for the line:

```
[05/Aug/2021 15:59:20] INFO [autoreduce_qp.queue_processor.reduction:219] Final Result directory: /home/username/.autoreduce/dev/reduced-data/INTER/RB1835571/autoreduced/63125/run-version-NNN
```

where `NNN` will increment every time you run `autoreduce-manual-submission INTER 63125`. Run a few and then run

```
rm -r ~/.autoreduce/dev/reduced-data/INTER/RB1835571/autoreduced/63125/run-version-N/
```
(replace N with whichever is the latest), then submit again, the final submission should have a run version 1 higher than the one that was deleted.

### Additional comments
 


Dagonite (commented):
LGTM!

PR #1406, opened 2021-11-01
1473 drop down menu ar 1473

Dagonite:
### Summary of work
Update get_script_and_arguments() to fetch the contents of any raw GitHub file URLs and store them within the arguments dictionary as the values of the appropriate corresponding key.

### DIFF
--- a/autoreduce_qp/model/database/records.py
+++ b/autoreduce_qp/model/database/records.py
@@ -1,10 +1,10 @@
-# ############################################################################
+# ############################################################################ #
 # Autoreduction Repository :
 # https://github.com/ISISScientificComputing/autoreduce
 #
 # Copyright &copy; 2021 ISIS Rutherford Appleton Laboratory UKRI
 # SPDX - License - Identifier: GPL-3.0-or-later
-# ############################################################################
+# ############################################################################ #
 """Contains various helper methods for managing or creating ORM records."""
 # pylint:disable=no-member
 import json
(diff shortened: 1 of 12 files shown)


autoreduce_qp/model/database/records.py
```diff
@@ -66,10 +68,30 @@ def get_script_and_arguments(instrument: Instrument, script: str, arguments: dic
...
+    GitHub files.
+    """
+    for k1, v1 in arguments.items():
+        for k2 in v1:
+            nested_value = arguments[k1][k2]
+            if isinstance(nested_value, dict):
+                if "url" in nested_value and "default" in nested_value:
+                    url = nested_value["url"] + nested_value["default"]
+
+                    req = requests.get(url)
```
dtasev, line 88: This should be in a try: ... except..., check out [how these are handled in the webapp](https://github.com/ISISScientificComputing/autoreduce-frontend/blob/master/autoreduce_frontend/reduction_viewer/views/configure_new_batch_run.py#L76-L88) though we don't want a similar `render_error` call

Also note that for the `ConnectionError` in the example to work you will need to import `from requests.exceptions import ConnectionError` because for whatever silly reason that is NOT the same as the `__builtins__.ConnectionError`...

autoreduce_qp/model/database/records.py
```diff
@@ -66,10 +68,30 @@ def get_script_and_arguments(instrument: Instrument, script: str, arguments: dic
...
 
+    fetch_from_remote_source(arguments)
     arguments_str = json.dumps(arguments, separators=(',', ':'))
     return script, arguments_str
 
 
+def fetch_from_remote_source(arguments):
+    """
+    Update a supplied dictionary to contain the texts within any supplied raw
+    GitHub files.
```
dtasev, line 79:
```suggestion
    Fetches the values for any variables that match the remote syntax expected in reduce_vars.py
```

dtasev (approved):

PR #1411, opened 2021-11-16
Remove ascii encode

james0209:
### Summary of work
Remove unused ascii encode from when project used older version of Python.

### DIFF
--- a/autoreduce_qp/queue_processor/reduction/service.py
+++ b/autoreduce_qp/queue_processor/reduction/service.py
@@ -181,7 +181,7 @@ class ReductionScript:
                 old_dict.update(source_dict)
                 setattr(self.module.web_var, dict_name, old_dict)
 
-            encoded_dict = {k: (v) for k, v in reduction_arguments[dict_name].items()}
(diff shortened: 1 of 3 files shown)


autoreduce_qp/queue_processor/reduction/service.py
```diff
@@ -181,11 +181,7 @@ def merge_dict_to_name(source_dict):
                 old_dict.update(source_dict)
                 setattr(self.module.web_var, dict_name, old_dict)
 
-            def ascii_encode(var):
-                """ ASCII encode var. """
-                return var.encode('ascii', 'ignore') if type(var).__name__ == "unicode" else var
-
-            encoded_dict = {k: ascii_encode(v) for k, v in reduction_arguments[dict_name].items()}
+            encoded_dict = dict(reduction_arguments[dict_name].items())
```
dtasev, line 184: I think it can just be `merge_dict_to_name(reduction_arguments[dict_name])` as dict of `.items()` basically makes a copy, what do you think?

PR #1412, opened 2021-11-16
move FROM to ghcr.io

james0209:
### Summary of work
Change FROM to use containers on ghcr.io


dtasev:
This has been updated to the new ghcr.io/autoreduction!

PR #1418, opened 2021-12-21
1506 generic processing environ ar 1506

james0209:
### Summary of work
Creates Generic processing environments for Reduction:

Changes the way Reduction is run from a Subprocess to Docker Container

- This allows for the autoreduction service to not be coupled to the same python version that Mantid requires
- Provided flexibility of available data treatment software
- Improves ability to truly reproduce results by tightly controlling versions of available software.
- Allows adoption of Autoreduction by scientific technique areas that cannot use the tools available in the current environment or require additional tools (e.g. neutron imaging).


autoreduce_qp/queue_processor/reduction/process_manager.py
```diff
@@ -27,33 +25,89 @@ def __init__(self, message: Message, run_name: str) -> None:
...
+            # Create a container and run it. Equivalent to docker run.
+            # To build runner-mantid image, in same directory as Dockerfile run:
+            # docker build -t runner-mantid .
 
-                # Copy and update the subprocess environment to inherit the
-                # parent one, and append the PYTHONPATH of the queue_processor module
-                environment = os.environ.copy()
+            # Pull image, containers.create doesn't pull everytime
+            client.images.pull('ghcr.io/autoreduction/runner-mantid:6.2.0')
+            mount = f'{os.path.expanduser("~")}/.autoreduce/dev/test-archive'
```
dtasev, line 52: This can use `ARCHIVE_ROOT` variable `from autoreduce_utils.settings import ARCHIVE_ROOT`

autoreduce_qp/queue_processor/reduction/process_manager.py
```diff
@@ -27,33 +25,89 @@ def __init__(self, message: Message, run_name: str) -> None:
...
+            if not os.path.exists(mount):
+                Path(mount).mkdir(parents=True, exist_ok=True)
+            os.chmod(mount, 0o777)
 
-                pythonpath = environment.get("PYTHONPATH", "")
-                pythonpath += f':{RUNNER_PATH.split("autoreduce_qp")[0]}'
-                pythonpath = environment["PYTHONPATH"] = pythonpath
+            Path(f'{os.path.expanduser("~")}/.autoreduce/dev/reduced-data').mkdir(parents=True, exist_ok=True)
```
dtasev, line 57: This should switch between `~/.autoreduce/dev/reduced-data` and `/instrument` depending on `os.environ["AUTOREDUCTION_PRODUCTION"]`, though it might be better to extract the base of that path in https://github.com/autoreduction/autoreduce-utils/blob/main/autoreduce_utils/settings.py#L46

autoreduce_qp/queue_processor/reduction/process_manager.py
```diff
@@ -27,33 +25,89 @@ def __init__(self, message: Message, run_name: str) -> None:
...
+                        'mode': 'rw'
+                    },
+                    f'{os.path.expanduser("~")}/.autoreduce/dev/reduced-data': {
+                        'bind': '/instrument/',
+                        'mode': 'rw'
+                    },
+                    temp_dir.name: {
+                        'bind': '/output/',
+                        'mode': 'rw'
+                    },
```
dtasev, line 79: Mount paths not used from above?

autoreduce_qp/queue_processor/reduction/process_manager.py
```diff
@@ -27,33 +25,89 @@ def __init__(self, message: Message, run_name: str) -> None:
...
-                pythonpath = environment.get("PYTHONPATH", "")
-                pythonpath += f':{RUNNER_PATH.split("autoreduce_qp")[0]}'
-                pythonpath = environment["PYTHONPATH"] = pythonpath
+            Path(f'{os.path.expanduser("~")}/.autoreduce/dev/reduced-data').mkdir(parents=True, exist_ok=True)
+            os.chmod(f'{os.path.expanduser("~")}/.autoreduce/dev/reduced-data', 0o777)
 
-                # Run process until finished and check the exit code for success
-                subprocess.run(args, check=True, env=environment)
+            container = client.containers.create(
+                image="ghcr.io/autoreduction/runner-mantid:6.2.0",
```
dtasev, line 61: Should store the image name in a var somewhere above.

autoreduce_qp/queue_processor/reduction/process_manager.py
```diff
@@ -27,33 +25,89 @@ def __init__(self, message: Message, run_name: str) -> None:
...
+                        'bind': '/instrument/',
+                        'mode': 'rw'
+                    },
+                    temp_dir.name: {
+                        'bind': '/output/',
+                        'mode': 'rw'
+                    },
+                },
+                tty=True,
+                environment=["AUTOREDUCTION_PRODUCTION=1"],
```
dtasev, line 82: Does this need to be added? I don't think the runner containers should care whether it's production or not, as the only thing that changes is the mount src paths

dtasev (changes requested):
Looking good! Will give a test locally

autoreduce_qp/queue_processor/reduction/runner.py
```diff
@@ -114,6 +115,21 @@ def _get_mantid_version():
         return None
 
 
+def write_reduction_message(reduction, temp_output_file=None):
+    """
+    Write the reduction message to the reduction directory
+    """
+    if temp_output_file is None:
```
dtasev, line 122: I'm not sure there's a reason to try and maintain compatibility with `temp_output_file` since with this new structure that parameter will never be passed

autoreduce_qp/queue_processor/reduction/runner.py
```diff
@@ -123,7 +139,11 @@ def main():
     Additionally, the resulting Message is written to a temporary file which the
     parent process reads back to mark the result of the reduction run in the DB.
     """
-    data, temp_output_file, run_name = sys.argv[1], sys.argv[2], sys.argv[3]
+    data, run_name = sys.argv[1], sys.argv[2]
+    if len(sys.argv) > 3:
+        temp_output_file = sys.argv[3]
+    else:
+        temp_output_file = None
```
dtasev, line 146: Same for this as comment above - since no `temp_output_file` is expected anymore it will always end up at `/output/output.txt`

dtasev (changes requested):

autoreduce_qp/queue_processor/reduction/process_manager.py
```diff
@@ -27,33 +25,89 @@ def __init__(self, message: Message, run_name: str) -> None:
...
 
-                pythonpath = environment.get("PYTHONPATH", "")
-                pythonpath += f':{RUNNER_PATH.split("autoreduce_qp")[0]}'
-                pythonpath = environment["PYTHONPATH"] = pythonpath
+            Path(f'{os.path.expanduser("~")}/.autoreduce/dev/reduced-data').mkdir(parents=True, exist_ok=True)
+            os.chmod(f'{os.path.expanduser("~")}/.autoreduce/dev/reduced-data', 0o777)
 
-                # Run process until finished and check the exit code for success
-                subprocess.run(args, check=True, env=environment)
+            container = client.containers.create(
```
dtasev, line 60: Perhaps this can use `client.containers.run` and pass in the args directly?

Something like:

```
 logs = client.containers.run(
                image="ghcr.io/autoreduction/runner-mantid:6.2.0",
                command=args,
                volumes={...},
                environment=["AUTOREDUCTION_PRODUCTION=1"],
            )
```
From the docs it seems this will just return the STDOUT that we want to log afterwards with something like `logger.info("Container logs %s", logs.decode("utf-8"))`

autoreduce_qp/queue_processor/reduction/process_manager.py
```diff
@@ -27,33 +25,89 @@ def __init__(self, message: Message, run_name: str) -> None:
     def run(self) -> Message:
         """Run the reduction subprocess."""
         try:
+            # Create a temp directory to store reduction result
+            # Volume bind for the container to write to and host to read from
+            temp_dir = tempfile.TemporaryDirectory()  # pylint: disable=consider-using-with
+            os.chmod(temp_dir.name, 0o777)
```
dtasev, line 31: I think this chmod needs to always be done, but the rest can be done only `if not AUTOREDUCTION_PRODUCTION` and that can be handled in Ansible

dtasev (changes requested):

dtasev (changes requested):
Just a couple small comments, but looks really good! I will test it locally tomorrow

autoreduce_qp/queue_processor/reduction/process_manager.py
```diff
@@ -27,33 +26,88 @@ def __init__(self, message: Message, run_name: str) -> None:
...
+                logger.info("Calling: %s %s %s %s ", "python3", "runner.py", serialized_vars_truncated, self.run_name)
+
+                # Return a client configured from environment variables
+                # The environment variables used are the same as those used by the Docker command-line client
+                # https://docs.docker.com/engine/reference/commandline/cli/#environment-variables
+                client = docker.from_env()
+
+                # Pull all tags of runner-mantid as a list
+                # Could be used to populate a dropdown menu of images
+                images = client.images.pull('ghcr.io/autoreduction/runner-mantid', all_tags=True)
```
dtasev, line 49: The issue I see with `all_tags=True` is that if you push a development one then it'll be pulled onto the production qp machines. While not terrible, it seems like an unnecessary waste of disk space.

And ideally software will be selected either in reduce_vars (with a new special attribute), or from the webapp (which will fill the options from the DB, rather than communicating with the queue processor)

autoreduce_qp/queue_processor/reduction/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ghcr.io/autoreduction/autoreduce:latest
+COPY . .
+RUN export DEBIAN_FRONTEND=noninteractive
+ARG DEBIAN_FRONTEND=noninteractive
+RUN python3 -m pip install autoreduce-qp==22.0.0.dev52 --extra-index-url https://api.packagr.app/public
+RUN python3 -m pip install pytest
+RUN python3 -m pip install parameterized
+USER isisautoreduce
```
dtasev, line 8: Can this dockerfile be moved to https://github.com/autoreduction/autoreduce-containers or a new repo just for these runner images? 

After this PR is merged I guess we want to stop using the extra index `https://api.packagr.app/public` and move to the production package

dtasev (changes requested):
It looks good! Quickly tested locally and I see no issues.

There's some potential refactoring around the image name, but that will be part of a follow up issue where users can select the software

I've left a couple of comments just to discuss and I'm happy to merge

PR #1460, opened 2022-07-11
Dockerfiles for Queue-Processor, and Autoreduce-Runner with Mantid 6.3 & 6.4

james0209:
### Summary of work

- Create 2 main Dockerfiles: one for queue-processor; one for queue-processor with Mantid in the environment
- Version of Mantid required (from conda) can be passed as a build arg to the respective Dockerfile
- Make use of _conda-pack_ to reduce Image size. _A tool that let’s you package a Conda environment into a standalone environment, with no need for the Conda toolchain. Once environment has been packaged that way, it can be copied into a new smaller image that only contains the self-contained environment_


container/runner_py38_mantid.Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+    gnupg \
+    git \
+    software-properties-common \
+    python3-dev \
+    default-libmysqlclient-dev \
+    build-essential \
+    gcc
+
+# Install conda-pack into the base environment
+RUN conda install conda-pack
```
Pasarus, line 19: Is `conda-pack` necessary as you can install an environment to a specific directory using `--prefix`?

Pasarus (changes requested):
Besides the one question looks great!

Another great PR!

PR #1464, opened 2022-08-02
Fix writing temp output file (AR-1633)

james0209:
## Curent problem
The current implementation creates a TemporaryDirectory for the runner.py to write the output message to. This works great locally, but is causing issues with permissions when run on qp_external (mounting a tempdir in a Docker on Docker situation is causing Permission Denied Error 13 errors).

## Solution
The solution in this PR is to use an already existing volume (which mounts the user's ~/.autoreduce to /home/isisautoreduce/.autoreduce) and write the file there. The same file (output.txt) can be used be re-written as a copy is then made in the database. 

## Further Work
For local deployments, there is use of chmod 777 on all of the volume binds - we should look into changing this in the future if possible (either lower permission level or use something like chown?). These chmod's do not apply when the AUTOREDUCTION_PRODUCTION env var is set.


autoreduce_qp/queue_processor/reduction/process_manager.py
```diff
@@ -32,75 +31,67 @@ def __init__(self, message: Message, run_name: str, software: Software) -> None:
...
+
+            # Return a client configured from environment variables
+            # The environment variables used are the same as those used by the Docker command-line client
+            # https://docs.docker.com/engine/reference/commandline/cli/#environment-variables
+            client = docker.from_env()
+
+            image = get_correct_image(client, self.software)
+
+            if "AUTOREDUCTION_PRODUCTION" in os.environ:
+                reduced_data = Path('/instrument')
```
Pasarus, line 50: Could this line be added to our test code coverage?

Pasarus (changes requested):

autoreduce_qp/queue_processor/reduction/process_manager.py
james0209: I don't think it's worth it as it's just a variable assignment. Codecov just has a weird thing with if-else statements (even if it's just assigning a variable)

Pasarus: I appreciate that, however, the code after this variable assignment has different output based on the variable assignment right?

Pasarus (approved):

PR #1465, opened 2022-08-02
Update setup-python to v4.1.0

Pasarus:
Run tests CI passes

#AR-1636

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4.1.0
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Set up Python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4.1.0
```
DanNixon, line 17:
```suggestion
        uses: actions/setup-python@v4
```

`v4` == `v4.*` which seems more like what we actually want.

DanNixon (changes requested):

james0209 (approved):

DanNixon (approved):

PR #1496, opened 2022-09-01
Use the DJANGO_SECRET_ENV environment variable for key

### LINKED ISSUE #1483: Make Queue-Processor and REST-API get SECRET_KEY from env
Currently, the Django Secret Key is public and hard-coded into the queue-processor and rest-api. This is insecure and should be changed to the implementation in the frontend.

The frontend uses:

`SECRET_KEY = os.getenv('DJANGO_SECRET_KEY')`

Changes that need doing:

Generate secret keys for qp and rest-api, add them to Ansible vault (prod and a fake key for dev vault)

Change qp and rest-api to get secret key from env (like in command above)

Add reference to Django Secret Key in rest_api, queue_processor, and qp_external roles (rest-api shares K8s Secret with frontend, so either have DJANGO_FRONTEND_SECRET_KEY and DJANGO_REST_API_SECRET_KEY or create seperate Secret for rest-api) (First option would be easier to not have multiple Secrets with same data apart from one line)

Pasarus:
### Summary of work

I checked that the queue-processor has the DJANGO_SECRET_ENV environment variable, in the k8s repo we have a secrets file [here](https://github.com/autoreduction/k8s/blob/b50f70f6506a6c3eab2d688713008e473f00191a/ansible/roles/queue_processor/templates/qp-secret.yaml#L8), and that is given the queue-processor in the k8s repo aswell as a secretRef in this [file](https://github.com/autoreduction/k8s/blob/b50f70f6506a6c3eab2d688713008e473f00191a/ansible/roles/queue_processor/templates/deployment.yaml#L31)

### How to test your work

Just code review here.

### Additional comments

Fixes #1483 

No release notes as not user facing and just a security update.

### DIFF
--- a/autoreduce_qp/autoreduce_django/settings.py
+++ b/autoreduce_qp/autoreduce_django/settings.py
@@ -9,6 +9,7 @@ https://docs.djangoproject.com/en/3.2/topics/settings/
 For the full list of settings and their values, see
 https://docs.djangoproject.com/en/3.2/ref/settings/
 """
+import os
 from pathlib import Path
 
 from autoreduce_db.autoreduce_django.settings import DATABASES as autoreduce_db_settings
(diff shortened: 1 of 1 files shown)


DanNixon (approved):
:+1:

DanNixon:
FWIW https://github.com/autoreduction/k8s/blob/b50f70f6506a6c3eab2d688713008e473f00191a/ansible/roles/qp_external/tasks/deploy_qp.yaml#L27 is where the variable is set for the external queue processor (which is currently the one we use).

